PR #170, opened 2015-04-14

Lukx19:

tested in ROS Indigo

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -475,6 +475,7 @@ namespace RobotLocalization
       mapFrameId_ = tf2NameSanitizer(mapFrameId_);
       odomFrameId_ = tf2NameSanitizer(odomFrameId_);
       baseLinkFrameId_ = tf2NameSanitizer(baseLinkFrameId_);
+      worldFrameId_ = tf2NameSanitizer(worldFrameId_);
     // Transform future dating
     double offsetTmp;
     nhLocal_.param("transform_time_offset", offsetTmp, 0.0);
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -361,6 +361,25 @@ namespace RobotLocalization
   }
 
   template<typename T>
+  std::string RosFilter<T>::tf2NameSanitizer(const std::string & param )
+  {
+    if(!tfPrefix_.empty())
+    { // tf prefix is set
+      if(param.at(0)=='/')
+        return tfPrefix_ + param;
+      else
+        return tfPrefix_+ "/" + param;
+    }else
+    {
+       // remove leading slash
+       if(param.at(0)=='/')
+        return param.substr(1);
+    }
+    // everything is OK
+    return param;
+  }
+
+  template<typename T>
   void RosFilter<T>::loadParams()
   {
     // For diagnostic purposes, collect information about how many different
@@ -408,12 +427,11 @@ namespace RobotLocalization
     }
 
     // Try to resolve the tf_prefix
-    nh_.param("/tf_prefix", tfPrefix_, std::string(""));
-
-    if (tfPrefix_.empty())
-    {
-      nh_.param("tf_prefix", tfPrefix_, std::string(""));
-    }
+    std::string tf_prefix_path;
+    if (nhLocal_.searchParam("tf_prefix", tf_prefix_path))
+      nhLocal_.getParam(tf_prefix_path, tfPrefix_);
+    else
+      tfPrefix_="";
 
     // These params specify the name of the robot's body frame (typically
     // base_link) and odometry frame (typically odom)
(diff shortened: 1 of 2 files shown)


Lukx19:
In my original pull I forgot about worldFrameId. This frame should be tf_prefixed as well. My last commit adds this small change. See  commit 2a18eac643298571d4239ffcc0c4443da43b945b. I have not added it to other branches. If it is needed please tell me.

src/ros_filter.cpp
```diff
@@ -408,12 +427,11 @@ namespace RobotLocalization
     }
 
     // Try to resolve the tf_prefix
-    nh_.param("/tf_prefix", tfPrefix_, std::string(""));
-
-    if (tfPrefix_.empty())
-    {
-      nh_.param("tf_prefix", tfPrefix_, std::string(""));
-    }
+    std::string tf_prefix_path;
```
ayrton04: Please change this variable name to tfPrefixPath so as to maintain consistency with variable names

src/ros_filter.cpp
```diff
@@ -361,6 +361,25 @@ namespace RobotLocalization
   }
 
   template<typename T>
+  std::string RosFilter<T>::tf2NameSanitizer(const std::string & param )
+  {
+    if(!tfPrefix_.empty())
+    { // tf prefix is set
```
ayrton04: Line break here, before the comment (pedantic, I know, but I want to be consistent)

src/ros_filter.cpp
```diff
@@ -361,6 +361,25 @@ namespace RobotLocalization
...
+      if(param.at(0)=='/')
+        return tfPrefix_ + param;
```
ayrton04: Please use curly braces for all conditional statements, and use line breaks:

if(foo)
{
  statement;
}

src/ros_filter.cpp
```diff
@@ -361,6 +361,25 @@ namespace RobotLocalization
...
+      else
+        return tfPrefix_+ "/" + param;
+    }else
```
ayrton04: New line in between } and the else, please

src/ros_filter.cpp
```diff
@@ -408,12 +427,11 @@ namespace RobotLocalization
...
-    nh_.param("/tf_prefix", tfPrefix_, std::string(""));
-
-    if (tfPrefix_.empty())
-    {
-      nh_.param("tf_prefix", tfPrefix_, std::string(""));
-    }
+    std::string tf_prefix_path;
+    if (nhLocal_.searchParam("tf_prefix", tf_prefix_path))
+      nhLocal_.getParam(tf_prefix_path, tfPrefix_);
+    else
```
ayrton04: Same

src/ros_filter.cpp
```diff
@@ -361,6 +361,25 @@ namespace RobotLocalization
...
+  {
+    if(!tfPrefix_.empty())
+    { // tf prefix is set
+      if(param.at(0)=='/')
+        return tfPrefix_ + param;
+      else
+        return tfPrefix_+ "/" + param;
+    }else
+    {
+       // remove leading slash
```
ayrton04: Capitalize comment, please

src/ros_filter.cpp
```diff
@@ -361,6 +361,25 @@ namespace RobotLocalization
...
+       if(param.at(0)=='/')
+        return param.substr(1);
+    }
+    // everything is OK
```
ayrton04: Same

src/ros_filter.cpp
```diff
@@ -361,6 +361,25 @@ namespace RobotLocalization
...
+    { // tf prefix is set
+      if(param.at(0)=='/')
+        return tfPrefix_ + param;
+      else
+        return tfPrefix_+ "/" + param;
+    }else
+    {
+       // remove leading slash
+       if(param.at(0)=='/')
+        return param.substr(1);
```
ayrton04: Two spaces for indentation, please

src/ros_filter.cpp
```diff
@@ -408,12 +427,11 @@ namespace RobotLocalization
...
     // Try to resolve the tf_prefix
-    nh_.param("/tf_prefix", tfPrefix_, std::string(""));
-
-    if (tfPrefix_.empty())
-    {
-      nh_.param("tf_prefix", tfPrefix_, std::string(""));
-    }
+    std::string tf_prefix_path;
+    if (nhLocal_.searchParam("tf_prefix", tf_prefix_path))
+      nhLocal_.getParam(tf_prefix_path, tfPrefix_);
```
ayrton04: Curly braces

src/ros_filter.cpp
```diff
@@ -408,12 +427,11 @@ namespace RobotLocalization
...
+    else
+      tfPrefix_="";
```
ayrton04: Please put spaces between the variable, =, and the value

ayrton04:
Just a few things for you to update, if you don't mind. Don't bother with the other branches. I'll merge this into them when it's done.

PR #171, opened 2015-04-14

Lukx19:

tested in ROS Indigo , but it should work elsewhere as well


ayrton04:
Closing. We'll work from #170 and I'll merge it into the other branches when it's done

PR #172, opened 2015-04-14

Lukx19:

tested in ROS Indigo ,but it should work elsewhere as well

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -353,6 +353,14 @@ namespace RobotLocalization
                         Eigen::VectorXd &measurement,
                         Eigen::MatrixXd &measurementCovariance);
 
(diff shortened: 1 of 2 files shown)


ayrton04:
Great, thanks for taking the time to do this. I know `tf2` does away with `tf_prefix` in general and also eliminates the use of leading slashes in `frame_id`s, but as the package has not yet been converted to tf2 (it's on my list), I'll want to make sure this won't break any existing systems. I'll test it soon and merge when I'm done. Thanks again!

ayrton04:
I have completed the tf2 migration on my local fork. Once I test a bit more and get everything merged into this repo, I will merge your changes. Thanks again!

ayrton04:
Closing. We'll work from #170 and I'll merge it into the other branches when it's done

PR #173, opened 2015-04-22
remove diagnostics_msgs STALE from hydro build

servos:
diagnostic_msgs::DiagnosticStatus::STALE does not exist in hydro. Therefore this breaks the build currently.


ayrton04:
Yeah, I need to think about how I want to address this. I've been trying to keep all the non-EOL tracks identical, and this is the first ROS change I've had to work around. It looks like I'll have to actually start having different versions of the software for the different tracks. 

Anyway, thanks for the PR!

PR #194, opened 2015-05-05
Disable diagnostic error for missing Z component in 2D mode

servos:
Don't output error for not setting a Z component source when we are in 2D mode.


ayrton04:
Nice catch, thanks!

PR #202, opened 2015-05-10
Start relative poses at target frame

T045T:
Disclaimer: This is a late-night development to get one specific use case to work. I'd appreciate a review and comment on whether this makes sense for the project at large :)

The description in dcb1b00 describes the change in some detail - basically, with this change, setting a pose input "relative" should result in the initial ```odom -> base_link``` tf being the identity, rather than ```base_link -> sensor_frame```.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2076,21 +2076,23 @@ namespace RobotLocalization
     // We'll need this later for storing this measurement for differential integration
     tf2::Transform curMeasurement;
 
-    // 2. If this sensor is relative, remove the initial measurement for it
+    // 2. Get the target frame transformation
+    tf2::Transform targetFrameTrans;
+    bool canTransform = lookupTransformSafe(finalTargetFrame, poseTmp.frame_id_, poseTmp.stamp_, targetFrameTrans);
+
+    // 3. If this sensor is relative, remove the initial measurement for it
+    //    *after* transforming to the target frame! This way, the relative pose always
+    //    starts as identity w.r.t the final target frame.
     if(relative)
     {
       if(initialMeasurements_.count(topicName) == 0)
       {
-        initialMeasurements_.insert(std::pair<std::string, tf2::Transform>(topicName, poseTmp));
+        initialMeasurements_.insert(std::pair<std::string, tf2::Transform>(topicName, poseTmp * targetFrameTrans));
       }
 
       tf2::Transform initialMeasurement = initialMeasurements_[topicName];
       poseTmp.setData(initialMeasurement.inverseTimes(poseTmp));
     }
-
-    // 3. Get the target frame transformation
-    tf2::Transform targetFrameTrans;
-    bool canTransform = lookupTransformSafe(finalTargetFrame, poseTmp.frame_id_, poseTmp.stamp_, targetFrameTrans);
 
     // 4. robot_localization lets users configure which variables from the sensor should be
     //    fused with the filter. This is specified at the sensor level. However, the data


ayrton04:
Thanks for the PR! I'll look this over tonight. I think you're probably correct on this. Thanks again!

ayrton04:
Looking at this a little deeper, I think we may be looking at another case where IMU data has to be handled differently. For example, imagine a robot that drives sideways on the wall. Its world coordinate frame is rotated by 90 degrees from the ground plane. Let's say we are running an instance of ekf_localization_node on some device in the ground plane, and we wanted to know the position of the wall robot w.r.t. the ground plane. If we turn on relative mode, we take the wall vehicle's world frame X-Y position (and remember that since it started sideways, its X-Y plane is the wall itself), and "removes" its initial X-Y position. This leaves it in some position in the X-Y frame.*Now* we need to transform that position so that it's in the X-Z frame, so we apply the target frame transformation. In other words, the order of operations was (1) compute the difference, and (2) apply the transform.

I think part of the issue you're having is that IMUs always report their orientations w.r.t. the world frame. In our wall robot example, it will start out with a roll of 90, and then after moving around a bit, it will still have a roll of 90. When we take the difference, we get 0 (correct), but then if we apply the transform to the ground plane device running the EKF, it adds 90 again. 

If the IMU were to always report its orientation data w.r.t. its world frame when the robot started, the method I have would work: if the robot on the wall turned so it was driving straight up the wall, its world frame *yaw* should be 90. In that case, the transformation scheme I'm using would work: first we remove the initial measurement, yielding some yaw value, and then we apply the transformation, so that yaw for the wall vehicle is pitch for the device in the ground plane. Unfortunately, IMUs don't work this way.

One other thing: you're applying the transform in the wrong direction. You're multiplying the pose times the target frame as the initial measurement, so what you get is this:

First time we get a measurement (line 2090 in your version):
{{{
initial_meas = pose_1 * target_trans;
}}}

Then, for some future measurement at time N (line 2094 in the your version):
{{{
relative_meas = initial_meas^-1 * pose_N 
relative_meas = (pose_1 * target_trans)^-1 * pose_N
relative_meas = target_trans^-1 * pose_1^-1 * pose_N = target_trans^-1 * (pose_1^-1 * pose_N)
}}}

Finally, we apply the target frame transformation again (lines 2270-2284 in your version, but only because we're using IMU data):
{{{
final_meas = target_trans * transformed_meas 
final_meas = target_trans * target_trans^-1 * (pose_1^-1 * pose_N)
final_meas = pose_1^-1 * pose_N
}}}

In the end, all you're doing is removing the target frame transformation entirely. While this gives the correct result for what you want, there's a better way. The real issue here - and I concede there is a bug - is that for relative mode measurements of IMU data, we *don't* want to apply the orientation correction for the IMU. If your IMU is rolled over and we always "remove" that initial orientation, then we have no need to carry out the transform.

Try this: before line 2248 in your version, you need to insert this:
{{{
  else if(relative)
  {
    if(!imuData)
    {
      poseTmp.mult(targetFrameTrans, poseTmp);
    }
  }
}}}

Try inserting that and then undoing your changes. Let me know how it goes!

T045T:
First off, thanks for taking the time to type this up! I'm not 100%  sure I got everything you talked about, but I think I did for the major points.

You're obviously right about me removing ```targetFrameTrans``` entirely. It was late and I didn't notice, sorry.

I haven't tried out your suggestion yet, but I will, as soon as possible.
However, I'm not sure that simply omitting targetFrameTrans for relative measurements is correct, either, if we're taking into account the entire IMU pose, rather than just its orientation;
Consider a long robot with ```base_link``` at the front and the IMU mounted way at the back.
If we lift that back end, but leave the front at the same height, the IMU will report its pitch changing accordingly, but __also__ its z coordinate increasing.
But since we're interested in the movement of ```base_link```, that ist not what we want.

Instead, I think we need to do something like the following (caveat: it is late again and I'm not sure if the order is correct this time...):
```
initial_meas  = pose_0 * target_transform
relative_meas = initial_meas^-1 * pose_N * target_transform
              = (pose_0 * target_transform)^-1  * pose_N * target_transform
              = target_transform^-1 * pose_0^-1 * pose_N * target_transform
```
Does that make sense?

ayrton04:
IMUs don't report position, only orientation (so no Z). They report linear acceleration, which is handled in a different callback function. Sensors that provide Z position that are not IMUs are handled differently in that callback function.

ayrton04:
However this does raise an interesting point, which is unrelated to the relative parameter: let's say we have a depth sensor on the rear end of a UUV. The depth sensor is measuring data in the world frame (*odom* or *map*). Right now, the `poseCallback` function attempts to transform pose data from the reported message `frame_id` to the *odom* (or whatever your *world_frame* parameter is set to) frame. So for our depth sensor, you would need to define a transform from *odom* to the *depth_sensor_frame* that technically has the transform from *base_link*->*depth_sensor_frame* in it.  This is wrong, as it won't handle the vehicle's orientation. What we really want to do is this:

1. Define the transform of the sensor from *base_link*->*depth_sensor_frame* statically. 
2. When we get a new measurement for the depth sensor, we need to rotate the *base_link*->*depth_sensor_frame* transform *by* the *odom*->*base_link* orientation (but we should not apply the world position!). This will give us the rotated offset of the depth sensor w.r.t. the the vehicle center.
3. We then apply that transform to the depth sensor measurement itself. 

So, for example, if we have a depth sensor that has a X offset of -1.5 meters from the vehicle center, and a +Z offset of 0.5 meters, then when the vehicle is level, then (assuming our depth sensor reads negative values as it gets deeper, so -Z is down), we just subtract the Z offset from the raw measurement to get the vehicle center measurement. However, if the vehicle is pitched down, then the "true" vertical offset in the *odom* frame is a function of both the X offset *and* the Z offset. Therefore, we need a combination of the *base_link*->*depth_sensor_frame* transform and the vehicle's current orientation, which is specified by part of the *odom*->*base_link* transform.

T045T:
Oh. Sorry, I got a bit lost in ```ros_filter.cpp``` there and didn't notice translation is handled elsewhere.

I'm not sure I follow your other point. There should be a transform from *odom* to *base_link* - possibly published by one of the ```robot_localization``` nodes, and including the vehicle's orientation - and one from *base_link* to *depth_sensor_frame*, most likely as part of the vehicle's URDF. Why wouldn't that TF tree handle orientation?

ayrton04:
For this reason (and I've run into this before):

Let's say we simply use the `tf` method and go from the *sensor_frame* to *base_link*. What really happens is that the measurement gets transformed as
```
final_meas = odom_base_link_trans * base_link_sensor_trans * measurement
```

We first apply the *base_link* to *depth_sensor* transformation. Let's say that the sensor has an offset of (-1, 0, 0.5) from the vehicle center, and for now, we'll ignore the orientation of the sensor w.r.t. the body frame, as it doesn't matter (a depth sensor rotated on its side reads the same depth). We get a depth measurement from the sensor of 10 meters. Now we apply the *base_link_sensor_trans* transform, and we get a depth of 10.5 meters. Now, let's say our vehicle is at pose (0, 0, 10), with RPY of (0, 0.4, 0). This is the same as the *odom_base_link_trans* (technically its inverse, but that doesn't matter). We apply that transform and we get (0, 0, 10.5) rotated by an RPY of (0, 0, 0.4), and then we add (0, 0, 10) to it, yielding something like (just making this up) (2.3, 0, 19.2). 

What we lose in this transform chain is the fact that we only want to rotate the linear *offset* of the sensor from the vehicle origin. Imagine that the vehicle is pitched down 90 degrees. The offset from the sensor to the vehicle centroid is 1 meter (the X offset). Now imagine it's level. The offset from the vehicle centroid is 0.5 meters. In both of these cases, you want to rotate that linear offset (*base_link*->*sensor_frame*) by *only* the vehicle's current orientation (*just* the orientation of the *odom*->*base_link* transform).

It should be pretty straightforward to write a few lines of code with `tf` that go through these steps.

T045T:
I see. So the depth sensor always measures distance straight down?
We've encountered a similar situation, and yeah, the TF tree isn't perfectly suited to setups like that. We ended up using a virtual link to 'undo' the rotation of *base_link*, so the measurement frame is always level with the ground.

For sensors that measure in a straight line from their reference frames (say, a Kinect or laser scanner), this wouldn't be an issue, would it?

ayrton04:
If a sensor is measuring a range to some landmark, then any node that consumes that information should be able to use that information directly. If a sensor is using a Kinect to determine its 3D pose, then  I think this could be an issue. I'm not seeing many responses to my ROS Driver Sig post on this topic, so I'm tempted for now to push any such logic to the driver and just carry out the transforms as I am. 

In any case, can you verify if the code change (reverting your changes and adding my lines) worked? Thanks!

ayrton04:
Worked on this a lot over the weekend. Long story short: I know how to fix this, but it isn't with the code change I gave you here. I just pushed a release, but they haven't been synched yet, so I'll try to fix this and release it ASAP.

T045T:
Thanks for continuing to look into this!
I'm afraid I haven't found the time to test your suggestion yet, sorry.

ayrton04:
I think I've addressed this issue in the latest release. It's out for Indigo, and Jade/Hydro will follow whenever the next package sync happens. Let me know if you have any issues when you've updated.

T045T:
Just to formally wrap this up: The new release works as expected. Thanks!

PR #233, opened 2015-09-02
Fix Jacobian for EKF.

adnanademovic:
While looking into the code, I've noticed inconsistency in lines 277-279. Deriving the RPY rotational matrix (or just looking at the 2 lines above), makes this error obvious.


ayrton04:
Yes, just double checked the math, and you are correct. Thanks a bunch!

adnanademovic:
You're welcome!

PR #238, opened 2015-09-29
Switched cropping of orientation data with mahalanobis check.

tscho1l:
Switched cropping of orientation data in inovationSubset with mahalanobis check to prevent excluding measurements with orientations bigger/smaller than ± PI.

### DIFF
--- a/src/ekf.cpp
+++ b/src/ekf.cpp
@@ -162,29 +162,29 @@ namespace RobotLocalization
 
     innovationSubset = (measurementSubset - stateSubset);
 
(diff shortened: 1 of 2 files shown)


tscho1l:
@ayrton04 You can check and merge if proper.

ayrton04:
Yes, been busy, sorry. Looks good to me, but would you mind dropping it in the UKF as well?

https://github.com/cra-ros-pkg/robot_localization/blob/indigo-devel/src/ukf.cpp#L243

Makes more sense to put them both in a single PR.

Issue #223, opened 2015-07-30
Fix gravitational acceleration correction when orientation data is missing


ayrton04:
If the orientation data isn't filled out in an IMU message (per its [specification](http://docs.ros.org/api/sensor_msgs/html/msg/Imu.html)) and the `remove_gravitational_acceleration` parameter is set to *true*, then when we attempt to remove the gravitational acceleration, we get nans in the measurement (which are ignored, but it shouldn't get that far). We should print a diagnostic message in this case and either ignore the measurement or fuse it without correcting it. I think I prefer the former.

kphil:
An IMU does not provide orientation by default unless there is an attitude computer (and most commonly a magnetometer). Therefore it would be nice being able to use an IMU just providing acceleration and angular velocities with `remove_gravitational_acceleration` parameter set to true.
I would suggest to use orientation from our filters state vector to remove gravity in this case.

ayrton04:
The rotation angles that matter for correcting linear acceleration are roll and pitch, which are typically provided by accelerometer data, yes? You may be able to pass the accelerometer data through another node like imu_filter_madwick as well, which would give you the orientation values you need. In any case, yes, I agree that this should be a feature, but see my comment on the change you proposed.

kphil:
Yes, there are many ways to provide an orientation and many IMUs come with an built-in attitude computer. But if i have a minimal setup with an IMU just providing acceleration and angular velocity it would be nice being able to run it with this package. That's the idea, if someone will provide an orientation we will use it, if not using current states orientation.

ayrton04:
Sounds good. Thanks for the PR!

PR #247, opened 2016-01-15
#223 fix gravitational acceleration correction

kphil:
As suggested in comment to corresponding issue

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2106,7 +2106,18 @@ namespace RobotLocalization
           // Imu message contains no orientation, so we should use orientation
           // from filter state to transform and remove acceleration
           const Eigen::VectorXd &state = filter_.getState();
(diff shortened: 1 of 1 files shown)


src/ros_filter.cpp
```diff
@@ -2100,7 +2100,18 @@ namespace RobotLocalization
         tf2::Vector3 normAcc(0, 0, 9.80665);
         tf2::Quaternion curAttitude;
         tf2::Transform trans;
-        tf2::fromMsg(msg->orientation, curAttitude);
+
+        if(::fabs(msg->orientation_covariance[0] + 1) < 1e-9)
```
ayrton04, line 2104: I like this idea, but this assumes that the IMU and the robot are not offset from each other rotationally (i.e., that you IMU rests level on the robot, and not on its side or upside-down). Really, you need to rotate the state's RPY value back into the IMU's coordinate frame, and then apply this logic.

kphil:
You are absolutely right. I added the transform.

ayrton04:
Looks good, thanks!

Issue #224, opened 2015-08-06
Warn users about bad time stamps in input data


ayrton04:
If messages have bad/empty time stamps, then we should let them know via the diagnostics.

PR #274, opened 2016-04-08
Add reporting of bad timestamps via diagnostics

iwanders:
The first situation which is reported is if the timestamp is before or equal to  the timestamp of the last pose reset. This report is also triggered when the  timestamp is empty (zero) since `lastSetPoseTime_` is also initialised to zero.

The second situation which is reported is when a message is received that has a timestamp prior to the last message received for that topic.

In both situations the behaviour of the filter is to disregard the message and reporting this via diagnostics should be useful. As is proposed in issue #224.

There are four methods which are used as callbacks for the four accepted input types. These four all start with the `(msg->header.stamp <= lastSetPoseTime_)` check at the start of the method. Catching the empty timestamps or those before the last `set_pose` call is straightforward.

Catching the second situation where the timestamps are older than those previously received is done in the `accelerationCallback`, `poseCallback` and `twistCallback`. One issue is that the imu message are split and sent to these three methods with `_pose`, `_acceleration` and `_twist` appended to create the faux topicnames. This results in somewhat overzealous reporting in the diagnostics when the imu timestamp is wrong, it is still clear enough in my opinion. If this is an issue it could be fixed by using a string replacement to remove the addition such that the key for all diagnostic messages from these three ends up being the same `imu0_timestamp`.

In regards to testing the added behaviour: As there currently are no testcases for any of the diagnostic messages I chose to verify the added functionality with a small Python script, not included in the pull request but available in  aa573e3d4ed40aa9ee6f715f18e95e7f8959ef58.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,11 +163,19 @@ if (CATKIN_ENABLE_TESTING)
   catkin_add_gtest(filter_base-test test/test_filter_base.cpp)
   target_link_libraries(filter_base-test filter_base ${catkin_LIBRARIES})
 
+  # This test uses ekf_localization node for convenience.
+  add_rostest_gtest(test_filter_base_diagnostics_timestamps
+                    test/test_filter_base_diagnostics_timestamps.test
+                    test/test_filter_base_diagnostics_timestamps.cpp)
+  target_link_libraries(test_filter_base_diagnostics_timestamps ${catkin_LIBRARIES} ${rostest_LIBRARIES})
+  add_dependencies(test_filter_base_diagnostics_timestamps ekf_localization_node)
+
   #### EKF TESTS #####
   add_rostest_gtest(test_ekf
                     test/test_ekf.test
                     test/test_ekf.cpp)
   target_link_libraries(test_ekf ros_filter ekf ${catkin_LIBRARIES} ${rostest_LIBRARIES})
+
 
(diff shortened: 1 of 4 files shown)


ayrton04:
Hi Ivor! Thanks for the PR. Looks good. I like the idea of having a test for this; would it be too much to add it to the PR and also make it one of the automatically run rostests? I think you'll have to create a new `.test` file (just a launch file) in the test directory that runs your test. See [this](http://wiki.ros.org/rostest/Writing) page (Python example at the bottom, I think) for an example.

iwanders:
Hi Tom!
I agree a test case makes sense, if this feature is included we want to be sure that it actually works and keeps working in the future.

The other test cases are all in C++, so I thought it would be better to stick with that for uniformity. I tried to keep it as future-proof as possible, by iterating over the received diagnostic messages and verifying that we have received ours (by key, not string content).

Conceptually I have a point of concern: We want to test the functionality of an abstract class, but we need a running node that's sending messages on the `/diagnostics` topic. For now I circumvented this problem by just using `ekf_localization_node`. What's your thought on that?

ayrton04:
Sorry for the delay, Ivor. I will respond to this soon.

ayrton04:
Yep, I think this looks good. I will merge this in after I do a release for Indigo and Jade of the current version. Thanks again!

iwanders:
Glad to hear that. If anything comes up that prevents the merge please let me know.

PR #275, opened 2016-04-25
Add parameter to make transform publishing optional

afakihcpr:
It is useful in many use-cases to have the ability to run r_l without TF publising.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -56,6 +56,7 @@ namespace RobotLocalization
       lastSetPoseTime_(0),
       nhLocal_("~"),
       printDiagnostics_(true),
+      publishTransform_(true),
       twoDMode_(false)
   {
     stateVariableNames_.push_back("X");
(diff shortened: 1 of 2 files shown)


src/ros_filter.cpp
```diff
@@ -1342,56 +1346,59 @@ namespace RobotLocalization
 
         // If the worldFrameId_ is the odomFrameId_ frame, then we can just send the transform. If the
         // worldFrameId_ is the mapFrameId_ frame, we'll have some work to do.
-        if (filteredPosition.header.frame_id == odomFrameId_)
```
afakihcpr, line 1345: All the following changes are indentation changes due to the addition of the if (publishTransform_)

l0g1x:
+1

Krystian Gebis
University of Illinois - Chicago
Chicago EDT Robotics - Vice President
www.chicagoedt.org

ayrton04:
Yeah, this is definitely a good idea. I've implemented this same thing more than once, but somehow it's never made its way back into the repo. I want to get Indigo and Jade released with the same source as Kinetic (just released yesterday), and then I'll merge this. Thanks for your help!

src/ros_filter.cpp
ayrton04, line 1345: Thanks for the heads-up.

Issue #276, opened 2016-04-27
Subscribe data with tcpNoDelay()


jbler:
Hi,

I am running the EKF at 200 Hz. At that frequency I am not able to get every spin a new measurement but my subscribing measurement is sent a 200 Hz. Instead my data is often arriving at bunches of a few measurments together and therefore with a lower frequency.  

Would it be possible to subscribe a topic  using the ros::TransportHints().tcpNoDelay() option which seems a reasonable try for me.

ayrton04:
Have you tried increasing the queue size for the sensor in question?

ayrton04:
But the short answer is that yes, I think it would be a good idea. Would you be willing to fork the repo and try it?

jbler:
Thanks for the fast reply. 
Sure, I will give it a try later!

PR #277, opened 2016-04-28
Subscribe packages using tcpNoDelay() 

jbler:
Hi,

this PR is based on #276 .
I added the necessary lines in ros_filter.cpp to assure a faster receiving of incoming measurements.


ayrton04:
So did you first try to just increase the queue size for the message in question?

jbler:
I tried first to increase the queue size but it doesn't solved the problem.  Using this command to give the packages a higher priority instead lead to better results. During my test case it could be shown clearly that the packages are arriving faster. 
This implementation is working fine for my case.

src/ros_filter.cpp
```diff
@@ -559,7 +559,7 @@ namespace RobotLocalization
     setPoseSub_ = nh_.subscribe<geometry_msgs::PoseWithCovarianceStamped>("set_pose",
                                                                           1,
                                                                           &RosFilter<T>::setPoseCallback,
-                                                                          this);
+                                                                          this,ros::TransportHints().tcpNoDelay());
```
ayrton04: Sorry to be picky, but can you please add spaces after all of the commas?

ayrton04:
Looking at the documentation for TCP_NODELAY, I think it would be better if this option were controlled by a per-topic parameter. This particular setting may not be desirable for everyone, and we want to make the default behavior remain as-is, so do you think you could add a `<topicname>_tcp_no_delay` setting to each input type, and make it default to false? Thanks for the PR!

jbler:
Yes, you are right it should be set as false on default behavior. I will implement these changes during this week.

jbler:
I added those changes.  I can not see the conflicts which my implementation produced. I am sure that the code will work fine though. Please tell me any further instructions I have to follow.

ayrton04:
Looks good! Just follow the instructions using the command line (right below this comment). It's most likely complaining about the launch template file. I've moved params into their own directory and made them .yaml files, and the template launch files just point at them. Don't forget to update the UKF template parameters as well. Thanks again!

ayrton04:
:+1:

Issue #306, opened 2016-09-22
Bug report. nav module, use_odometry_yaw=true will cause tf transform error.


asimay:
Bug report. 
when set `use_odometry_yaw = true` in navsat_transform_template.launch,  will report `"could not obtain transform from  to base_link, Error was....."`, and the print info loop infinitely. Screenshot as belows:
![robotbug](https://cloud.githubusercontent.com/assets/8703449/18741095/e4538a86-80df-11e6-9a74-cb57b91bea75.png)
after investigating the code, I find the code missing initialization part of imu msg header initialization, when call `lookupTransformSafe()` , parameter `msg->header.frame_id` is null.
at here:
https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/navsat_transform.cpp#L706
missing:
`imu->header.frame_id = "imu_link";`
`imu->header.stamp = msg->header.stamp;`
the module should define `imu_frame_id` in code.

I modify on my own machine, and test and works well.
Thanks!

ayrton04:
Nice catch! Instead of giving it a static string value, however, you need to give it the `base_link_frame_id_` frame, so that when we look up the transform in the IMU callback, we get the identity transform:

```
sensor_msgs::Imu *imu = new sensor_msgs::Imu();
imu->orientation = msg->pose.pose.orientation;
imu->header.frame_id = base_link_frame_id_;
sensor_msgs::ImuConstPtr imuPtr(imu);
imuCallback(imuPtr);
```

Alternatively, you could do this:

```
sensor_msgs::Imu *imu = new sensor_msgs::Imu();
imu->orientation = msg->pose.pose.orientation;
imu->header.frame_id = msg->child_frame_id;
sensor_msgs::ImuConstPtr imuPtr(imu);
imuCallback(imuPtr);
```

asimay:
ok. Thanks!
I'm not sure whether to use base_link_frame_id_ or imu_frame_id here.
generally, people will define their own imu frame id. 
and ROS [REP 145](https://github.com/paulbovbel/rep/blob/5b0c670130da8222ab28d5aba75dce5a26442ffe/rep-0145.rst)
define the imu frame id is imu_link.....
the `msg->header.stamp` is also used in `lookupTransformSafe()`.

ayrton04:
No, I think you're missing what this code is doing.

[...]

PR #307, opened 2016-09-23
when use_odometry_yaw=true will cause tf transform error.

asimay:
Bug report. nav module, use_odometry_yaw=true will cause tf transform error. #306


src/navsat_transform.cpp
```diff
@@ -705,6 +705,8 @@ namespace RobotLocalization
     {
       sensor_msgs::Imu *imu = new sensor_msgs::Imu();
       imu->orientation = msg->pose.pose.orientation;
+      imu->header.frame_id = msg->header.frame_id;
```
ayrton04: This should be

```
imu->header.frame_id = msg->child_frame_id;
```

PR #308, opened 2016-09-24
use_odometry_yaw=true cause tf transform error

asimay:
I'm so sorry.... I pasted wrong code...


ayrton04 (approved):
Looks good to me, thanks!

PR #316, opened 2016-11-02
Fix more links

jspricke:


ayrton04:
The dangers of copy/paste. Thanks again. I'll leave this open for a bit in case you find more.

Issue #314, opened 2016-11-01
Expose acceleration state via ROS interface


reinzor:
The acceleration state is currently not exposed since the `nav_msgs/Odometry` does not hold the acceleration attribute.

Here: http://www.ros.org/reps/rep-0147.html , people suggested adding an acceleration attribute to the message; however, this would break the current API. An other possibility is exposing the acceleration state via a `geometry_msgs/AccelWithCovarianceStamped` message.

I would like to integrate this into the ROS filter, what do you think about this?

ayrton04:
Sounds like a good idea! Should be pretty straightforward. Just so we don't introduce a new topic for everyone, please make publication optional, with the default parameter value set to false.

reinzor:
OK thanks, I will send a PR later this week.

-Rein

ayrton04:
Merged.

PR #317, opened 2016-11-04
Optionally publish acceleration state

reinzor:
Expose acceleration state via a geometry_msgs/AccelWithCovarianceStamped message. This is configurable by the `publish_acceleration` parameters which defaults to `false`.

#314

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -166,11 +166,16 @@ template<class T> class RosFilter
                    std::vector<int> &updateVector);
 
     //! @brief Retrieves the EKF's output for broadcasting
-    //! @param[out] odomMsg - The standard ROS odometry message to be filled
-    //! @param[out] accelMsg - The standard ROS acceleration message to be filled
+    //! @param[out] message - The standard ROS odometry message to be filled
     //! @return true if the filter is initialized, false otherwise
     //!
-    bool getFilteredOdometryAndAccelMessage(nav_msgs::Odometry &odomMsg, geometry_msgs::AccelWithCovarianceStamped& accelMsg);
+    bool getFilteredOdometryMessage(nav_msgs::Odometry &message);
+
+    //! @brief Retrieves the EKF's acceleration output for broadcasting
+    //! @param[out] message - The standard ROS acceleration message to be filled
+    //! @return true if the filter is initialized, false otherwise
+    //!
+    bool getFilteredAccelMessage(geometry_msgs::AccelWithCovarianceStamped &message);
 
     //! @brief Callback method for receiving all IMU messages
     //! @param[in] msg - The ROS IMU message to take in.
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -47,6 +47,7 @@
 #include <geometry_msgs/TwistStamped.h>
 #include <geometry_msgs/TwistWithCovarianceStamped.h>
 #include <geometry_msgs/PoseWithCovarianceStamped.h>
+#include <geometry_msgs/AccelWithCovarianceStamped.h>
 #include <tf2_ros/transform_listener.h>
 #include <tf2_ros/transform_broadcaster.h>
 #include <tf2_ros/message_filter.h>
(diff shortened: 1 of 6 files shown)


include/robot_localization/ros_filter.h
```diff
@@ -165,10 +166,11 @@ template<class T> class RosFilter
                    std::vector<int> &updateVector);
 
     //! @brief Retrieves the EKF's output for broadcasting
-    //! @param[out] message - The standard ROS odometry message to be filled
+    //! @param[out] odomMsg - The standard ROS odometry message to be filled
+    //! @param[out] accelMsg - The standard ROS acceleration message to be filled
     //! @return true if the filter is initialized, false otherwise
     //!
-    bool getFilteredOdometryMessage(nav_msgs::Odometry &message);
+    bool getFilteredOdometryAndAccelMessage(nav_msgs::Odometry &odomMsg, geometry_msgs::AccelWithCovarianceStamped& accelMsg);
```
ayrton04: I think I'd prefer that we add a second method for assembling the acceleration message. Then we can fill it conditionally base on the value of `publishAcceleration_`, and we're not changing this method signature.

src/ros_filter.cpp
```diff
@@ -268,27 +269,27 @@ namespace RobotLocalization
...
-      message.pose.pose.orientation.y = quat.y();
-      message.pose.pose.orientation.z = quat.z();
-      message.pose.pose.orientation.w = quat.w();
-      message.twist.twist.linear.x = state(StateMemberVx);
-      message.twist.twist.linear.y = state(StateMemberVy);
-      message.twist.twist.linear.z = state(StateMemberVz);
-      message.twist.twist.angular.x = state(StateMemberVroll);
-      message.twist.twist.angular.y = state(StateMemberVpitch);
-      message.twist.twist.angular.z = state(StateMemberVyaw);
+      //! Fill out the odometry message
```
ayrton04: To keep the PR focused, I'd prefer that we not change anything unless needed.

src/ros_filter.cpp
```diff
@@ -646,9 +668,12 @@ namespace RobotLocalization
     FilterUtilities::appendPrefix(tfPrefix, baseLinkFrameId_);
     FilterUtilities::appendPrefix(tfPrefix, worldFrameId_);
 
-    // Whether we're publshing the world_frame->base_link_frame transform
+    // Whether we're publishing the world_frame->base_link_frame transform
```
ayrton04: I realize this is a typo, but I'd rather have a PR that addresses comments and such specifically.

src/ros_filter.cpp
```diff
@@ -1630,8 +1655,11 @@ namespace RobotLocalization
     worldBaseLinkTransMsg_.transform = tf2::toMsg(tf2::Transform::getIdentity());
     mapOdomTransMsg.transform = tf2::toMsg(tf2::Transform::getIdentity());
 
-    // Publisher
-    ros::Publisher positionPub = nh_.advertise<nav_msgs::Odometry>("odometry/filtered", 20);
+    // Publishers
+    ros::Publisher odomPub = nh_.advertise<nav_msgs::Odometry>("odometry/filtered", 20);
+    ros::Publisher accelPub;
+    if (publishAcceleration_)
```
ayrton04, line 1678: Please use curly braces, even for one-line conditional statements.

src/ros_filter.cpp
```diff
@@ -1647,28 +1675,29 @@ namespace RobotLocalization
       integrateMeasurements(curTime);
 
       // Get latest state and publish it
-      nav_msgs::Odometry filteredPosition;
+      nav_msgs::Odometry odomMsg;
```
ayrton04: As above: if we don't have to change this, I'd prefer that we leave it alone, even if the variable name is admittedly not great. This rename ends up creating a lot of changed lines in the PR that don't actually change any functionality.

ayrton04 (changes requested):
Looks like a nice addition, thank you. I'd ask that we keep the PR focused on only the changes that need to be made; I give some more detail in the individual comments.

Thank you for the PR!

reinzor:
I just processed the comments and committed to make the changes minimal

ayrton04:
Looks good. I will test this out and merge. Thanks again!

ayrton04:
Just testing PR jobs on the build server. Please ignore the close and open.

ayrton04:
Testing PR job again.

ayrton04:
Ignore build failure. That's a separate issue with the CMakeLists file. I pulled this down and tested it, and it looks good. Thanks again for the PR!

PR #320, opened 2016-11-08
Added transform_timeout parameter

reinzor:
Due to https://github.com/cra-ros-pkg/robot_localization/pull/319 , the node will throw a lot of errors if the GPS timestamp is newer than the latest transform. Adding this functionality (transform_timeout) will make the tf2 lookup wait for the specified timeout to check if the transform at that timestamp becomes available. This allows the user to process the GPS measurements at the right timestamps but at a later timestamp.


doc/state_estimation_nodes.rst
```diff
@@ -62,6 +62,9 @@ The default values for ``map_frame``, ``odom_frame``, and ``base_link_frame`` ar
 ~transform_time_offset
 ^^^^^^^^^^^^^^^^^^^^^^
 Some packages require that your transforms are future-dated by a small time offset. The value of this parameter will be added to the timestamp of *map->odom* or *odom->base_link* transform being generated by the state estimation nodes in ``robot_localization``.
+~transform_timeout
```
ayrton04: This doesn't render correctly, I'm afraid. I think you need a line break above it:

image

ayrton04 (changes requested):
So I can see how this will be useful for `navsat_transform_node`, but I would advise against its use in the state estimation nodes, for two reasons:
1. Adding a transform timeout means the lookup will block until the transform is available, which will throw off the filter's update frequency.
2. If for whatever reason the user provides a measurement that requires the _world_frame_->_base_link_frame_ transform (i.e., the transform that the state estimation nodes are producing), and the time at which the transform is requested is too recent, this will always fail.

Ideally, we'd add a comment to that effect in the parameter documentation for the state estimation nodes.

ayrton04:
Also, thanks again for the PR!

doc/state_estimation_nodes.rst
ayrton04: If we can fix this documentation issue, I can test and merge this. No rush at all, though.

reinzor: I just added a commit to fix this issue

src/ros_filter_utilities.cpp
```diff
@@ -98,14 +98,15 @@ namespace RosFilterUtilities
                            const std::string &targetFrame,
                            const std::string &sourceFrame,
                            const ros::Time &time,
+                           const ros::Duration &timeout,
```
ayrton04, line 101: Just reading back over this again. I'm a bit concerned that this method will cause an ABI breakage and will also mean that anyone who may be using this method will have to change their source. While I sincerely doubt there are any users who are actually using any of these methods, I think I'd prefer that we add a new overload of this method that has this signature, and keep the old signature intact. You can even make it so that this method simply calls the new one with a 0 timeout parameter.

ayrton04 (changes requested):

src/ros_filter_utilities.cpp
reinzor: I will take a look at this coming Thursday and update the PR

reinzor: Did not have time up until now, will try to look at this on Monday next week :)

reinzor: Hi Tom, took some time but finally got to it. Added an additional method to maintain the old signature.

reinzor:
Processed comments and synced PR with the upstream

ayrton04:
I'll close and reopen this to kick the build server again.

Also, I'm looking at how `tf2` implements the overload of `lookupTransform` when you specify a timeout. The standard `lookupTransform` method is in the `tf2::BufferCore` class, from which `tf2_ros::Buffer` inherits. The overload you're using is in `tf2_ros::Buffer`, and it first calls `canTransform` before calling the base class `lookupTransform` method:

http://docs.ros.org/jade/api/tf2_ros/html/c++/buffer_8cpp_source.html#l00052

So this will add a method call for every measurement, but we'll have to see how that affects performance.

doc/state_estimation_nodes.rst
```diff
@@ -63,6 +63,10 @@ The default values for ``map_frame``, ``odom_frame``, and ``base_link_frame`` ar
 ^^^^^^^^^^^^^^^^^^^^^^
 Some packages require that your transforms are future-dated by a small time offset. The value of this parameter will be added to the timestamp of *map->odom* or *odom->base_link* transform being generated by the state estimation nodes in ``robot_localization``.
 
+~transform_timeout
```
ayrton04, line 66: One last thing: can you add a note than non-zero values may affect the filter's timing? For example, if you tell it to output at 50 Hz, but it has to wait for a transformation for a given measurement, it will block, and you won't get 50 Hz.

reinzor: I just updated the PR

ayrton04 (approved):
LGTM. When this gets past the build server, I will merge.

Issue #260, opened 2016-03-01
Allow to dynamically reconfigure the process noise


falkoschindler:
There are situations where the process noise might change while the Kalman filter is running. This is, e.g., when the robot stops and knows precisely that it's standing still. Then we'd like to see no error growth in the filtered pose. See this post for more background information: http://answers.ros.org/question/227775/how-to-fix-the-ekf-pose-when-the-robot-is-definitely-not-moving/

I guess, making the process noise covariance matrix a dynamically reconfigurable parameter would solve this issue.

ayrton04:
This has become a higher priority. Here's a branch that computes the process noise covariance for pose variables dynamically:

https://github.com/cra-ros-pkg/robot_localization/tree/dynamic-process-noise

I'm debating simplifying it a bit and just using the linear velocity to scale the pose process noise covariance. As it is now, I'm doing that, but I'm using the covariance block for the twist variables and rotating it.

PR #322, opened 2016-11-11
Dynamic process noise

ayrton04:
This addresses #260. I waffled a bit on the best way to implement this, but as I'm assuming the most popular reason for wanting this change is to stop the state covariance growth on pose variables when stationary, I went with a simplified approach. I'd welcome feedback from anyone.

FYI @falkoschindler @tappan-at-git @matthewbwhitaker @servos

### DIFF
--- a/include/robot_localization/filter_base.h
+++ b/include/robot_localization/filter_base.h
@@ -158,6 +158,14 @@ class FilterBase
     //!
     virtual ~FilterBase();
 
+    //! @brief Computes a dynamic process noise covariance matrix using the parameterized state
+    //!
+    //! This allows us to, e.g., not increase the pose covariance values when the vehicle is not moving
+    //!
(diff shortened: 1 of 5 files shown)


tappan-at-git:
:+1: I like it.

falkoschindler:
Hi Tom, it's great to hear that you tackled #260. Unfortunately, I won't be able to try it out in very near future. But I'm very much looking forward to pull and make use of these changes after completing some other subprojects, which are currently in progress.

ayrton04:
No worries! Just tagged you in the event you wanted to weigh in.

ayrton04:
I'll probably merge this later this week, unless I hear otherwise. The new parameter defaults to false, so it won't change the behavior.

ayrton04:
Merging. Feel free to comment further here if there are any more opinions, and we can make changes later.

PR #324, opened 2016-11-14
Feature/robot localization listener

rokusottervanger:
This adds a class that can extrapolate using buffered states (the EKF way), called RobotLocalizationEstimator; a wrapper (RosRobotLocalizationListener) that subscribes to a nav_msgs/Odometry topic and a geometry_msgs/AccelWithCovarianceStamped topic and adds them to its instance of the estimator; and a node that holds an instance of the Listener and also offers the state with covariance at a requested time via a service call.


ayrton04:
Looks like you're missing a dependency:

```
08:59:15 /tmp/catkin_workspace/src/robot_localization/src/ros_robot_localization_listener.cpp:35:41: fatal error: eigen_conversions/eigen_msg.h: No such file or directory
```

ayrton04:
PR job failed because of the `tf/tf.h` include. This is due to the fact that the package uses only `tf2`. Would it be much work to change it? The datatypes are all the same.

rokusottervanger:
If all checks pass, this should be it. The only thing that is missing is the coordinate transformation of covariance matrices. This is relevant when the state of a different frame than the base frame is requested (line 255 in ros_robot_localization_listener.cpp). If anyone could help with that, that would be greatly appreciated.

ayrton04:
Great, thanks! It'll take me some time to review. Re: rotating covariance matrices, see this:

https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ros_filter.cpp#L2796

...and this:

http://robotics.stackexchange.com/questions/2556/how-to-rotate-covariance

CMakeLists.txt
```diff
@@ -15,7 +15,8 @@ find_package(catkin REQUIRED COMPONENTS
   std_msgs
   tf2
   tf2_geometry_msgs
-  tf2_ros)
+  tf2_ros
+  eigen_conversions)
```
ayrton04: Please alphabetize

CMakeLists.txt
```diff
@@ -40,6 +41,7 @@ add_service_files(
   FILES
   SetPose.srv
   SetDatum.srv
+  GetState.srv
```
ayrton04: Please alphabetize, and feel free to alphabetize `SetDatum` as well.

CMakeLists.txt
```diff
@@ -75,6 +78,7 @@ catkin_package(
     tf2
     tf2_geometry_msgs
     tf2_ros
+    eigen_conversions
```
ayrton04: Please alphabetize

CMakeLists.txt
```diff
@@ -109,11 +118,15 @@ target_link_libraries(filter_base filter_utilities ${catkin_LIBRARIES} ${EIGEN3_
...
 target_link_libraries(ros_filter ekf ukf ros_filter_utilities ${catkin_LIBRARIES} ${EIGEN3_LIBRARIES})
+target_link_libraries(robot_localization_estimator filter_utilities ${EIGEN3_LIBRARIES})
+target_link_libraries(ros_robot_localization_listener robot_localization_estimator ${catkin_LIBRARIES} ${EIGEN3_LIBRARIES})
+target_link_libraries(robot_localization_listener_node ros_robot_localization_listener ${catkin_LIBRARIES})
 target_link_libraries(ekf_localization_node ros_filter ${catkin_LIBRARIES})
 target_link_libraries(ukf_localization_node ros_filter ${catkin_LIBRARIES})
 target_link_libraries(navsat_transform filter_utilities ros_filter_utilities ${catkin_LIBRARIES} ${EIGEN3_LIBRARIES})
 target_link_libraries(navsat_transform_node navsat_transform ${catkin_LIBRARIES})
 
+
```
ayrton04: Please remove

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+};
+
+//! @brief Robot Localization Estimator State
+//!
+//! The Estimator State data structure bundles the state information of the
+//! estimator.
+//!
+struct EstimatorState
+{
+    EstimatorState():
```
ayrton04: My goal is to make any new classes/files added to the package conform to [ROS coding standards](http://wiki.ros.org/CppStyleGuide). To that end, indentations should be two spaces.

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+
+    //! @brief Returns the current buffer size
+    //!
+    //! @return current buffer size
+    //!
+    unsigned int size() const;
+
+  private:
+
+    boost::circular_buffer<EstimatorState> state_buffer_;
```
ayrton04: Sorry to be pedantic, but can we move the variables below the method declarations?

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+    //! @brief When we make a prediction using the transfer function, we
+    //! add this matrix (times deltaT) to the state estimate covariance
+    //! matrix. This models the fact that the prediction is never perfect.
+    //!
+    Eigen::MatrixXd processNoiseCovariance_;
+
+    //! @brief Keeps the state's Roll, Pitch, Yaw angles in the range
+    //! [-pi, pi]
+    //!
+    void wrapStateAngles(EstimatorState state) const
```
ayrton04: Nothing for you to change here, but it looks like this method should be added to one of the utility classes, since we define the same method elsewhere.

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+
+    //! @brief Sets the current internal state of the listener.
+    //!
+    //! @param[in] state - The new state vector to set the internal state to
+    //!
+    void setState(const EstimatorState& state);
+
+    //! @brief Returns the state at a given time
+    //!
+    //! Projects the current state and error matrices forward using a model of
```
ayrton04: This looks like copy-and-paste description code from the `predict` method.

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+    //!
+    //! @param[in] capacity - The new buffer capacity
+    //!
+    void setBufferCapacity(const int capacity);
+
+    //! @brief Returns the buffer capacity
+    //!
+    //! @return buffer capacity
+    //!
+    unsigned int capacity() const;
```
ayrton04: Can we make this `getBufferCapacity` so it matches the set method?

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+
+    //! @brief Returns the current buffer size
+    //!
+    //! @return current buffer size
+    //!
+    unsigned int size() const;
```
ayrton04: Same here: `getBufferSize()` is clearer. Also, it might be good to add a comment about what's different between capacity and size.

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+    void extrapolate(const EstimatorState& boundary_state, const double requested_time, EstimatorState& state_at_req_time) const;
+
+    //! @brief Interpolates the given state to a requested time stamp
+    //!
+    //! @param[in] given_state_1 - last state update before requested time
+    //! @param[in] given_state_2 - next state update after requested time
+    //! @param[in] requested_time - time stamp to extrapolate to
+    //! @param[out] state_at_req_time - predicted state at requested time
+    //!
+    void interpolate(const EstimatorState& given_state_1, const EstimatorState& given_state_2, const double requested_time, EstimatorState& state_at_req_time) const;
```
ayrton04: This looks like it's a lot longer than 120 lines. I'd run roslint on this and correct whatever it says.

include/robot_localization/robot_localization_estimator.h
```diff
@@ -0,0 +1,196 @@
...
+
+    //! @brief When we make a prediction using the transfer function, we
+    //! add this matrix (times deltaT) to the state estimate covariance
+    //! matrix. This models the fact that the prediction is never perfect.
+    //!
+    Eigen::MatrixXd processNoiseCovariance_;
```
ayrton04: It's new code, so I'd use the correct format of `process_noise_covariance_`. Also, I'm a bit confused as to why this variable is here. Is this used in extrapolation?

include/robot_localization/ros_robot_localization_listener.h
```diff
@@ -0,0 +1,162 @@
...
+
+//! @brief RosRobotLocalizationListener class
+//!
+//! This class wraps the RobotLocalizationEstimator. It listens to
+//! topics over which the (filtered) robot state is published (odom
+//! and accel) and pushes them into its instance of the
+//! RobotLocalizationEstimator. It exposes a getState method to
+//! offer the user the estimated state at a requested time. This
+//! class offers the option to run this listener without the need to
+//! to run a separate node. If you do wish to run this functionality
```
ayrton04: Handy

include/robot_localization/ros_robot_localization_listener.h
```diff
@@ -0,0 +1,162 @@
...
+    //!
+    ~RosRobotLocalizationListener();
+
+    //! @brief Get a state from the localization estimator
+    //!
+    //! Requests the predicted state and covariance at a given time
+    //! from the Robot Localization Estimator.
+    //!
+    //! @param[in] time - time of the requested state
+    //! @param[out] frame_id - state at the requested time
```
ayrton04: Fix this comment

include/robot_localization/ros_robot_localization_listener.h
```diff
@@ -0,0 +1,162 @@
...
+    //!
+    ros::NodeHandle nh_p_;
+
+    //! @brief Subscriber to the odometry state topic (output of a filter node)
+    //!
+    message_filters::Subscriber<nav_msgs::Odometry> odom_sub_;
+
+    //! @brief Subscriber to the acceleration state topic (output of a filter node)
+    //!
+    message_filters::Subscriber<geometry_msgs::AccelWithCovarianceStamped> accel_sub_;
```
ayrton04: Looks like an extra line and a missed indentation here

package.xml
```diff
@@ -28,6 +28,7 @@
   <depend>tf2</depend>
   <depend>tf2_geometry_msgs</depend>
   <depend>tf2_ros</depend>
+  <depend>eigen_conversions</depend>
```
ayrton04: Please alphabetize

src/robot_localization_estimator.cpp
```diff
@@ -0,0 +1,330 @@
...
+{
+  return state_buffer_.capacity();
+}
+
+unsigned int RobotLocalizationEstimator::size() const
+{
+  return state_buffer_.size();
+}
+
+void RobotLocalizationEstimator::extrapolate(const EstimatorState& boundary_state,
```
ayrton04, line 178: This is a lot of repeat code from the EKF. Any reason we can't just instantiate an EKF, manually set the process noise in the constructor, set the state and estimate error covariance here, and call predict, then getState? Also, we have two different filters, and I am working on 2D versions of each of them. It might be better to use a parameter that specifies what extrapolation model to use (i.e., EKF or UKF). You can then have a `FilterBase *filter_` member variable that you instantiate.

src/robot_localization_estimator.cpp
```diff
@@ -0,0 +1,330 @@
...
+                                             const EstimatorState& given_state_2,
+                                             const double requested_time,
+                                             EstimatorState& state_at_req_time) const
+{
+  /*
+   * TODO: Right now, we only extrapolate from the last known state before the requested time.
+   * But as the state after the requested time is also known, we may want to perform
+   * interpolation between states.
+   */
+  extrapolate(given_state_1, requested_time, state_at_req_time);
```
ayrton04: I think this method *does* need to interpolate with the states on either side of the requested time. Otherwise, your requested state data is subject to the EKF/UKF motion model, which won't be accurate for every vehicle. Besides, you have the state data, might as well use it.

src/robot_localization_listener_node.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include "robot_localization/ros_robot_localization_listener.h"
+#include "robot_localization/GetState.h"
+
+class RobotLocalizationListenerNode
+{
+public:
+  RobotLocalizationListenerNode()
+  {
+    service_ = n_.advertiseService("get_robot_localization_state", &RobotLocalizationListenerNode::getStateCallback, this);
```
ayrton04: We can probably remove the node name from the service call. Maybe just `get_state`?

src/robot_localization_listener_node.cpp
```diff
@@ -0,0 +1,91 @@
...
+private:
+  RobotLocalization::RosRobotLocalizationListener rll_;
+  ros::NodeHandle n_;
+  ros::ServiceServer service_;
+
+  bool getStateCallback(robot_localization::GetState::Request  &req,
+                        robot_localization::GetState::Response &res)
+  {
+    Eigen::VectorXd state(RobotLocalization::STATE_SIZE);
+    Eigen::MatrixXd covariance(RobotLocalization::STATE_SIZE,RobotLocalization::STATE_SIZE);
```
ayrton04: Maybe add a `using RobotLocalization::STATE_SIZE` or something? Seems awkward to have to use the qualifier everywhere.

src/robot_localization_listener_node.cpp
```diff
@@ -0,0 +1,91 @@
...
+    {
+      res.state[i] = (*(state.data() + i));
+    }
+
+    for (size_t i = 0; i < RobotLocalization::STATE_SIZE*RobotLocalization::STATE_SIZE; i++)
+    {
+      res.covariance[i] = (*(covariance.data() + i));
+    }
+
+    ROS_INFO("Robot Localization Listener Node: Listener responded with state and covariance at the requested time.");
```
ayrton04: Will this get spammy if users are making frequent calls to this service? Might be better for a debug message.

src/ros_robot_localization_listener.cpp
```diff
@@ -0,0 +1,278 @@
...
+    estimator_(0),
+    tf_listener_(tf_buffer_),
+    base_frame_id_("")
+  {
+    int buffer_size;
+    nh_p_.param("buffer_size", buffer_size, 10);
+
+    estimator_.setBufferCapacity(buffer_size);
+
+    sync_.registerCallback(&RosRobotLocalizationListener::odomAndAccelCallback, this);
```
ayrton04: Indentation

src/ros_robot_localization_listener.cpp
```diff
@@ -0,0 +1,278 @@
...
+    if ( base_frame_id_.empty() )
+      base_frame_id_ = odom.child_frame_id;
+
+    // Pose: Position
+    state.state(StateMemberX) = odom.pose.pose.position.x;
+    state.state(StateMemberY) = odom.pose.pose.position.y;
+    state.state(StateMemberZ) = odom.pose.pose.position.z;
+
+    // Pose: Orientation
+    tf2::Quaternion orientation_quat;
```
ayrton04: I definitely have a `quatToRpy` method in one of the utility headers.

src/robot_localization_estimator.cpp
```diff
@@ -0,0 +1,330 @@
...
+    }
+  }
+}
+
+int RobotLocalizationEstimator::getState(const double time, EstimatorState& state) const
+{
+  // If there's nothing in the buffer, there's nothing to give.
+  if ( state_buffer_.size() == 0 )
+  {
+    return -1;
```
ayrton04: No magic numbers, please. Add an enumeration for return codes if you really need them, or just make this a bool and don't worry about passing on the reason for failure.

src/ros_robot_localization_listener.cpp
```diff
@@ -0,0 +1,278 @@
...
+      }
+      else
+      {
+        ROS_ERROR("Ros Robot Localization Listener: The base frame id is not set. Is the child_frame_id in the odom messages set?");
+      }
+
+      return false;
+    }
+
+    if ( estimator_.getState(time, estimator_state) == -2 )
```
ayrton04: Yeah, make some enumerated return codes.

src/ros_robot_localization_listener.cpp
```diff
@@ -0,0 +1,278 @@
...
+    state(StateMemberZ) = target_pose_odom.translation().z();
+
+    Eigen::Vector3d ypr = target_pose_odom.rotation().eulerAngles(2, 1, 0);
+
+    state(StateMemberRoll)  = ypr[2];
+    state(StateMemberPitch) = ypr[1];
+    state(StateMemberYaw)   = ypr[0];
+
+    // Now let's calculate the twist of the target frame
+    Twist base_vel, target_vel_base;
```
ayrton04: One variable declaration per line, please. Most of this will be caught by roslint or just from reading over the style guide.

src/ros_robot_localization_listener.cpp
```diff
@@ -0,0 +1,278 @@
...
+    base_vel.linear = Eigen::Vector3d(estimator_state.state(StateMemberVx),
+                                      estimator_state.state(StateMemberVy),
+                                      estimator_state.state(StateMemberVz));
+    base_vel.angular = Eigen::Vector3d(estimator_state.state(StateMemberVroll),
+                                       estimator_state.state(StateMemberVpitch),
+                                       estimator_state.state(StateMemberVyaw));
+
+    Eigen::Vector3d target_position_base = target_pose_base.translation();
+    target_vel_base.linear = base_vel.linear + base_vel.angular.cross(target_position_base);
```
ayrton04: I need to work this out on paper, but why are we not rotating the linear velocity before applying the cross product of the offset with the angular velocity? Here's how I'm handing it in the rest of the filter:

https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ros_filter.cpp#L2769

ayrton04: Also, see this in the TF library:

https://github.com/ros/geometry/blob/4421e22243a38ba6a130494698b33117d2995a18/tf/src/tf.cpp#L317

ayrton04 (changes requested):
First, thanks for the PR! There are some things I'd request that you change. I think you could re-use code from the EKF/UKF for handling extrapolation, and it would be good to do (linear) interpolation between adjacent states rather than extrapolating from the previous state. Other than that, it's just  matter of a few coding standard violations. Thanks again!

rokusottervanger:
Thanks for the review! I will process your comments as soon as possible, but it may be a week or two before I have the time.

ayrton04:
I take it this is still in progress, right?

src/ros_robot_localization_listener.cpp
```diff
@@ -31,248 +31,359 @@
...
+  state(StateMemberVx) = target_velocity.linear(0);
+  state(StateMemberVy) = target_velocity.linear(1);
+  state(StateMemberVz) = target_velocity.linear(2);
+
+  state(StateMemberVroll) = target_velocity.angular(0);
+  state(StateMemberVpitch) = target_velocity.angular(1);
+  state(StateMemberVyaw) = target_velocity.angular(2);
+
+  // TODO: transform covariance
+  covariance = estimator_state.covariance;
```
rokusottervanger: This covariance transformation is still a todo. A choice needs to be made though, because the rotational uncertainty in combination with translation implies a non-Gaussian uncertainty after transformation, so some approximation is needed. Do you have a suggestion, @ayrton04?

src/robot_localization_estimator.cpp
```diff
@@ -319,9 +202,8 @@ void RobotLocalizationEstimator::interpolate(const EstimatorState& given_state_1
                                              EstimatorState& state_at_req_time) const
 {
   /*
-   * TODO: Right now, we only extrapolate from the last known state before the requested time.
-   * But as the state after the requested time is also known, we may want to perform
-   * interpolation between states.
+   * TODO: Right now, we only extrapolate from the last known state before the requested time. But as the state after
```
rokusottervanger: This todo is still open. The state interpolation can be done using state estimates on both sides of the requested time as constraints, but I have no knowledge about interpolation of the covariance matrices. By the way, interpolation does not take into account the system model from the filter, so extrapolation from the nearest known time might actually be a nicer option.

src/ros_robot_localization_listener.cpp
rokusottervanger: Fixed in 3bcb157b52e4420891528ad23312e280c1e57b0b. The comment still holds, but it is now solved the same way as in [navsat_transform](https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/navsat_transform.cpp#L573)

src/ros_robot_localization_listener.cpp
```diff
@@ -0,0 +1,442 @@
...
+  return;
+}
+
+bool RosRobotLocalizationListener::getState(const double time,
+                                            const std::string& frame_id,
+                                            Eigen::VectorXd& state,
+                                            Eigen::MatrixXd& covariance,
+                                            std::string world_frame_id)
+{
+  EstimatorState estimator_state;
```
rokusottervanger: This is going to need some check for the nature of the requested transform. Right now, requesting the state of "world" would give a reasonable result without complaining, but it would use the robot state to transform between odom and base, and then tf to transform between base and utm. We can either choose to throw an error if base is not a parent of the requested frame id, or only transform using tf if this is the case.

src/ros_robot_localization_listener.cpp
rokusottervanger: No rotation is necessary, because all vectors here are expressed in the base-fixed frame.

src/robot_localization_estimator.cpp
rokusottervanger: This was fixed in 65d256cd4b297d430bb3c540a9eb907e3218d7bc and f24a843aad897e60282a1c3c4862af8743434bb9

src/robot_localization_estimator.cpp
rokusottervanger: I found a method for blending the covariance matrix [here](http://www.agi.com/downloads/resources/white-papers/CovarianceInterpolationFinal.pdf). But so far I haven't been able to find a good way for blending/interpolation of the Euler angles in the state representation, taking into account the wrapped state angles and velocity constriants on both sides. Do you have suggestions for this?

ayrton04:
Hi! Sorry for the lapse in comms. What's the status on this?

rokusottervanger:
I will merge the upstream and resolve any conflicts. 
Last unresolved comment is about interpolation between states and corresponding covariance matrices. However, I currently don't know of a decent way to do both interpolation/blending between states and covariance matrices **and** take into account the system model. AFAICT, A prediction based on the previous state will suffice for all practical cases, and I would like to leave this feature open for future development.

ayrton04:
Sounds good.

ayrton04:
Are you happy for me to merge this at this point?

rokusottervanger:
Yes, that would be great.

ayrton04 (approved):

ayrton04:
TY for this PR!

Issue #337, opened 2017-01-18
Feature: Detect if time stamps of sensor messages are far behind wall time


sharst:
It would be nice if the diagnostics of robot_localization would print a warning if one of the incoming sensors messages lags behind _a lot_. 
Particularly, I had a problem stemming from a clock issue where timestamps of some sensor messages were erroneously set to 0. As the filter can integrate readings from the past, my guess is that they were included but didn't affect the current filter reading in any way. I had a hard time finding out why my sensor readings were seemingly ignored completely.

ayrton04:
It's certainly doable, but I'm not sure what is a good way to define "a lot." I also waffle a bit on how much onus should be on the user to make sure their data streams are correct, but in the end, if the node can detect an issue with sensor data and report it, it should.

sharst:
For starters, I would suggest that any message with Time 0 looks fishy and there could be a warning. I ran into this issue while playing back data from a bag and fusing it with other data I generated on the fly. Due to an error in rqt_bag, the clock wasn't published correctly and the newly generated messages would all have timestamp 0, while the timestamps of the played back messages were all okay.
This is a very special case and also due to a bug, but it seems to me that timing issues might be common when operating on played back data and therefore a warning (e.g. in the diagnostics) might be helpful.

ayrton04:
Agreed.

ayrton04:
I think this was addressed by #340? Let me know if you find otherwise.

PR #340, opened 2017-02-01
adds reset_on_jump_in_time option

SGSSGene:
With this PR I ant to resolve the issue mentioned in #337.

This adds an option to optionally reset the filter if jumps back in time on a single topic are detected.
For this I added a flag to turn this feature on and off. Also I added reset() functions which will reset the queues and filters to the initial uninitialized state.

It also is a good base to cleanup the setPose() function, by using the reset() function inside it.
It also gives the possibility to add a "reset()" service for the whole node.

### DIFF
--- a/doc/state_estimation_nodes.rst
+++ b/doc/state_estimation_nodes.rst
@@ -270,6 +270,11 @@ If *true*, will dynamically scale the ``process_noise_covariance`` based on the 
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 The estimate covariance, commonly denoted *P*, defines the error in the current state estimate. The parameter allows users to set the initial value for the matrix, which will affect how quickly the filter converges. For example, if users set the value at position :math:`[0, 0]` to a very small value, e.g., `1e-12`, and then attempt to fuse measurements of X position with a high variance value for :math:`X`, then the filter will be very slow to "trust" those measurements, and the time required for convergence will increase. Again, users should take care with this parameter. When only fusing velocity data (e.g., no absolute pose information), users will likely *not* want to set the initial covariance values for the absolute pose variables to large numbers. This is because those errors are going to grow without bound (owing to the lack of absolute pose measurements to reduce the error), and starting them with large values will not benefit the state estimate.
 
+~reset_on_time_jump
+^^^^^^^^^^^^^^^^^^^
+If set to *true* and ``ros::Time::isSimTime()`` is *true*, the filter will reset to uninitial state when a jump back in time is noticed on a topic.
+This is useful while playing a bag file and jumping forwards and backwards in time.
+
 Node-specific Parameters
 ------------------------
 The standard and advanced parameters are common to all state estimation nodes in ``robot_localization``. This section details parameters that are unique to their respective state estimation nodes.
--- a/include/robot_localization/filter_base.h
+++ b/include/robot_localization/filter_base.h
@@ -158,6 +158,10 @@ class FilterBase
     //!
     virtual ~FilterBase();
 
+    //! @brief Resets filter to its unintialized state
+    //!
+    void reset();
+
     //! @brief Computes a dynamic process noise covariance matrix using the parameterized state
     //!
     //! This allows us to, e.g., not increase the pose covariance values when the vehicle is not moving
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -117,6 +117,10 @@ template<class T> class RosFilter
     //!
     ~RosFilter();
 
+    //! @brief Resets the filter to its initial state
+    //!
+    void reset();
+
     //! @brief Callback method for receiving all acceleration (IMU) messages
     //! @param[in] msg - The ROS IMU message to take in.
     //! @param[in] callbackData - Relevant static callback data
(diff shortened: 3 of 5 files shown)


ayrton04:
Thanks for the PR! I'll look into it soon.

include/robot_localization/filter_base.h
```diff
@@ -158,6 +158,10 @@ class FilterBase
     //!
     virtual ~FilterBase();
 
+    //! @brief Resets filter to it's unintialized state
```
ayrton04: *its

include/robot_localization/ros_filter.h
```diff
@@ -430,6 +434,13 @@ template<class T> class RosFilter
     //!
     double historyLength_;
 
+    //! @brief Wether to reset the filters when backwards jump in time is detected
```
ayrton04: *Whether

src/filter_base.cpp
```diff
@@ -124,6 +124,11 @@ namespace RobotLocalization
   {
   }
 
+  void FilterBase::reset()
+  {
+    initialized_ = false;
```
ayrton04: I notice this doesn't reset any of the state variables. This class should ideally work independently of the `RosFilter` class, so I would think a reset should reset all the member variables. Perhaps we can move all of the constructor logic into this method and then call `reset()` from the constructor?

src/ros_filter.cpp
```diff
@@ -90,6 +90,39 @@ namespace RobotLocalization
...
+    previousMeasurementCovariances_.clear();
+
+    // Clear the measurement queue.
+    // This prevents us from immediately undoing our reset.
+    while (!measurementQueue_.empty() && ros::ok())
+    {
+      measurementQueue_.pop();
+    }
+
+    ros::getGlobalCallbackQueue()->clear();
```
ayrton04: So I found out through some not-so-fun testing that this method does what it advertises, but if called from within a callback, it may cause the callback *itself* to exit prematurely, which means all the lines after this one may not get called. If possible, can we put this call at the very end of this method?

src/ros_filter.cpp
```diff
@@ -706,6 +743,9 @@ namespace RobotLocalization
     nhLocal_.param("smooth_lagged_data", smoothLaggedData_, false);
     nhLocal_.param("history_length", historyLength_, 0.0);
 
+    // Wether we reset filter on jump back in time
+    nhLocal_.param("reset_on_jump_back_in_time", resetOnJumpBackInTime_, false);
```
ayrton04: This parameter name says what it does, but I wonder if we could think of something less verbose.

ayrton04 (changes requested):
Just a few small things. I think this will make a great addition, thanks!

src/ros_filter.cpp
SGSSGene: I "stole" the idea of calling getGlobalCallbackQueue from `RosFilter<T>::setPoseCallback` it mainly does something similar. a prematurely exit sounds like an error in the code somewhere. So just moving it just to the end might not be a good solution.

src/ros_filter.cpp
SGSSGene: I also didn't had a good idea. I going to through in some ideas, what about:
1. jump_back_reset
2. time_jump_reset
3. reset_on_time_jump
4. reset_time_jump
5. reset_on_loop
6. reset_loop

I really don't know. Pick one :-)

src/ros_filter.cpp
SGSSGene: Hi. I'm trying to figure out why it is exiting itself prematurely. It shouldn't do that at all. How did you figure it out?

ayrton04: http://docs.ros.org/diamondback/api/roscpp/html/classros_1_1CallbackQueue.html#ab960789600b6a6d775249347c234019a

I've also experienced it. If you use this method from within a callback, there's no guarantee the rest of the callback will  execute. I need to fix the other call.

src/ros_filter.cpp
ayrton04: `reset_on_time_jump` seems good.

ayrton04:
BTW, I noticed that there are some handy methods on the `ros::Time` object, specifically the `isSimTime()` method. Might be good to add that condition to your checks for whether we should reset.

ayrton04 (approved):

ayrton04:
I want to get through my PR queue this week. Some of the changes (like this one) I will merge into kinetic, but others (like nodelets) will make more sense if we add them to the lunar branch that I am about to spin up.

doc/state_estimation_nodes.rst
```diff
@@ -270,6 +270,11 @@ If *true*, will dynamically scale the ``process_noise_covariance`` based on the
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 The estimate covariance, commonly denoted *P*, defines the error in the current state estimate. The parameter allows users to set the initial value for the matrix, which will affect how quickly the filter converges. For example, if users set the value at position :math:`[0, 0]` to a very small value, e.g., `1e-12`, and then attempt to fuse measurements of X position with a high variance value for :math:`X`, then the filter will be very slow to "trust" those measurements, and the time required for convergence will increase. Again, users should take care with this parameter. When only fusing velocity data (e.g., no absolute pose information), users will likely *not* want to set the initial covariance values for the absolute pose variables to large numbers. This is because those errors are going to grow without bound (owing to the lack of absolute pose measurements to reduce the error), and starting them with large values will not benefit the state estimate.
 
+~reset_on_time_jump
+^^^^^^^^^^^^^^^^^^^
+If set to *true* and ``ros::Time::isSimTime()`` is *true*, the filter will reset to uninitial state when a jump back in time is noticed on a topic.
```
ayrton04, line 275: Any chance we can change "to uninitial state" to "to its uninitialized state"?

ayrton04:
I'd like to cut a release soon and think this change is worth having in the next release. Any chance you can just tweak the verbiage in the documentation (see my comment)?

ayrton04:
Changed the documentation myself

Issue #285, opened 2016-06-03
Nodelet support


asimay:
enhancement advise: due to large quantity data iput, the module should use "nodelet" to avoid data copy and improve the performance.
Please consider this suggestion. Thanks!

SGSSGene:
@asimay I'm trying to implement nodelets in PR #343 . The implemented test doesn't work. Maybe you have any Idea why?

SGSSGene:
No worries. Found the error. Going to fix everything and hopefully send a PR for nodelets soon.

ayrton04:
Addressed in #343

PR #341, opened 2017-02-08
Preperation for nodelets

SGSSGene:
In long term I want to solve #285 

This diff applies changes to the structure on how the nodes work. The goal
is to have a structure that allows later on to transform them into
nodelets.

To archive nodelet compatible structure, I have to remove *::run() methods and the
while(ros::ok()); loops.

The current run method has actually two parts:
  1. initialization
  2. the main loop

Because of this I decided to move the initialization part into the
constructors of the nodes. The main loop is put into a "periodicUpdate"
function, which is registerd with a ros timer for periodic call.

This is done for navsat_transformation_node and ukf/ekf_localiaztion_nodes.

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -117,6 +117,10 @@ template<class T> class RosFilter
     //!
     ~RosFilter();
 
+    //! @brief Initialize filter
+    //
+    void initialize();
+
     //! @brief Resets the filter to its initial state
     //!
     void reset();
@@ -227,10 +231,6 @@ template<class T> class RosFilter
                       const CallbackData &callbackData,
                       const std::string &targetFrame,
                       const bool imuData);
-
-    //! @brief Main run method
-    //!
-    void run();
 
     //! @brief Callback method for manually setting/resetting the internal pose estimate
     //! @param[in] msg - The ROS stamped pose with covariance message to take in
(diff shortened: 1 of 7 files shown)


ayrton04:
Looks interesting. As it touches the main control loops, however, I will need to do some thorough testing before merging. With 91 forks, I want to be careful that we don't break or alter the behavior for anyone.

SGSSGene:
I understand. I really don't wanted to break anything. This is also the reason why I made a pull request just with the preperation and not with the actual implementation. My expectation are that my changes are do almost not alter the behavior at all.

ayrton04:
Agreed. It looks like everything ought to work, but it'll just take me some time to make sure of that. Thanks for the PR!

ayrton04:
I think I want this and the nodelet change to go into the lunar branch (which I will spin up when the smaller changes going into kinetic are merged).

include/robot_localization/ros_filter.h
```diff
@@ -613,6 +619,38 @@ template<class T> class RosFilter
     // front() refers to the measurement with the earliest timestamp.
     // back() refers to the measurement with the latest timestamp.
     MeasurementHistoryDeque measurementHistory_;
+
+    //! @brief broadcaster of worldTransform tfs
+    //!
+    tf2_ros::TransformBroadcaster worldTransformBroadcaster_;
+
```
ayrton04, line 626: Extra space here.

ayrton04:
OK, I am going to merge this. We can address any issues after-the-fact.

PR #343, opened 2017-02-16
Feature/adding nodelets

SGSSGene:
Work In Progress. Don't merge. Pending on PR #341 

adds nodelet for ekf, ukf and navsat_transform

Because Node and Nodelets handle NodeHandles a little different. I needed to move
NodeHandle and private NodeHandle from RosFilter and NavSatTransform into
the main function and the Nodelet implementation.

I added a test_[ekf|ukf]_localization_nodelet_bag1.test to show that
nodelets works just as nodes do.

Also an example is shown in ekf_nodelet_template.launch on how the
ekf_localization can be used with easy switch between node and nodelet
in a more complex environment.

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -109,7 +109,7 @@ template<class T> class RosFilter
     //! The RosFilter constructor makes sure that anyone using
     //! this template is doing so with the correct object type
     //!
-    explicit RosFilter(std::vector<double> args = std::vector<double>());
+    explicit RosFilter(ros::NodeHandle nh, ros::NodeHandle nh_priv, std::vector<double> args = std::vector<double>());
 
     //! @brief Destructor
     //!
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -117,6 +117,10 @@ template<class T> class RosFilter
     //!
     ~RosFilter();
 
+    //! @brief Initialize filter
+    //
+    void initialize();
+
     //! @brief Resets the filter to its initial state
     //!
     void reset();
@@ -227,10 +231,6 @@ template<class T> class RosFilter
                       const CallbackData &callbackData,
                       const std::string &targetFrame,
                       const bool imuData);
-
-    //! @brief Main run method
-    //!
-    void run();
 
     //! @brief Callback method for manually setting/resetting the internal pose estimate
     //! @param[in] msg - The ROS stamped pose with covariance message to take in
@@ -385,6 +385,12 @@ template<class T> class RosFilter
                       std::vector<int> &updateVector,
                       Eigen::VectorXd &measurement,
                       Eigen::MatrixXd &measurementCovariance);
+
+
+    //! @brief callback function which is called for periodic updates
+    //!
+    void periodicUpdate(const ros::TimerEvent& event);
+
 
     //! @brief tf frame name for the robot's body frame
     //!
@@ -613,6 +619,38 @@ template<class T> class RosFilter
     // front() refers to the measurement with the earliest timestamp.
     // back() refers to the measurement with the latest timestamp.
     MeasurementHistoryDeque measurementHistory_;
+
+    //! @brief broadcaster of worldTransform tfs
+    //!
+    tf2_ros::TransformBroadcaster worldTransformBroadcaster_;
+
+
+    //! @brief position publisher
+    //!
+    ros::Publisher positionPub_;
+
+    //! @brief optional acceleration publisher
+    //!
+    ros::Publisher accelPub_;
+
+    //! @brief optional signaling diagnostic frequency
+    //!
+    std::auto_ptr<diagnostic_updater::HeaderlessTopicDiagnostic> freqDiag_;
+
+    //! @brief last call of periodicUpdate
+    //!
+    ros::Time lastDiagTime_;
+
+    //! @brief timer calling periodicUpdate
+    //!
+    ros::Timer periodicUpdateTimer_;
+
+    //! @brief minimal frequency
+    //!
+    double minFrequency_;
+
+    //! @brief maximal frequency
+    double maxFrequency_;
 };
 
 }  // namespace RobotLocalization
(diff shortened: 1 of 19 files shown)


include/robot_localization/ros_filter.h
```diff
@@ -594,6 +600,41 @@ template<class T> class RosFilter
     // front() refers to the measurement with the earliest timestamp.
     // back() refers to the measurement with the latest timestamp.
     MeasurementHistoryDeque measurementHistory_;
+
+    //! @brief broadcaster of worldTransform tfs
+    //!
+    tf2_ros::TransformBroadcaster worldTransformBroadcaster_;
+
```
ayrton04, line 626: Extra space

include/robot_localization/ros_filter.h
```diff
@@ -594,6 +600,41 @@ template<class T> class RosFilter
...
+    //! @brief last call of periodicUpdate
+    //!
+    ros::Time lastDiagTime_;
+
+    //! @brief timer calling periodicUpdate
+    //!
+    ros::Timer periodicUpdateTimer_;
+
+	//! @brief minimal frequency
```
ayrton04: Indentation

include/robot_localization/ros_filter.h
```diff
@@ -594,6 +600,41 @@ template<class T> class RosFilter
...
+	//!
+    double minFrequency_;
+
+    //! @brief maximal frequency
+    double maxFrequency_;
+
```
ayrton04: Extra spaces. Might want to run roslint on the whole thing.

package.xml
```diff
@@ -32,15 +32,18 @@
   <build_depend>message_generation</build_depend>
   <build_depend>python-catkin-pkg</build_depend>
   <build_depend>roslint</build_depend>
+  <build_depend>nodelet</build_depend>
```
ayrton04, line 36: Please alphabetize

src/ekf_localization_node.cpp
```diff
@@ -38,9 +38,12 @@ int main(int argc, char **argv)
...
+  ros::NodeHandle nh;
+  ros::NodeHandle nh_priv("~");
 
-  ekf.run();
+  RobotLocalization::RosEkf ekf(nh, nh_priv);
+  ekf.initialize();
+  ros::spin();
 
-  return 0;
+  return EXIT_SUCCESS;
```
ayrton04, line 48: What defines this?

src/ekf_localization_nodelet.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "robot_localization/ros_filter_types.h"
+
+#include <ros/ros.h>
+#include <nodelet/nodelet.h>
+
+namespace RobotLocalization
+{
+  class EkfNodelet : public nodelet::Nodelet
+  {
+  std::auto_ptr<RosEkf> ekf;
```
ayrton04: Even though this is supported in Kinetic, I'd like all of these changes to not require C++11, as it means I'll have to only allow this in Kinetic, and will have to maintain Indigo separately. If we have no choice, then we'll deal with it, but I'd prefer to support every ROS distribution that's currently not EOL, if possible.

src/ekf_localization_nodelet.cpp
```diff
@@ -0,0 +1,58 @@
...
+  public:
+    virtual void onInit() {
```
ayrton04: No cuddled braces. Please see http://wiki.ros.org/CppStyleGuide. I realize the variables  don't adhere to standards, but I try to keep everything else that way.

src/ekf_localization_nodelet.cpp
```diff
@@ -0,0 +1,58 @@
...
+      ros::NodeHandle nh      = getNodeHandle();
+      ros::NodeHandle nh_priv = getPrivateNodeHandle();
+
+      ekf.reset(new RosEkf(nh, nh_priv));
+      ekf->initialize();
+    }
+  };
+}
+
+#include <pluginlib/class_list_macros.h>
```
ayrton04: Can this include go at the top?

src/navsat_transform_nodelet.cpp
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2014, 2015, 2016, Charles River Analytics, Inc.
```
ayrton04: Don't need to make this CRA's copyright.

ayrton04 (changes requested):
I'll likely have more, and I'm guessing some of these changes are in the other PR, but there are a few things to change.

src/ekf_localization_node.cpp
SGSSGene: This is standard c/c++: http://en.cppreference.com/w/cpp/utility/program/EXIT_status

src/navsat_transform_nodelet.cpp
SGSSGene: Good point :-). This will take some time to clarify. Not sure which license I have to apply.

ayrton04: The package license is BSD, so we don't need to change that verbiage, but the copyright holder can be you or your organization. Note that I am not a copyright lawyer, but I'm pretty sure that's how it works. :)

src/ekf_localization_node.cpp
ayrton04: Right, but it needs you to `#include <cstdlib>`.

src/ekf_localization_nodelet.cpp
SGSSGene: std::auto_ptr is c++03. It is not my favorite feature, but seemed better then a naked pointer.
In c++11 you would use std::unique_ptr instead.

bochen87:
@ayrton04 Any reason this is not in kinetic?

ayrton04:
It represents a pretty major change to the package, and was added well into Kinetic's lifecycle. I don't strictly adhere to the standards that OSRF suggests for managing changes within and across ROS distros, but this certainly seemed like a significant enough change that releasing it in Kinetic seemed questionable.

bochen87:
ok, thanks for your explanation. Was just wondering, as it does seem to be a very useful feature!

ayrton04:
If you don't mind working from source, the Lunar and Melodic branches will build just fine in Kinetic.

bochen87:
ok, thanks for the hint!

Issue #321, opened 2016-11-10
Use true course in navsat transform node


reinzor:
Some GPS sensors also output a 'true course' which is used by the nmea navsat driver to calculate a velocity vector:

https://github.com/ros-drivers/nmea_navsat_driver/blob/jade-devel/src/libnmea_navsat_driver/driver.py#L159

This information could also be used in the navsat transform node in case an IMU is not available. I would like to implement this functionality but this implies I think a lot of changes in the code. 

My approach would be adding a TwistStamped subscriber of which the orientation is and velocity is extracted. This could be merged in the odom message.

What do you think? 

Thanks,

Rein

reinzor:
Or is this where we should use the wait_for_datum functionality for and use only position updates?

ayrton04:
I'd be tempted to request that `nmea_navsat_driver` (optionally) output an IMU message with that heading.

reinzor:
I am now calling a /datum `setDatum` service with both information as an input:

```
    q = transformations.quaternion_from_euler(0, 0, atan2(TWIST.twist.linear.y, TWIST.twist.linear.x))

    r = SetDatumRequest()
    r.geo_pose.position.latitude = GPS.latitude
    r.geo_pose.position.longitude = GPS.longitude
    r.geo_pose.position.altitude = GPS.altitude
    r.geo_pose.orientation = Quaternion(x=q[0], y=q[1], z=q[2], w=q[3])

    rospy.loginfo("Calling setDatum")
    try:
        resp1 = set_datum(r)

    except rospy.ServiceException, e:
        rospy.logerr("Service call failed: %s"%e)
```

Seems also to be a nice solution, should we add some documentation about this ..?

ayrton04:
How does it behave when the GPS is still? Would it make more sense to add a check on the velocity magnitude?

In any case, sure, that seems like something worthwhile to put in the documentation.

reinzor:
I would say that is up to the user. I can make a python scripts and add this to the repo if desired. I have to add rospy to the deps though.

ayrton04:
Nah, I think the documentation is sufficient. Thanks!

ayrton04:
What's the status on this? No rush on my end, just want to know whether to keep this open.

PR #346, opened 2017-02-21
Added documentation on setting datum

reinzor:
Should close issue #321

### DIFF
--- a/doc/integrating_gps.rst
+++ b/doc/integrating_gps.rst
@@ -67,7 +67,7 @@ IMU Data
 
 Users should make sure their IMUs conform to `REP-105 <http://www.ros.org/reps/rep-0105.html>`_. In particular, check that the signs of your orientation angles increase in the right direction. In addition, users should look up the `magnetic declination <http://www.ngdc.noaa.gov/geomag-web/#declination>`_ for their robot's operating area, convert it to radians, and then use that value for the ``magnetic_declination_radians`` parameter.
 
-This information is only relevant if the user is not manually specifying the origin via the ``datum`` parameter or the ``set_datum`` service.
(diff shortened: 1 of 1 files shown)


doc/integrating_gps.rst
```diff
@@ -60,16 +60,14 @@ The excellent `nmea_navsat_driver <http://wiki.ros.org/nmea_navsat_driver>`_ pac
...
-This information is only relevant if the user is not manually specifying the origin via the ``datum`` parameter or the ``set_datum`` service.
+This information is only relevant if the user is not manually specifying the origin via the ``datum`` parameter or the ``set_datum`` service. Note that a ``datum`` is required in order for the ``navsat_transform_node`` to work. This datum will 'null' the GPS sensor by storing the initial pose (latitude, longitude and heading) and using this initial pose for each newly incoming GPS message, of which this initial pose is subtracted in order to create a ``nav_msgs/Odometry``. From this IMU message, only the heading is used; this heading could also be derived from a GPS sensor that outputs a UTM velocity vector, such as the nmea driver in the `nmea_navsat_driver <http://wiki.ros.org/nmea_navsat_driver>`_ package. The published ``TwistStamped`` on the ``fix_velocity`` topic could be used to derive the initial heading by taking the ``atan2`` of the two vector components, if the magnitude of the vector is large enough. This way, no additional IMU is required in order to obtain the initial heading to set the datum of our ``navsat_transform_node``.
```
ayrton04, line 70: I'm not crazy about this bit, for the following reasons:

1. This section is about IMU data, but this message moves into a general discussion of how `n_t_n` works.
2. I already discuss the datum under "required inputs":
image
Is there  anything I can add to that section that would make it clearer? I think this information belongs there, if you don't think it's already there. I'd also want to adjust the verbiage in that case.
1. The bit about using a GPS to obtain a heading is good to have, but we should add it as a new paragraph, and say something like "An alternative means of generating an earth-referenced heading is..."

PR #347, opened 2017-02-24
Feature/optionally publish utm frame as parent

reinzor:
In order to create a transformation tree with multiple robots, a common
parent frame is required. So far, it was only possible to publish "utm"
as child frame of the odometry frame. Adding this feature, the user can
specify whether to publish the "utm" transform as parent or child of the
odometry frame.

### DIFF
--- a/doc/navsat_transform_node.rst
+++ b/doc/navsat_transform_node.rst
@@ -63,4 +63,4 @@ Published Topics
 
 Published Transforms
 ====================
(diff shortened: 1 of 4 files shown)


ayrton04:
This is a simple change and looks like it was implemented just fine, but ideally, we'd implement the "earth" frame that was somewhat-recently added to [REP-105](http://www.ros.org/reps/rep-0105.html). It's meant to be the common parent of all the map frames, as you suggested.

reinzor:
Would you like me to rename "utm" to "earth"?

ayrton04:
Kicking the build server...

reinzor:
Hi @ayrton04 , can this be merged?

ayrton04:
It can, yes.

Issue #328, opened 2016-11-28
Fix issue with basic IO tests


ayrton04:
This happens occasionally, and it's annoying:

http://build.ros.org/job/Kpr__robot_localization__ubuntu_xenial_amd64/12/testReport/junit/(root)/InterfacesTest/ImuTwistBasicIO/

SGSSGene:
This seems to be an conversion issue.
Simple Code to reproduce:

```
  double r = 0.;
  double p = -M_PI/2.0 -1e-10;
  double y = 0.
  Matrix3x3 mat;
  mat.setRPY(r, p, y);
  mat.getRPY(r, p, y);
  // r, p, y are now as expected roughly (0, -M_PI/2, 0)
```
if you change p just a little:
```
  double r = 0.;
  double p = -M_PI/2.0 + 1e-10; // this time add 1e-10 instead of substracting
  double y = 0.
  Matrix3x3 mat;
  mat.setRPY(r, p, y);
  mat.getRPY(r, p, y);
  // r, p, y are now (M_PI, -M_PI/2, M_PI)
```

The second case always happen if there is some delay in the communication between the nodes. If you extend the the for loop to 55 instead of 50 iteration, it will trigger the bug much more often: https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/test/test_ukf_localization_node_interfaces.cpp#L642

I'm not sure how to fix this unit test in a clean way.

ayrton04:
Thanks for the PR!

PR #348, opened 2017-03-01
fixes ukf IMUTwistBasicIO test

SGSSGene:
See Issue #328 
There is an issue with the way how rotation is tested isn't
reliable. Because of Gimbel lock, similar rotation can have huge changes
in euler angle representation. This was an issue in IMUTwistBasicIO.

Instead of comparing Euler Angles, quaternion are being compared, making
the test much more reliable.


ayrton04 (approved):
The fun of working with Euler angles! Thanks.

PR #349, opened 2017-03-01
cleanup 'catkin_lint -W2' and 'catkin_make roslint_robot_localization'

SGSSGene:
this commit removes all complaints from 'catkin_make
roslint_robot_localization' and most of 'catkin_lint -W2'

The only complaint left from 'catkin_lint -W2' is about
- robot_localization: notice: target name 'navsat_transform_node' might
not be sufficiently unique'

This can't be changed without breaking compatibility.

This commit shouldn't change any behavior at all.


SGSSGene:
Any interest in this PR?

ayrton04:
Apologies, I am in the midst of a very busy time at work, so I haven't had the cycles to review any of these in depth. Thank you very much for submitting it! I will get to it as soon as I have time.

SGSSGene:
Please don't apologies. I just wanted to make sure it wasn't overseen.

package.xml
```diff
@@ -2,7 +2,7 @@
 <package format="2">
   <name>robot_localization</name>
   <version>2.3.1</version>
-  <description>The robot_localization package provides nonlinear state estimation through sensor fusion of an abritrary number of sensors.</description>
+  <description>Nonlinear state estimation through sensor fusion of an abritrary number of sensors.</description>
```
ayrton04: If we're shortening this, I'd prefer starting with a verb, e.g., "Provides nonlinear...".

ayrton04:
The remaining warning is strange. BTW, can you see what happens if you add some `roslint_cpp` targets and `roslint_add_test()` to the [CMakeLists](http://wiki.ros.org/roslint)? It would be good for linting to be added as a test so that we can't check in code that violates any rules. However, if the warning you mentioned would cause the test to fail, I don't think I'd want it.

ayrton04:
(You can skip adding specific `roslint_cpp` targets and just keep the default as-is, as well).

package.xml
SGSSGene: I shorten it, because one of the linter was complaining the description is starting with a boilerplate or something similar. I'll check and add the "Provides nonlinear..." if it doesn't complain.

ayrton04: It won't. However, when the linter complains, you can make note of the issue category (usually given in brackets at the end of the error when it is reported) and [filter out](http://wiki.ros.org/roslint#Tips) that warning in the CMakeLists.txt.

SGSSGene:
I rebased this to the current kinetic-devel. I also adjusted the package name.
The remaining warning is a "catkin lint -W2" warning, not roslint. roslint runs without errors.
```
  ~/ws/src/robot_localization $ catkin lint -W2
  robot_localization: notice: target name 'navsat_transform_node' might not be sufficiently unique
  catkin_lint: checked 1 packages and found 1 problems
```

ayrton04 (approved):

PR #350, opened 2017-03-02
fix(logging): Add 5.0 seconds delay and throttle if lookup fails

reinzor:
This will often go wrong since the odometry localization node is often
not initialized before the map localization node. This results in errors
on startup. Adding 5.0 delay gives the estimator time to properly
initialize.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1758,7 +1758,8 @@ namespace RobotLocalization
             }
             catch(...)
             {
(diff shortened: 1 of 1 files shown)


ayrton04:
Whoops, I think you may have increased the scope of this PR. :)

reinzor:
Hmm, thank you Travis, I forgot it merges the target branch while doing its CI. I will update the PR in a bit.

reinzor:
Fixed :), sorry for the inconvenience

ayrton04 (approved):
Good call.

ayrton04:
Thanks for the PR!

PR #351, opened 2017-03-03
Parameterize gravity

gerkey:
We're using this package for estimation in an environment where we don't experience Earth gravity. This PR makes the acceleration due to gravity configurable with a parameter.


ayrton04 (approved):
LGTM

ayrton04:
Thanks for this (and for providing the documentation)!

PR #353, opened 2017-03-13
Fixing xmlrpcpp dependency

ayrton04:


ayrton04:
This can go into kinetic, but doesn't play well in Indigo. Looks like the separate branches will finally matter.

PR #355, opened 2017-03-24
Fix acceleration and deceleration gains default logic

efernandez:
Note that resize on a vector of size S doesn't change the values of the
current S elements.

See: http://en.cppreference.com/w/cpp/container/vector/resize

As a result, when the gains weren't set, the `resize` in https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ros_filter.cpp#L784 was doing nothing because the vector is initialized to `TWIST_SIZE` by default here on construction: https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ros_filter.cpp#L737. But the default value is `0.0` instead of `1.0`. This makes all accelerations limited to `0.0` when this code is run because `gain` is always `0.0` when no gain params are given and the wrong defaults are used: https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/include/robot_localization/filter_base.h#L375

This PR fixes the logic that sets the default for the gains, as well as some typos and the logic for the acceleration limits default when the size is wrong: https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ros_filter.cpp#L774.

Note that I didn't use `std::min` in `std::fill_n` because `TWIST_SIZE` is `int` and `size()` returns `size_t`, so `std::min` (template) is not defined for mixed input types. The code is simpler if we just fill all elements for the given vector size; it might be larger than `TWIST_SIZE` but I don't think we need to worry about computational speed on this part of the code.

The logic with `useAccelLimits` was also wrong because it remains `false` when the acceleration limits are provided with the proper size. I removed it because IMHO the code is cleaner w/o it.

Locally, I made the `RF_DEBUG` message that prints the param values a `ROS_ERROR_STREAM` to check that different/wrong sizes are handled well, and default values are now vector of size `TWIST_SIZE` with all elements equal to `1.0`.

@ayrton04

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -733,10 +733,10 @@ namespace RobotLocalization
     bool stampedControl = false;
     double controlTimeout = sensorTimeout;
     std::vector<int> controlUpdateVector(TWIST_SIZE, 0);
-    std::vector<double> accelerationLimits(TWIST_SIZE, 0.0);
-    std::vector<double> accelerationGains(TWIST_SIZE, 0.0);
-    std::vector<double> decelerationLimits(TWIST_SIZE, 0.0);
-    std::vector<double> decelerationGains(TWIST_SIZE, 0.0);
+    std::vector<double> accelerationLimits(TWIST_SIZE, 1.0);
+    std::vector<double> accelerationGains(TWIST_SIZE, 1.0);
+    std::vector<double> decelerationLimits(TWIST_SIZE, 1.0);
+    std::vector<double> decelerationGains(TWIST_SIZE, 1.0);
 
     nhLocal_.param("use_control", useControl_, false);
     nhLocal_.param("stamped_control", stampedControl, false);
@@ -771,21 +771,20 @@ namespace RobotLocalization
       else
       {
         ROS_WARN_STREAM("use_control is set to true, but acceleration_limits is missing. Will use default values.");
-        accelerationLimits.resize(TWIST_SIZE, 1.0);
       }
 
-      accelerationGains.resize(TWIST_SIZE, 1.0);
       if(nhLocal_.getParam("acceleration_gains", accelerationGains))
       {
-        if(accelerationGains.size() != TWIST_SIZE)
+        const int size = accelerationGains.size();
+        if(size != TWIST_SIZE)
         {
           ROS_ERROR_STREAM("Acceleration gain configuration must be of size " << TWIST_SIZE <<
-            ". Provided config was of size " << accelerationGains.size() << ". All gains will be assumed to be 1.");
+            ". Provided config was of size " << size << ". All gains will be assumed to be 1.");
+          std::fill_n(accelerationGains.begin(), std::min(size, TWIST_SIZE), 1.0);
           accelerationGains.resize(TWIST_SIZE, 1.0);
         }
       }
 
-      bool useAccelLimits = false;
       if(nhLocal_.getParam("deceleration_limits", decelerationLimits))
       {
         if(decelerationLimits.size() != TWIST_SIZE)
@@ -800,21 +799,23 @@ namespace RobotLocalization
         ROS_INFO_STREAM("use_control is set to true, but no deceleration_limits specified. Will use acceleration "
           "limits.");
         decelerationLimits = accelerationLimits;
-        useAccelLimits = true;
       }
 
-      decelerationGains.resize(TWIST_SIZE, 1.0);
       if(nhLocal_.getParam("deceleration_gains", decelerationGains))
       {
-        if(decelerationGains.size() != TWIST_SIZE)
+        const int size = decelerationGains.size();
+        if(size != TWIST_SIZE)
         {
-          ROS_ERROR_STREAM("Acceleration gain configuration must be of size " << TWIST_SIZE <<
-            ". Provided config was of size " << decelerationLimits.size() << ". All gains will be assumed to be 1.");
+          ROS_ERROR_STREAM("Deceleration gain configuration must be of size " << TWIST_SIZE <<
+            ". Provided config was of size " << size << ". All gains will be assumed to be 1.");
+          std::fill_n(decelerationGains.begin(), std::min(size, TWIST_SIZE), 1.0);
+          decelerationGains.resize(TWIST_SIZE, 1.0);
         }
       }
-      else if(useAccelLimits)
+      else
       {
-        ROS_INFO_STREAM("Using acceleration gains for deceleration");
+        ROS_INFO_STREAM("use_control is set to true, but no deceleration_gains specified. Will use acceleration "
+          "gains.");
         decelerationGains = accelerationGains;
       }
     }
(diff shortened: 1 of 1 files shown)


ayrton04:
Great, thanks @efernandez!

src/ros_filter.cpp
```diff
@@ -733,10 +733,10 @@ namespace RobotLocalization
     bool stampedControl = false;
     double controlTimeout = sensorTimeout;
     std::vector<int> controlUpdateVector(TWIST_SIZE, 0);
-    std::vector<double> accelerationLimits(TWIST_SIZE, 0.0);
-    std::vector<double> accelerationGains(TWIST_SIZE, 0.0);
-    std::vector<double> decelerationLimits(TWIST_SIZE, 0.0);
-    std::vector<double> decelerationGains(TWIST_SIZE, 0.0);
+    std::vector<double> accelerationLimits(TWIST_SIZE, 1.0);
```
efernandez, line 736: Set defaults to `1.0`, not `0.0`, according with https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/params/ekf_template.yaml#L200 and the WARN messages from below, when the size is small, or the param is omitted.

src/ros_filter.cpp
```diff
@@ -771,21 +771,20 @@ namespace RobotLocalization
       else
       {
         ROS_WARN_STREAM("use_control is set to true, but acceleration_limits is missing. Will use default values.");
-        accelerationLimits.resize(TWIST_SIZE, 1.0);
```
efernandez, line 774: This was doing nothing, so the default was all `0.0`. If the intention is to have all `0.0`, that can be changed on the construction before. The `resize` does nothing here. Same for the `resize` below to the gains.

I've change the default from `0.0` to `1.0` because if the limits are `0.0` by default, then the gain defaulting to `1.0` won't do anything. IMHO it's more intuitive/expectable to have the limits default to `1.0`.

src/ros_filter.cpp
```diff
@@ -771,21 +771,20 @@ namespace RobotLocalization
...
       if(nhLocal_.getParam("acceleration_gains", accelerationGains))
       {
-        if(accelerationGains.size() != TWIST_SIZE)
+        const size_t size = accelerationGains.size();
+        if(size != TWIST_SIZE)
         {
           ROS_ERROR_STREAM("Acceleration gain configuration must be of size " << TWIST_SIZE <<
-            ". Provided config was of size " << accelerationGains.size() << ". All gains will be assumed to be 1.");
+            ". Provided config was of size " << size << ". All gains will be assumed to be 1.");
+          std::fill_n(accelerationGains.begin(), size, 1.0);
```
efernandez: We need to fill with `1.0` the all the elements of the vector given on the param server, so the ERROR message is correct.

If you have concerns with filling in vectors of a very large size, I can pass `std::min(TWIST_SIZE, static_cast<int>(size))` instead of `size`.

src/ros_filter.cpp
```diff
@@ -771,21 +771,20 @@ namespace RobotLocalization
...
           accelerationGains.resize(TWIST_SIZE, 1.0);
         }
       }
 
-      bool useAccelLimits = false;
```
efernandez, line 788: As I mentioned on the description, if `deceleration_limits` is given, we don't set `useAccelLimits` to `true`. Either way, IMHO it's more consistent with the rest of the logic to remove this variable and always set `decelerationGains = accelerationGains` when the `deceleration_gains` param is omitted.

efernandez (commented):
Changes are minimal, but the diff seems to have a lot of changes. So I've added a few comments to help with the review.

ayrton04:
Appreciated.

src/ros_filter.cpp
ayrton04: :+1:

src/ros_filter.cpp
ayrton04: That, or we force it to be of size TWIST_SIZE and then fill with 1.0.

src/ros_filter.cpp
ayrton04: Whoops, yeah, good call.

src/ros_filter.cpp
```diff
@@ -800,21 +799,23 @@ namespace RobotLocalization
...
-        if(decelerationGains.size() != TWIST_SIZE)
+        const size_t size = decelerationGains.size();
+        if(size != TWIST_SIZE)
         {
-          ROS_ERROR_STREAM("Acceleration gain configuration must be of size " << TWIST_SIZE <<
-            ". Provided config was of size " << decelerationLimits.size() << ". All gains will be assumed to be 1.");
+          ROS_ERROR_STREAM("Deceleration gain configuration must be of size " << TWIST_SIZE <<
+            ". Provided config was of size " << size << ". All gains will be assumed to be 1.");
+          std::fill_n(decelerationGains.begin(), size, 1.0);
+          decelerationGains.resize(TWIST_SIZE, 1.0);
```
ayrton04: Would it make sense to reverse lines 811 and 812? That way, we avoid filling a very large vector if `decelerationGains.size()` is massive (I can't see this being the case, but it can't hurt to be safe). Also, your logic here is closer to what I would have expected for your changes on line [783](https://github.com/cra-ros-pkg/robot_localization/pull/355/files#diff-dd5f9ec904b27c736579e4457a23a746R783), but you aren't doing the `resize` there. I may be missing your intent, though.

src/ros_filter.cpp
```diff
@@ -841,9 +842,9 @@ namespace RobotLocalization
              "\ncontrol_config is " << controlUpdateVector <<
              "\ncontrol_timeout is " << controlTimeout <<
              "\nacceleration_limits are " << accelerationLimits <<
-             "\nacceleration_gains are " << accelerationLimits <<
+             "\nacceleration_gains are " << accelerationGains <<
```
ayrton04, line 845: Whoops, thanks.

ayrton04 (commented):
Just one small difference between the way you're handling the acceleration gains and the deceleration gains. Other than that, LGTM.

src/ros_filter.cpp
efernandez: Done. I've forced `size` to be `int`, so the instruction that fills in the vector with ones is cleaner:

`std::fill_n(accelerationGains.begin(), std::min(size, TWIST_SIZE), 1.0);`

I mean I don't need the verbose `static_cast<int>`.

This way we're actually doing the minimum number of operations:

- We only fill with ones the `size` elements of the wrong size vector given by the user, but only up to `TWIST_SIZE`
- And we resize to `TWIST_SIZE` filling with ones only for new elements; and in the case of reducing the size, the discarded elements are never filled in with ones in the previous step.

The order of the `fill_n` and `resize` instructions can be swapped, but doesn't make any difference.

src/ros_filter.cpp
efernandez: Changed as for `accelerationGains`; see new comment above.

efernandez:
@ayrton04 I've updated the `fill_n` for the acceleration and deceleration gains logic when the user input vector has a wrong size. Now it'd be safe even for a massive size vector, if that ever happens.

I honestly don't see the difference you're mentioning regarding the logic that handles acc. and dec. gains. Maybe I'm missing something, or you actually mean the gains vs. limits logic. But that logic remains close to the original one, if not the same.

ayrton04:
Please ignore, I missed [line 784](https://github.com/cra-ros-pkg/robot_localization/pull/355/files#diff-dd5f9ec904b27c736579e4457a23a746R784), which wasn't modified.

LGTM. Thanks again.

PR #376, opened 2017-09-21
Install params directory.

mikepurvis:
Addresses concern raised here about missing params directory:

https://answers.ros.org/question/271213/how-can-i-install-robot_localization/


ayrton04:
D'oh. Thanks Mike!

PR #378, opened 2017-10-03
ensure all raw sensor input orientations are normalized even if messages are not

SteveMacenski:
I've ran into issues where IMU data may or may not always be correctly normalized for TF's liking. Robot Pose EKF deals with this issue gracefully via tf::quaternionMsgToTF which will normalize the orientation quaternion before throwing it into the EKF. I have added the normalization call between Quaternion data unpack and utilization.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2318,7 +2318,11 @@ namespace RobotLocalization
         else
         {
           tf2::fromMsg(msg->orientation, curAttitude);
-          curAttitude.normalize();
+          if (fabs(curAttitude.length() - 1.0) > 0.01)
+          {
+            ROS_WARN_ONCE("An input was not normalized, this should NOT happen, but will normalize.");
+            curAttitude.normalize();
+          }
         }
         trans.setRotation(curAttitude);
         tf2::Vector3 rotNorm = trans.getBasis().inverse() * normAcc;
@@ -2477,7 +2481,11 @@ namespace RobotLocalization
     else
     {
       tf2::fromMsg(msg->pose.pose.orientation, orientation);
-      orientation.normalize();
+      if (fabs(orientation.length() - 1.0) > 0.01)
+      {
+        ROS_WARN_ONCE("An input was not normalized, this should NOT happen, but will normalize.");
+        orientation.normalize();
+      }
     }
 
     // Fill out the orientation data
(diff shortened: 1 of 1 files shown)


ayrton04:
Thanks for the PR!

Is it possible for us to detect this condition, and spit one a `ROS_WARN_ONCE` message before we correct it? I feel like sensor data ought to conform to standards, and while I'm happy to correct it within the package, it could speak to other issues with the sensor or sensor drive that the user ought to know about and address.

SteveMacenski:
That's a very thoughtful suggestion. I will of course add it this evening. 

Separate of this PR, I was curious if you had any thoughts on [this ticket in ros answers](https://answers.ros.org/question/272226/robot_localization-ouputs-nans-when-imuodom-messages-not-naning-works-in-robotposeekf/).  I found this minor shortcoming in robot_localization while trying to fight this issue.

ayrton04:
I've answered your question. Let me know how it behaves once you correct the IMU messages.

src/ros_filter.cpp
```diff
@@ -2318,6 +2318,11 @@ namespace RobotLocalization
         else
         {
           tf2::fromMsg(msg->orientation, curAttitude);
+          if (fabs(curAttitude.length() - 1.0) > 0.01)
```
ayrton04, line 2321: Do we know what kind of tolerance for normalization we require? The PR job passed, so this isn't affecting any of the bags in the test data set, which means that either their quaternion lengths were within this tolerance, or the normalization doesn't have a strong effect (I'm assuming the former).

SteveMacenski:
Thanks for the feedback. I am looking into it next week; if it turns out to be the case, I will try to build infrastructure for robot_localization to error to the user the cause if its inverting. I've seen a number of ros answers tickets that could benefit from it as well.

I added the ros warn once before normalizations. Let me know if you want me to do anything else before merging the PR.

ayrton04:
I am happy to merge this PR, but I am not sure how I will feel about a PR that checks for or corrects for bad covariance matrices in the input. If I'm just checking diagonal values for being non-negative or zero, that's one thing, but checking the determinant of every measurement covariance matrix is going to have an impact on performance. At some point, I think the onus needs to be on the user to make sure that the data that they are feeding to the filter is valid.

ayrton04:
Thanks again for the PR!

SteveMacenski:
My thought would have been to check a value in the covariance matrix before inversion and after and see if it turns from a number to a NaN and spit out an error indicating to the user where it happened and why. At least that way people can make up their own decision as to what to do. Is that more in line with something you would accept?

ayrton04:
That could work. It's this line:

https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ekf.cpp#L160

PR #379, opened 2017-10-11
Add enabling service

miquelmassot:
This PR adds an optional parameter to disable the filter, and allows it to be enabled using a service call.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -887,7 +887,10 @@ namespace RobotLocalization
 
     // Check if the filter should start or not
     nhLocal_.param("disabled_at_startup", disabledAtStartup_, false);
-    if (!disabledAtStartup_) enabled_ = true;
+    if (!disabledAtStartup_)
+    {
+      enabled_ = true;
+    }
 
 
     // Debugging writes to file
@@ -1777,10 +1780,14 @@ namespace RobotLocalization
     ros::Rate loop_rate(frequency_);
 
     // Wait for the filter to be enabled
-    while (!enabled_ && ros::ok()) {
+    while (!enabled_ && ros::ok())
+    {
       ROS_WARN_STREAM_ONCE("[" << ros::this_node::getName() << ":] This filter is disabled. To enable it call the service " << ros::this_node::getName() << "/enable");
       ros::spinOnce();
-      if (enabled_) break;
+      if (enabled_)
+      {
+        break;
+      }
     }
 
     while (ros::ok())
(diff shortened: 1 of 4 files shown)


src/ros_filter.cpp
```diff
@@ -1957,6 +1974,42 @@ namespace RobotLocalization
   }
 
   template<typename T>
+  bool RosFilter<T>::enableFilterSrvCallback(std_srvs::Empty::Request& request,
```
ayrton04: I have to imagine that this is warning you when you build, no? I would think that you can get rid of the `request` parameter name, since it's unused (same goes for the declaration).

src/ros_filter.cpp
```diff
@@ -1957,6 +1974,42 @@ namespace RobotLocalization
...
+      ROS_WARN_STREAM("[" << ros::this_node::getName() << ":] Asking for enabling filter service, but the filter was already enabled! Use param disabled_at_startup.");
+    } else {
+      ROS_INFO_STREAM("[" << ros::this_node::getName() << ":] Enabling filter...");
+      enabled_ = true;
+    }
+    return true;
+  }
+
+  template<typename T>
+  std::string RosFilter<T>::tfFailureReasonString(const tf2_ros::FilterFailureReason reason)
```
ayrton04: Thanks for implementing this method. Not sure how that got added to the header and never defined. However, as it's not used, I think I'm more inclined to remote the method from the header.

src/ros_filter.cpp
```diff
@@ -883,6 +885,11 @@ namespace RobotLocalization
       }
     }
 
+    // Check if the filter should start or not
+    nhLocal_.param("disabled_at_startup", disabledAtStartup_, false);
+    if (!disabledAtStartup_) enabled_ = true;
```
ayrton04: Nitpick: would you mind making sure that (a) we don't nest braces (like in lines 1780, 1981, etc.) and (b) we don't do in-line if statements? As much as this package violates ROS coding standards, I do try in general to follow the same standard within the package.

ayrton04 (commented):
Thanks very much for the PR! Just a few small things.

src/ros_filter.cpp
```diff
@@ -887,7 +887,10 @@ namespace RobotLocalization
 
     // Check if the filter should start or not
     nhLocal_.param("disabled_at_startup", disabledAtStartup_, false);
-    if (!disabledAtStartup_) enabled_ = true;
+    if (!disabledAtStartup_)
```
ayrton04: Actually, sorry, should have pointed this out before, but consider just doing `enabled_ = !disabledAtStartup_`.

ayrton04 (approved):
LGTM, with one more comment, but you should feel free to ignore it.

ayrton04:
LGTM, thanks for the PR!

Issue #382, opened 2017-10-20
Backport to Indigo


jacobperron:
Due to #353, `indigo-devel` has not been able to take advantage of updates.

By changing all instances of `#include <xmlrpcpp/XmlRpcException.h>` to `#include <XmlRpcException.h>` it seems to have resolved build issues under Indigo.

I'm not sure what best practice is for backporting changes. I have the fix in [my fork](https://github.com/jacobperron/robot_localization/tree/indigo-backport), should I open a PR to `indigo-devel`? Or is there a more elegant solution?

ZakariaChekakta:
@jacobperron  sorry for this question but i don't understand why when we use#include <xmlrpcpp/XmlRpcException.h>  the compilation failed
but when we use #include </XmlRpcException.h> the package is compiled without any problem

jacobperron:
@chekakta No worries. I should have mentioned why.

It looks like the headers were pushed into a directory in [ros_comm](https://github.com/ros/ros_comm/commit/a8c5f9e95e6823417b67ee0da2fa0a91acd97ef5) starting with version 1.12.7. But, current ros_comm version for Indigo is 1.11.21 and so it doesn't have this change.

I posted about it in response to a similar issue on [ROS Answers](https://answers.ros.org/question/258860/about-xmlrpcppxmlrpcexceptionh/?comment=273713#post-id-273713).

jacobperron:
In retrospect, that change in ros_comm exports the old include directory for compatibility. I just confirmed that that both Kinetic and Indigo can build with the proposed change:

`#include <xmlrpcpp/XmlRpcException.h>` -> `#include <XmlRpcException.h>`

ayrton04:
Repeating my comments from the PR:

The lack of backports for new features to Indigo was not solely a result of the xmlrpcpp issue, but became an intentional move to do a better job of "freezing" releases for each ROS distribution. There are a lot of Indigo users out there, and I'm concerned about breaking functionality for them. Even though all of my tests pass, some of the changes between Indigo and Kinetic could conceivably change behavior in some subtle way that I had not anticipated.

I am still taking changes for Kinetic, though, and I put them in Lunar as well. However, some of the major changes (like nodelet support) are for Lunar only.

This obviously doesn't preclude users from building the Kinetic and Lunar branches from source.

jacobperron:
[...]

PR #383, opened 2017-10-21
Update xmlrpcpp includes for Indigo support

jacobperron:
Addresses #382 

This will allow syncing `indigo-devel` and `kinetic-devel` branches. I don't see the harm in using the backwards compatible header include.

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -58,7 +58,7 @@
 #include <diagnostic_updater/publisher.h>
 #include <diagnostic_msgs/DiagnosticStatus.h>
 
-#include <xmlrpcpp/XmlRpcException.h>
+#include <XmlRpcException.h>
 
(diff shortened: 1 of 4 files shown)


ayrton04:
So while this is a good PR (thanks!), the lack of backports for new features to Indigo was not solely a result of this issue, but became an intentional move to do a better job of "freezing" releases for each ROS distribution. There are a lot of Indigo users out there, and I'm concerned about breaking functionality for them. Even though all of my tests pass, some of the changes between Indigo and Kinetic could conceivably change behavior in some subtle way that I had not anticipated.

I am still taking changes for Kinetic, though, and I put them in Lunar as well. However, some of the major changes (like nodelet support) are for Lunar only.

This obviously doesn't preclude users from building off the Kinetic and Lunar branches, which I imagine is the point of this PR. It appears to make the build farm happy, so LGTM. Thanks again!

PR #386, opened 2017-11-15
Zero out rotation in GPS to base_link transform

Roboterbastler:
I had issues when using a GPS sensor frame that is rotated w.r.t. base_link. Because rotation is not relevant to GPS positions I think the transformation from base_link to the sensor frame should be set to identity. Can anyone confirm this?

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -429,7 +429,10 @@ namespace RobotLocalization
 
       if (can_transform)
       {
+        // Zero out rotation because we don't care about the orientation of the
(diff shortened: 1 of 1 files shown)


ayrton04:
I'll have to make sure I'm not missing anything, but yeah, the orientation of the GPS sensor is irrelevant. I've always been simply not specifying a rotational offset for the GPS, but I suppose that could be a problem with URDF models when the GPS is mounted, e.g., sideways.

ayrton04:
Also, thank you for the PR!

PR #398, opened 2018-02-13
adding log statements for nans in the invertable matrix

SteveMacenski:
From discussions in ros answers: [here](https://answers.ros.org/question/272226/robot_localization-ouputs-nans-when-imuodom-messages-not-naning-works-in-robotposeekf/)

Sorry it took so long. It dropped off my radar but I was back on some state estimation work and remembered I had promised to do this. Hopefully this help some other weary traveler.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1825,7 +1825,7 @@ namespace RobotLocalization
 
         // the filteredPosition is the message containing the state and covariances: nav_msgs Odometry
 
-        if (validateFilterOutput(filteredPosition))
+        if (!validateFilterOutput(filteredPosition))
         {
           ROS_ERROR_STREAM("Critical Error, NaNs were detected in the output state of the filter." <<
                 " This was likely due to poorly coniditioned process, noise, or sensor covariances.");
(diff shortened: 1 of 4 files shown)


ayrton04:
I like the idea, but I wonder if we couldn't find some way to validate the filter output in the ROS-aware `ros_filter.cpp`, such that we can then publish a diagnostic message.

SteveMacenski:
I could check what the output looks like from [here](https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ros_filter.cpp#L1815) (i.e. that none of the position or orientation components are NaN / non-normalized quaternion). 

My concern is that we'd be notifying the user about an error that could potentially be caused by any number of issues, I don't necessarily know that the only thing that would cause this is the poorly conditioned covariances manifested via a failed inversion.

What do you think?

ayrton04:
I think the warning/error would have to be suitably generic. `NaN` values in the covariance matrix can result from an invalid `process_noise_covariance`, and invalid measurement covariance, or an invalid `initial_estimate_covariance`. `NaN` values in the state estimate can result from the same issues, as well as invalid measurement values.

Alternatively, we can check for singular matrices before we pass them to the filter. This would have the added benefit of not blowing the state estimate up, though I'm not sure if we want to mask issues with covariance matrices that really ought to be fixed.

In any case, I think it would be most appropriate for the error to be published as a diagnostic, which would mean we'd have to do something in `ros_filter.cpp`.

SteveMacenski:
Good feedback, I agree. I will try to get this done this week between tasks at work.

SteveMacenski:
Hi,

I've gone back through and added the log mesages in `ros_filter` and validate that the full state is not in infinity or NaN. It's a little brute-force but there's no nice way to iterate over a ros message as far as I know. 

Let me know what you think. We can definitely play with wording as you find appropriate.

src/ros_filter.cpp
```diff
@@ -3088,6 +3096,38 @@ namespace RobotLocalization
     return retVal;
   }
 
+  template<typename T>
+  bool RosFilter<T>::validateFilterOutput(const nav_msgs::Odometry &message)
+  {
+    if (std::isnan(message.pose.pose.position.x) || std::isinf(message.pose.pose.position.x))
```
ayrton04: Three things:

1. It may be a bit easier to read like this:

```
return !std::isnan(message.pose.pose.position.x) && !std::isinf(message.pose.pose.position.x) &&
  !std::isnan(message.pose.pose.position.y) && !std::isinf(message.pose.pose.position.y) &&
...
```

2. I really need to enable C++11 for Kinetic and on, because [std::isfinite](http://en.cppreference.com/w/cpp/numeric/math/isfinite) does this a bit more cleanly.

3. It'll likely only warn us a cycle or two earlier, but I wonder if we should do

```
bool is_valid = !std::isnan(message.pose.pose.position.x) && !std::isinf(message.pose.pose.position.x) &&
...
```

as above, but then check the covariances as well:

```
for (size_t i = 0; i < message.pose.covariance.size(); ++i)
{
  const double &covar = message.pose.covariance[i];
  is_valid = is_valid && !std::isnan(covar) && !std::isinf(covar);
}
```

Then we'd repeat for twist. What do you think?

ayrton04 (commented):
Thanks again for taking the time to work on this! I just had a couple more suggestions, as long as we're adding this method.

src/ros_filter.cpp
SteveMacenski: On point 1, done. 
On point 2, would you like that to be part of this PR?
On point 3, for warning a couple cycles beforehand, I think while the state is still in the set of real numbers, we should let things play out. It's a couple more checks and if its going to NaN out, the behavior I would find most helpful if I could see the process that lead there in my logs.

ayrton04: No, (2) wasn't meant to be actionable. There should probably be a C++11 or C++14 PR at some point, but that's for me to worry about.

So for (3), are you advocating for checking the covariance or not? I'm really not that concerned, since we're probably look at (at most) one cycle between a `NaN/Inf` being in any part of the state or covariance and it ruining the entire thing. I'm fine to leave it as-is if you are.

ayrton04:
If you're happy with this, let me know, and I'll merge. Thanks again for the PR! I appreciate it.

src/ros_filter.cpp
SteveMacenski: I'm advocating for not checking- as you say its not much of a warning for some additional checks and I think it might be helpful for those logging outputs to be able to see where the NaN hits. It could have some insight.

SteveMacenski:
Let me test this for a couple days on my robots at work. I'll comment back when comfortable. 

edit: I'm a very nervous person when it comes to filters, I've been burned before by small innocuous seeming changes bringing down production systems.

SteveMacenski:
I ran for a couple days on a couple of robots and happy with it. Merge as you're able.

ayrton04:
Thanks again!

PR #405, opened 2018-04-07
Add published acceleration topic to documentation

okalachev:
From the documentation it wasn't clear, where acceleration data should be.


ayrton04:
Many thanks!

Issue #416, opened 2018-09-13
Dynamic Covariance Scaling 


SteveMacenski:
When the robot is not moving the dynamic scaling seems to zero out the covariance (the diagonal of `P` in the similarity transform and its transpose is `0`). 

When this is 0, that represents both a symmetric matrix as well as represent extremely high confidence in our measurements? I believe the intended application for this when the norm of the velocity vector is zero or near zero is to apply a high covariance to the process rather than a lower one. 

On thinking about it, for the general case, I would think the application would want to do the opposite scaling - rather than making the process covariance low when moving slow, we'd want it to be high since there are less ticks per rotation (assuming an encoder is involved) and therefore the expected inputs degrade, which is generally true of most non-inertial sensors. 

https://github.com/cra-ros-pkg/robot_localization/blob/f0b540bbd0d645c31872aabbf1ced3c2491c9dbb/src/filter_base.cpp#L132

Either way, making a case here when the velocity norm is near zero to make the covariance high to stop motion would be good. I'm willing to work on any and all of these.

Thoughts?

SteveMacenski:
For context, this came up when fusing in a couple of sensors with tricky and changing biases I'm still not 100% catching, so when standing still for long periods I see a little drift that I hadn't seen beforehand.

ayrton04:
So just backing up, the prediction step predicts the next covariance as `P_pred = JPJ' + Q`. If the robot isn’t moving at all, we zero out `Q` via the method you linked. This prevents the state estimate covariance `P` from growing when we’re stationary. If the robot is moving and then stops, it doesn’t make sense for its uncertainty to grow once it’s at rest.

As far as scaling with velocity, the process noise covariance can be used to model how well or poorly the kinematic model I’m using matches your robot’s kinematics. For example, if you zero out all the 3D variables and `Y` variables, you end up with the unicycle model. If you were using this to track the state of, say, an Ackermann vehicle, then the process noise could represent the mismatch in the vehicle’s true kinematics with the unicycle model.

In that case, if the vehicle made a very small motion (low velocity) over some time delta `dt`, the amount of process noise would be small, as our state prediction wouldn’t differ from reality that strongly. If it moved a larger distance (high velocity) over that same `dt`, then our predicted state should have more error, because it differs more from the robot's true state.

> since there are less ticks per rotation (assuming an encoder is involved)

[...]

PR #419, opened 2018-09-24
Processing Filter Toggle Mode

SteveMacenski:
In response to discussion in https://github.com/cra-ros-pkg/robot_localization/issues/416

Gives a service interface to pause accepting and processing new measurements but continuing to provide pose estimate information at the requested rate.

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -628,7 +628,7 @@ template<class T> class RosFilter
     bool enabled_;
 
     //! $brief Whether the filter should process new measurements or not.
-    bool toggled_on_;
+    bool toggledOn_;
 
     //! @brief Message that contains our latest transform (i.e., state)
     //!
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -298,6 +298,10 @@ template<class T> class RosFilter
     //!
     void clearExpiredHistory(const double cutoffTime);
 
+    //! @brief Clears measurement queue
+    //!
+    void clearMeasurementQueue();
+
     //! @brief Adds a diagnostic message to the accumulating map and updates the error level
     //! @param[in] errLevel - The error level of the diagnostic
     //! @param[in] topicAndClass - The sensor topic (if relevant) and class of diagnostic
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -38,6 +38,7 @@
 #include "robot_localization/filter_base.h"
 
 #include <robot_localization/SetPose.h>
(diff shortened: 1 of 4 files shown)


SteveMacenski:
Toggling CI

include/robot_localization/ros_filter.h
```diff
@@ -610,6 +627,9 @@ template<class T> class RosFilter
     //! @brief Whether the filter is enabled or not. See disabledAtStartup_.
     bool enabled_;
 
+    //! $brief Whether the filter should process new measurements or not.
+    bool toggled_on_;
```
ayrton04: As much as I hate it, this should be `toggledOn_`.

ayrton04 (commented):
This looks good. I do have one concern, though, which is that a large "toggle off" time will result in a huge time delta for the first prediction step that happens when the node is toggled back on. The process noise is [multiplied by the time delta](https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/ekf.cpp#L389) and then added to the state covariance during the prediction step, so your prediction step will have a massive covariance increase, which will only be corrected for any measured variables in your first "toggle on" cycle. You might want to verify this by looking at the first EKF output message post-toggle-on.

I _think_ one easy way to fix this is to call `setLastMeasurementTime` with the current time after clearing the measurement queue in your "toggle off" case.

EDIT: my comments obviously only apply if you aren't using dynamic PNC.

include/robot_localization/ros_filter.h
SteveMacenski: done

SteveMacenski:
This is a good point, I'll have to think about this for a little while. That's probably the right answer but something about it feels 'wrong' to me.

Edit: after reading the filter base, I agree, this is the right move. commit to follow.

ayrton04 (approved):

ayrton04:
Thanks for the PR!

Timple:
Would it make sense here to set the initial state of the toggle with a parameter?

ayrton04:
As in, a ROS parameter? There's already a parameter for that behavior. It's `disabled_at_startup`.

https://github.com/cra-ros-pkg/robot_localization/blob/c8b8e3f2ba478a6301cf0a9e6f2c9c3c6896f372/src/ros_filter.cpp#L915

The filter starts in a disabled state, and requires a service call to make it start processing measurements and producing estimates.

It would make sense to unify them, really, but this at least provides the functionality you may be after.

Timple:
That is exactly what I was looking for. I seemed to have missed that while looking at the diff.

Thanks!

ayrton04:
It's been around for a while, but it occurs to me that I may not have put it in the documentation. Sorry about that!

PR #425, opened 2018-11-18
major refactor unscented kalman filter

treyamador:


treyamador:
major refactor unscented kalman filter

PR #435, opened 2018-12-18
Changes for working with Ubuntu 18.10 (Cosmic)

stridera:
Add const& to catch values to prevent the error:  catching polymorphic type ‘class tf2::TransformException’ by value

And ZoneNumber by 0x3fU to prevent error: directive output may be truncated writing between 1 and 11 bytes into a region of size 4


ayrton04 (approved):

ayrton04:
Thank you!

PR #437, opened 2019-01-04
Rename odomBaseLinkTrans to baseLinkOdomTrans

thallerod:
Adhere to the naming convention \<fromFrame\>\<toFrame\>Trans used for worldBaseLinkTrans and mapOdomTrans.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1885,9 +1885,9 @@ namespace RobotLocalization
             tf2::Transform worldBaseLinkTrans;
             tf2::fromMsg(worldBaseLinkTransMsg_.transform, worldBaseLinkTrans);
 
-            tf2::Transform odomBaseLinkTrans;
+            tf2::Transform baseLinkOdomTrans;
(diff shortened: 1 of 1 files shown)


ayrton04:
Yes, agreed.

ayrton04:
Just as an aside, the argument order for `lookupTransform` is `target_frame`, `source_frame`, so the naming convention I used makes sense from that perspective. However, the resulting transform (IIRC) actually takes a position in the `target_frame` and gives its pose in the `source_frame`. I frequently get these backwards, though.

thallerod:
`lookupTransform` will return the transform to get data from the `source_frame` into the `target_frame`. This is the transform from the coordinate frame of the `target_frame` to the coordinate frame of the `source_frame` (as discussed e.g. [here](https://github.com/ros/geometry/issues/108) and [here](https://answers.ros.org/question/229463/confusing-tf-transforms/)).

ayrton04:
Yep, makes sense.

PR #439, opened 2019-01-15
Meridian convergence adjustment

pavloblindnology:
A major false assumption has been made that the UTM grid aligns with East and North, while it's not true.
For UTM projection (based on [Transverse Mercator projection](https://en.wikipedia.org/wiki/Transverse_Mercator_projection)) convergence (the angle between projected meridians and grid Y-axis) is zero on the equator and non-zero everywhere else. It increases as the poles are approached or as we're getting farther from central meridian.
In order to make our world frame aligned with ENU we need to account for this.
Otherwise, other ROS packages which rely on this property will misunderstand the orientation.
E.g. [rviz_satellite](https://github.com/gareth-cross/rviz_satellite) which uses [Normal Mercator projection](https://en.wikipedia.org/wiki/Mercator_projection) gives tiles rotated about world frame origin without convergence adjusted.
This PR adds the needed correction. Meridian convergence is computed as in [Spherical Transverse Mercator projection](https://en.wikipedia.org/wiki/Transverse_Mercator_projection#Convergence) which gives results very close to those used in UTM, based on [Ellipsoidal Transverse Mercator projection](https://en.wikipedia.org/wiki/Transverse_Mercator:_Redfearn_series#Point_scale_and_convergence).

### DIFF
--- a/include/robot_localization/navsat_conversions.h
+++ b/include/robot_localization/navsat_conversions.h
@@ -179,12 +179,16 @@ static inline char UTMLetterDesignator(double Lat)
  * East Longitudes are positive, West longitudes are negative.
  * North latitudes are positive, South latitudes are negative
  * Lat and Long are in fractional degrees
+ * Meridian convergence is computed as for Spherical Transverse Mercator,
+ * which gives quite good approximation.
+ * 
+ * @param[out] gamma meridian convergence at point (degrees).
  *
  * Written by Chuck Gantz- chuck.gantz@globalstar.com
  */
 static inline void LLtoUTM(const double Lat, const double Long,
                            double &UTMNorthing, double &UTMEasting,
-                           std::string &UTMZone)
+                           std::string &UTMZone, double &gamma)
 {
   double a = WGS84_A;
   double eccSquared = UTM_E2;
@@ -249,6 +253,8 @@ static inline void LLtoUTM(const double Lat, const double Long,
                *(A*A/2+(5-T+9*C+4*C*C)*A*A*A*A/24
                  + (61-58*T+T*T+600*C-330*eccPrimeSquared)*A*A*A*A*A*A/720)));
 
+  gamma = atan(tan(LongRad-LongOriginRad)*sin(LatRad)) * DEGREES_PER_RADIAN;
+
   if (Lat < 0)
           {
             // 10000000 meter offset for southern hemisphere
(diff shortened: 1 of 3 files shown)


ayrton04:
Thanks for the PR! I don't suppose you have a sample bag file or perhaps a visualization for this? I obviously see that this is a major fix, but for posterity, if we had a sample bag file or something to go with the PR, that would help.

Thanks again for taking the time to make this change.

pavloblindnology:
Below snapshot from Rviz is shown with:
robot_localization - uses datum LatLon [50.4048, 30.5255] and world_frame "map",
Rviz - Fixed Frame is "map", Grid is for "map" frame, Axes show "map" frame origin,
rviz_satellite - shows 3x3 Google tiles around the datum at zoom level 20,
Green rectangle in Rviz is formed by specifying rectangular LatLon area corresponding to the 3x3 Google tiles with left bottom [50.4045789161, 30.5248260498] and right top [50.4052353746, 30.5258560181], converted to "utm" frame (zone 36) with `NavsatConversions::LLtoUTM`.
`rosrun tf tf_echo utm map` returns

> At time 0.000
> - Translation: [324163.515, 5586566.487, 0.000]
> - Rotation: in Quaternion [0.000, 0.000, 0.000, 1.000]
>             in RPY (radian) [0.000, -0.000, 0.000]
>             in RPY (degree) [0.000, -0.000, 0.000]

That is, zero magnetic declination and yaw offset are specified.

As you can see, Google tiles are aligned with the grid - because rviz_satellite assumes that "map" frame is ENU aligned. While green rectangle is rotated by some angle around "map" frame origin - because robot_localization uses LatLon -> utm conversion, hence points with equal longitude\latitude are converted to points with different utm_easting\northing, hence in "map" frame they also have different X\Y.
To solve this, we need "map" to be really ENU aligned - that is to be rotated by meridian convergence angle relative to "utm" frame.
robot_localization_no_meridian_convergence

ayrton04:
Excellent. OK, thanks again. I realise now that one of the test bags that I have contains GPS data, so I can try this out with it, and then merge.

ayrton04:
Looks good. Thanks for this!

pavloblindnology:
What about melodic-devel? Is this going to be merged into?

ayrton04:
I will make sure the commits get cherry-picked into all active branches. I need to cut a release for this package soon anyway.

pavloblindnology:
Thanks.

PR #440, opened 2019-01-16
CPR changes on top of upstream melodic-devel

efernandez:
I've cherry picked the local changes from our fork and put them on top of upstream `melodic-devel`.

I looked in our history for `cpr-` tags to find the commits we had. You'll have to confirm I didn't miss anything.

For the disable service and inverse TF commits, I have to redo then in order to solve conflicts with upstream.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(robot_localization)
 
 find_package(catkin REQUIRED COMPONENTS
+  angles
   diagnostic_msgs
   diagnostic_updater
   eigen_conversions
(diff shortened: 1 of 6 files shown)


efernandez:
Sorry, not intended for upstream. At least not yet. But now that it got it here, maybe I can have some comments about what could be accepted upstream @ayrton04 :smiley:

ayrton04:
I'd say that's fairly broad, especially if it's stuff that's been in use for what I would presume would be a long time and on a lot of robots. :)

The only thing I have to consider now is that there are ROS 2 branches.

Just a heads-up that I anticipate [this](https://github.com/cra-ros-pkg/robot_localization/pull/439) getting merged soon. Definitely seems like it would be important for outdoor robots.

PR #443, opened 2019-01-30
Enable build optimisations if no build type configured.

gavanderhoorn:
As per subject.

Triggered by [Low update rate on robot_localization](https://answers.ros.org/question/310785) on ROS Answers.

See the comment on cc93cef0 for some rationale.

This setup is also used by MoveIt and has worked well there.


ayrton04 (approved):
LGTM. Thanks!

Issue #445, opened 2019-02-11
broadcast_utm_transform_as_parent_frame not found in documentation


diasdm:
Hello,
As far as I know, the parameter broadcast_utm_transform_as_parent_frame is used in [r_l](https://github.com/cra-ros-pkg/robot_localization/blob/f6bf556407945e7b192c78959c51d2f10f62e488/src/navsat_transform.cpp#L84), but it is not mentioned in the [documentation](http://docs.ros.org/melodic/api/robot_localization/html/state_estimation_nodes.html). Shouldn't it be?
It is already described [here](https://github.com/cra-ros-pkg/robot_localization/blob/f6bf556407945e7b192c78959c51d2f10f62e488/params/navsat_transform_template.yaml#L28), so a similar description can be used.
Thank you.

ayrton04:
Agreed. Care to submit a PR?

diasdm:
Done, #446

PR #446, opened 2019-02-11
Add broadcast_utm_transform_as_parent_frame description to doc

diasdm:
Regarding issue #445 
* Added the parameter description
* Fixed what I believe was a typo in the Published Transforms description

### DIFF
--- a/doc/navsat_transform_node.rst
+++ b/doc/navsat_transform_node.rst
@@ -50,7 +50,7 @@ If *true*, ``navsat_transform_node`` will wait to get a datum from either:
 ~broadcast_utm_transform_as_parent_frame
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 If *true*, ``navsat_transform_node`` will publish the utm->world_frame transform instead of the world_frame->utm transform. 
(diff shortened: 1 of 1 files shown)


Timple:
So if I use `broadcast_utm_transform_as_parent_frame` do I also still need `broadcast_utm_transform`?
Haven't looked at the code, just following documentation :)

diasdm:
[I believe so](https://github.com/cra-ros-pkg/robot_localization/blob/f6bf556407945e7b192c78959c51d2f10f62e488/src/navsat_transform.cpp#L288), if you want the transform to be published at all, you have to set ``broadcast_utm_transform`` to true.

Timple:
My point was that this was not clear from the documentation :). So this might be added.
(Thanks for looking it up though)

diasdm:
What is your suggestion? To change what is stated [here](https://github.com/cra-ros-pkg/robot_localization/pull/446/files#diff-67c66a3bfd4085a496b7ce4d7abbc936R53), to something else?
To `Note that for the transform to be published ``broadcast_utm_transform`` also has to be set to *true*.`, for example?

ayrton04:
I think both should be required, and I think the additional note is a good idea.

Thanks very much for doing this!

ayrton04 (approved):

Issue #441, opened 2019-01-17
Document undocumented parameters


maracuya-robotics:
Some parameters like `predict_to_current_time` (and maybe more?) are not documented.

ayrton04:
Thanks for the report. I'd welcome a PR!

maracuya-robotics:
Wow, that was fast!

Found another one, `disabled_at_startup`. IMHO documentation should be done by someone who knows what these parameters do. I can just guess by the name and by looking at the code.

diasdm:
I think these are all the undocumented parameters for the documented nodes.

In `navsat_transform.rst` is missing:
* transform_timeout (only documented for `state_estimation_nodes.rst`)

In `state_estimation_nodes.rst` is missing:
* base_link_frame_output
* predict_to_current_time
* disabled_at_startup

seebq:
Hey @maracuya-robotics and @diasdm I submitted a PR #454 for y'all.  Can you review and let @ayrton04 know if this documents the parameters that were missing?

ayrton04:
Assuming this is addressed by #454. Thanks @seebq.

PR #454, opened 2019-03-23
Add missing undocumented params

seebq:
Referencing #441.

I think, after looking at it, we should make `base_link_output_frame` an advanced param, and not in the main block.

Also, I removed the word "rejected" at end, might just been an error.

### DIFF
--- a/doc/navsat_transform_node.rst
+++ b/doc/navsat_transform_node.rst
@@ -52,6 +52,10 @@ If *true*, ``navsat_transform_node`` will wait to get a datum from either:
 If *true*, ``navsat_transform_node`` will publish the utm->world_frame transform instead of the world_frame->utm transform. 
 Note that for the transform to be published ``broadcast_utm_transform`` also has to be set to *true*.
 
+~transform_timeout
+^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


maracuya-robotics (approved):

doc/navsat_transform_node.rst
```diff
@@ -52,6 +52,10 @@ If *true*, ``navsat_transform_node`` will wait to get a datum from either:
 If *true*, ``navsat_transform_node`` will publish the utm->world_frame transform instead of the world_frame->utm transform. 
 Note that for the transform to be published ``broadcast_utm_transform`` also has to be set to *true*.
 
+~transform_timeout
+^^^^^^^^^^^^^^^^^^
+This parameter specifies how long we would like to wait if a transformation is not available yet. Defaults to 0 if not set. The value 0 means we just get us the latest available (see ``tf2`` implementation) transform.
```
diasdm, line 57: Hi, it does look good to me. I would only change this to "~~This parameter~~ Specifies..." so it is consistent with the rest.
Thank you!

ayrton04:
LGTM. Thanks for this!

PR #462, opened 2019-05-17
Better handling of covariance for raw GPS input

kunaltyagi:
Might require using a flag to switch between legacy mode and covariance in (deg, deg, m) mode.

## Rationale
Several drivers and ROS wrappers for GPS sensors output covariance in same units as (lat, long, altitude), however the processing is done using that covariance in (m, m, m) units which leads to overconfidence in GPS

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -475,12 +475,22 @@ namespace RobotLocalization
       latest_utm_pose_.setOrigin(tf2::Vector3(utmX, utmY, msg->altitude));
       latest_utm_covariance_.setZero();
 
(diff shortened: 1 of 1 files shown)


ayrton04:
The `sensor_msgs/NavSatFix` [specifies units](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/NavSatFix.html) for the covariance:

```
# Position covariance [m^2] defined relative to a tangential plane
# through the reported position. The components are East, North, and
# Up (ENU), in row-major order.
#
# Beware: this coordinate system exhibits singularities at the poles.
```

If there are ROS GPS drivers that are reporting covariance in other units, I think those drivers should be corrected, rather than making the package support incorrect units.

Issue #464, opened 2019-05-23
tf_prefix not added to base_link


faizansid77:
The node is running perfectly when I am not using any tf_prefix. Whenever I use tf_prefix, I get no tf_prefix in base_link. This issue is not observed in version 2.4.5. Please revert the code to whatever changes you made after 2.4.5

My /robot1/odom/filtered is publishing this  header:
`header: 
  seq: 181
  stamp: 
    secs: 627
    nsecs: 240000000
  frame_id: "robot1_tf/odom"
child_frame_id: "base_link"
`

My robot's namespace is "robot1" and tf_prefix is "robot1_tf" 

Edit: I am working on ros-kinetic

ayrton04:
Yep, that's a bug. Try this branch:

https://github.com/cra-ros-pkg/robot_localization/tree/fix-output-tf-prefix

PR #465, opened 2019-05-23
Fix issue with tf_prefix and the base_link frame_id

ayrton04:
Addresses #464.


faizansid77:
Why is this still not merged?
Please perform the same edit to kinetic-devel branch as well
This issue is major!

ayrton04:
Because I have a day job that rarely involves touching this package.

If you're desperate for the fix, please clone this repo and checkout the branch in question.

The fix will be backported once merged.

PR #467, opened 2019-06-18
Service for converting from map frame to longitude / latitude

oswinso:
Currently, after a transform has been found in `navsat_transform`, there is no way to convert points in the `map` frame to GPS coordinates.

This PR creates a service that allows for conversion of arbitrary points to and from the `map` frame to GPS coordinates using the transform and UTM zone stored.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,8 @@ add_service_files(
     SetDatum.srv
     SetPose.srv
     ToggleFilterProcessing.srv
+    FromLL.srv
+    ToLL.srv
 )
 
 generate_messages(
--- a/include/robot_localization/navsat_transform.h
+++ b/include/robot_localization/navsat_transform.h
@@ -136,6 +136,16 @@ class NavSatTransform
     //!
     void setTransformOdometry(const nav_msgs::OdometryConstPtr& msg);
 
+    //! @brief Transforms the passed in pose from utm to map frame
+    //! @param[in] utm_pose the pose in utm frame to use to transform
+    //!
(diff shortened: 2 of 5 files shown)


ayrton04:
Thanks for this. Will review soon.

seebq:
Hey @ayrton04 anything we can do to help review or cleanup that you'd like to see?  i.e. docs?  

Please delegate/assign if we can help.

oswinso:
Hey @ayrton04 the build seems to be failing the test result, even though none of my changes are related to the actual EKF filter. Any ideas what is happening?

ayrton04 (commented):
Thanks for the PR.

I think my main concern here is that we're repeating some code. For example, these blocks are obviously quite similar:

https://github.com/cra-ros-pkg/robot_localization/blob/46396c01157b79ec06874238bb381693374b9006/src/navsat_transform.cpp#L658

https://github.com/cra-ros-pkg/robot_localization/pull/467/files#diff-dc270f64cc0f5f1b532a944e353b01e0R368

I wonder if we could make some methods that do these conversions, and then call them in the existing code and in this PR. Otherwise, the service seems fine.

ayrton04:
Re: build farm breakages, I need to walk back through the history to see where that started breaking.

oswinso:
Ok, cool. I've done my best to make some changes to refactor out the common blocks that are used. Let me know if you think there are more improvements that I should do.

ayrton04 (approved):

ayrton04:
Kicking the build farm.

ayrton04 (approved):

ayrton04:
Thanks!

PR #472, opened 2019-07-09
fix: wall time used when `use_sim_time` is true

UnexpectedLobster:
Hi,

When using the param `use_sim_time`, the initialization of `ros::Time` would cause `ros::Time::now()` to return the wall time instead of the clock time, if it is not yet published. The expected behaviour would be to wait for the `/clock` topic, and returning `0` in the meantime. 
In addition, the time initialization is already done by the node init and should not be called by hand, except in a library. This does not seems to be the case here, as `ros_filter.cpp` is always part of a node.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -107,8 +107,6 @@ namespace RobotLocalization
   template<typename T>
   void RosFilter<T>::initialize()
   {
-    ros::Time::init();
-
     loadParams();
 
     if (printDiagnostics_)


ayrton04:
Yes, agreed, though this seems like it might cause some fairly major changes in behaviour. I wonder if anyone is relying on the current setup, even if it is incorrect.

Also, I wonder if it makes sense to add a block like this, from `navsat_transform_node`:

https://github.com/cra-ros-pkg/robot_localization/blob/2b57df66ded98b3ae4d08176f54beead4e9e1f5f/src/navsat_transform.cpp#L69

UnexpectedLobster:
From the implementation of isValid : http://docs.ros.org/melodic/api/rostime/html/src_2time_8cpp_source.html#l00324 this seems to wait for the ros::Time to be different from 0 in the case of `use_sim_time` is used. I would say this is rather important in order for the filter not to make any predict / update while the delta-time will obviously be 0.

SteveMacenski:
I don't think the ROS time init should cause issues. The ROS init is called before the filter is created [here](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ekf_localization_node.cpp#L41) so that time init should be called. I think that call is redundant and maybe from a time pre-ROS to make use of the timing library?

UnexpectedLobster:
@SteveMacenski here is how to reproduce the bug (sorry, I should have posted this at the beginning) :
Add the following `ROS_ERROR_STREAM_THROTTLE(0.1, "" << curTime);` [here](https://github.com/cra-ros-pkg/robot_localization/blob/086bcfcc81d422d357e2ecabbadcca68624a0062/src/ros_filter.cpp#L1831).

Now running a kalman node with `use_sim_time` set to `true`, you will see the `periodicUpdate` loop is running with wall time. Removing the `Time::init` fixes the issue.

ayrton04:
I can't recall why that's there, TBH, but I suppose that, as long as it doesn't break the behaviour of live robots, I'd be OK with applying the fix as-is. Still, I'd like to add a call to `waitForValid()`. Should we also gate it behind a new parameter? I can't see why you'd want the filter to run with 0 time stamps, so perhaps no state estimate is preferred.

I guess my main concern is a use case whereby some other node is affected by the lack of filter output for the time between when the software starts and when `/clock` starts publishing.

UnexpectedLobster:
> I can't see why you'd want the filter to run with 0 time stamps, so perhaps no state estimate is preferred.

Actually that's the behavior we get when removing the `Time::init`, the filter wait for `/clock` messages in order to output something. That's fine by me. 

> I guess my main concern is a use case whereby some other node is affected by the lack of filter output for the time between when the software starts and when /clock starts publishing.

In my opinion, this kind of race condition should be taken care of in the other node consuming the filter output. Publishing "as fast as possible" cannot guarantee other node to not be affected by this, as roslaunch and node lifecycles are somewhat non-deterministic in ROS1.

ayrton04:
> In my opinion, this kind of race condition should be taken care of in the other node consuming the filter output. Publishing "as fast as possible" cannot guarantee other node to not be affected by this, as roslaunch and node lifecycles are somewhat non-deterministic in ROS1.

Agreed, but it's the change in behaviour that concerns me. This is all completely hypothetical, but what worries me is we make this change, release, etc., and then we break someone's stack because of it. Should they fix the other nodes? Sure. Will tracking this down create a headache for them? Yes.

Having said that, this is for Melodic, and while that's been out for a while, I think I'd be willing to risk it, especially since, again, this should only affect sim.

In any case, would you mind also (a) adding a call to `waitForTime`, and (b) putting that call behind a new parameter?

UnexpectedLobster:
My bad, I didn't consider backward compatibility and breaking changes, you're absolutely right.
However, I don't understand the use of this `ros::Time::waitForValid`. What is this supposed to do ? In the case of `use_sim_time`, the node already waits for `/clock` messages before running. 

Wouldn't it be better to gate the call of `ros::Time::init` behind a (default true) parameter to ensure the previous behavior by default ?

UnexpectedLobster:
I hope this is what you meant. It does not induce any bugs, but I'm not sure what it brings either.

SteveMacenski:
Sorry re-reading my comment I can see how it could be taken the wrong way. The point I failed to articulate is that since time initialization happens on ROS initialization prior, the correct clock (sim, wall) is used and it should be totally harmless to remove that line as proposed on hardware robots and for simulated robots it should create consistent behavior.

ayrton04:
Ah, OK, now that I've had time to actually look at this, yes, I agree that the original approach was fine. Feel free to put the PR back to your original change and I'll merge. Sorry about that. Thanks!

zouhairm:
Glad to finally find this issue - I've been having this problem on melodic and it's been driving me crazy as I've been certain I've followed the right steps (setting use_sim_time, publishing clock, etc.) yet the debug file is showing wall time likewise with the frequency of the filter! 

When is this due to be merged? Are there any instructions somewhere on how to patch a ros package in the meantime?

zouhairm:
if someone else stumbles on it - I managed to use "overlays" by cloning @PymZoR 's branch into a catkin_ws and building it; although I'm still seeing in the debug file this:
`Integration time is 1564630170.833963871`

despite having set `rosparam set /use_sim_time true` 

EDIT:
I thought that maybe I needed the node to wait for valid time, which seems to have been the discussion above. BTW I do think that should be the default behavior as that's what the /use_sim_time documentation implies. I understand that others might be relying on the current non-confirming behavior, and exposing a fix as done in your ` 9fdbcc8` commit would be the right thing to do (I wish you hadn't force-pushed, makes it harder to properly get the commit from git history :p)

Despite that I'm still seeing integration time not starting at 0 even though that's the value I'm sending in /clock ?

ayrton04:
> Despite that I'm still seeing integration time not starting at 0 even though that's the value I'm sending in /clock

You're manually publishing to `/clock`? Are you perhaps also playing back a bag file with `--clock`? What are the time stamps on your sensor measurements?

ayrton04:
Also, I'll merge and release this next week. I'm on vacation at the moment. Thanks, @PymZoR!

UnexpectedLobster:
> if someone else stumbles on it - I managed to use "overlays" by cloning @PymZoR 's branch into a catkin_ws and building it; although I'm still seeing in the debug file this:
> `Integration time is 1564630170.833963871`
> 
> despite having set `rosparam set /use_sim_time true`
> 
> EDIT:
> I thought that maybe I needed the node to wait for valid time, which seems to have been the discussion above. BTW I do think that should be the default behavior as that's what the /use_sim_time documentation implies. I understand that others might be relying on the current non-confirming behavior, and exposing a fix as done in your ` 9fdbcc8` commit would be the right thing to do (I wish you hadn't force-pushed, makes it harder to properly get the commit from git history :p)
> 
> Despite that I'm still seeing integration time not starting at 0 even though that's the value I'm sending in /clock ?

@zouhairm : From my test setup, the commit `9fdbcc8` happened to do nothing. Using `sim_time` with my first commit did the magic : the node "blocked" until it received a valid time on `/clock`, and that's the behaviour I was searching for. Isn't it the case for you ? I'm not to understand the issue you had despite having the fix.

zouhairm:
@ayrton04 : yes I'm manually publishing to /clock from a python script that's streaming the data. I can verify that the timestamps of the messages printed in the debug log are what I expect (modulo a couple of oddities printing timestamps with decimal points less than .10, and the messages sometimes being out of order) . ~~The timestamp reported by the robot_localization ekf node however begins at the system clock, and eventually jumps. I am using rosbag but only to record data, and I verified using `rostopic echo /clock` that the messages do begin at 0 and that no other node is sending out a /clock message.Also the filter running rate seems to still be driven by the wall clock even though with the removed Time::Init() - I would expect that it should be sleeping until it receives the right timestamp.~~

My bad - in the process of debugging things i had moved the use_sim_time to be done by a `rosparam set` but apparently that doesn't work. I hadn't seen a difference before due to the bug caused by Time::Init, but I've now moved it back to be in a <param> in my launch file and things are  finally starting to work more like the way I expect them to!

Issue #473, opened 2019-07-10
Port to ROS2 Dashing


rotu:
There should probably be a ROS2 Dashing branch in this repo.
Feel free to pull from our fork https://github.com/RoverRobotics-forks/robot_localization

ayrton04:
Is your branch based on the Bouncy/Crystal branches on this repo? I really need to clean up the ROS2 branches here. I want to remove support for diagnostics until it's been officially released for ROS2.

rotu:
My branch is based on master. The diagnostic publisher seems to be working pretty solid.

ayrton04:
Sorry, so you branched the ROS1 version (we have no `master` branch) and also ported it to ROS2? I'm only asking because I made a ROS2 branch for Bouncy, and then we had a user contribution to get the tests running and re-establish the diagnostics.

rotu:
Oops. Yes, we branched from ROS1. Please open a dashing-devel branch. I'll rebase my changes on it and submit a PR.

ayrton04:
A few things.

First, thanks for taking the time to port to ROS2 Dashing. Is there anything about Dashing that would make the code significantly different than Crystal? I only ask because I (and others) have already ported this package to ROS2, and we have `bouncy-devel` and `crystal-devel` branches. When I originally did the ROS2 port, I also made changes to some code style and a few other things, and I'd like to retain them. I'd rather not have our Bouncy and Crystal branches come from our changes, and the Dashing branch come from a completely different source, if that makes sense.

I do need to make some major changes to the Crystal release, though. It's using diagnostics, which, as of Crystal, was not released as a binary package. So I'd welcome PRs to get things cleaned up.

Finally, I should note that I am not actively using ROS2, so I have limited cycles to devote to working on it. I'd really like to get the tests moved over to a proper ROS2 framework, rather than the complex setup they have now.

rotu:
Yes, the biggest change for dashing is that you need to declare all your parameters or things will crash. I understand the want to keep the version branches as similar as possible. I really just need a base branch to submit a PR on.

rotu:
@ayrton04 I rebased my changes in https://github.com/RoverRobotics-forks/robot_localization

ayrton04:
https://github.com/cra-ros-pkg/robot_localization/tree/dashing-devel

mabelzhang:
[...]

Issue #222, opened 2015-07-27
Handle crossing of UTM zone boundaries


ayrton04:
If the robot crosses into a new UTM zone, we need to recompute the transform. We can add some logic [here](https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/navsat_transform.cpp#L145) to check for a change and handle the situation accordingly.

ayrton04:
Need to update [this](http://answers.ros.org/question/214630/navsat-robot_localization/) ROS answers question when complete.

frontw:
Hi!
Is there any update on this?
It's really a problem to use this package for coordinates at zone boundaries. 
Could you, please, give a hint, how to possibly fix it?
Thanks!

ayrton04:
I haven't had the cycles to address it, apologies. I would start here:

https://github.com/cra-ros-pkg/robot_localization/blob/kinetic-devel/src/navsat_transform.cpp#L479

Store the UTM zone, and if it changes, you'd have to re-compute the transform. You'd also have to re-subscribe to the IMU data.

AGummyBear:
I have a branch that avoids the UTM zone boundary issue by utilizing an internal Local Cartesian coordinate system centered on the "origin" (first good gps) rather than UTM coordinates. I configured this as a parameter, so you can switch between using UTM and the Local Cartesian. The branch is in a fork here: https://github.com/Greenzie/robot_localization/tree/agummybear/local_cartesian_option

I've tested this on some bag files, and will hopefully get to test in the real world soon.

`use_local_cartesian = false`
![localization_utm](https://user-images.githubusercontent.com/17852683/82948838-9305b600-9f70-11ea-8585-eb835cb61037.png)

`use_local_cartesian = true`
![localization_local_cartesian](https://user-images.githubusercontent.com/17852683/82948861-9dc04b00-9f70-11ea-8b67-dd55a106e6d4.png)

Timple:
I would be against jumping the `utm` transform. @ayrton04 @frontw what is your use-case for this? The utm zones allow extending outside their boundary. All is fine as long as you don't cross them by a large amount (relative to UTM grid size).

Jumping can cause unexpected behavior. Imagine a path defined in `utm` at 10cm parallel to an `utm` border. Now if the robot crosses the zone for just one sample, the `utm` jumps 500km to the right and the robot will (thus) try to go 500km to the right while it was actually doing pretty okay.
[...]

PR #476, opened 2019-08-05
ROS2 Navsat_transform_node in Dashing

klintan:
I saw there was already an issue and a fork (https://github.com/cra-ros-pkg/robot_localization/issues/473) for Dashing, but made some changes remove all the deprecation warnings.

There might be some `declare_parameters` which could go back to `get_parameter`. 

Also I guess it makes sense to create a new branch dashing-devel and if these changes makes sense, I'll change the PR to that branch instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,9 +5,13 @@ if(NOT WIN32)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
 endif()
 
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
-#find_package(geographic_msgs REQUIRED)
+find_package(geographic_msgs REQUIRED)
 find_package(diagnostic_msgs REQUIRED)
 find_package(diagnostic_updater REQUIRED)
 find_package(geometry_msgs REQUIRED)
@@ -20,12 +24,12 @@ find_package(tf2_ros REQUIRED)
 find_package(Eigen3 REQUIRED)
 
 rosidl_generate_interfaces(${PROJECT_NAME}
-  #"srv/SetDatum.srv"
+  "srv/SetDatum.srv"
   "srv/SetPose.srv"
   DEPENDENCIES
     builtin_interfaces
     geometry_msgs
-    #geographic_msgs
+    geographic_msgs
     diagnostic_msgs
   ADD_LINTER_TESTS
 )
@@ -35,6 +39,7 @@ include_directories(
   include
   ${EIGEN3_INCLUDE_DIRS}
   ${rclcpp_INCLUDE_DIRS}
+  ${geographic_msgs_INCLUDE_DIRS}
   ${geometry_msgs_INCLUDE_DIRS}
   ${diagnostic_msgs_INCLUDE_DIRS}
   ${diagnostic_updater_INCLUDE_DIRS}
@@ -50,6 +55,7 @@ add_library(
   src/ukf.cpp
   src/filter_base.cpp
   src/filter_utilities.cpp
+  src/navsat_transform.cpp
   src/ros_filter.cpp
   src/ros_filter_utilities.cpp)
 
@@ -61,11 +67,17 @@ add_executable(
   src/se_node.cpp
 )
 
+add_executable(
+  navsat_transform_node
+  src/navsat_transform_node.cpp
+)
+
 target_link_libraries(
   rl_lib
   ${angles_LIBRARIES}
   ${EIGEN3_LIBRARIES}
   ${rclcpp_LIBRARIES}
+  ${geographic_msgs_LIBRARIES}
   ${geometry_msgs_LIBRARIES}
   ${diagnostic_msgs_LIBRARIES}
   ${diagnostic_updater_LIBRARIES}
@@ -81,10 +93,16 @@ target_link_libraries(
   rl_lib
 )
 
+target_link_libraries(
+  navsat_transform_node
+  rl_lib
+)
+
 install(
   TARGETS
   rl_lib
   se_node
+  navsat_transform_node
   DESTINATION lib/${PROJECT_NAME}
 )
 
@@ -95,7 +113,6 @@ install(
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
-
   find_package(ament_cmake_cppcheck REQUIRED)
   find_package(ament_cmake_cpplint REQUIRED)
   find_package(ament_cmake_lint_cmake REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,13 +28,13 @@ set(library_name rl_lib)
 set(libraries
   ${library_name} 
   ${angles_LIBRARIES}
-  ${EIGEN3_LIBRARIES}
-  ${rclcpp_LIBRARIES}
-  ${geographic_msgs_LIBRARIES}
-  ${geometry_msgs_LIBRARIES}
   ${diagnostic_msgs_LIBRARIES}
   ${diagnostic_updater_LIBRARIES}
+  ${EIGEN3_LIBRARIES}
+  ${geographic_msgs_LIBRARIES}
+  ${geometry_msgs_LIBRARIES}
   ${nav_msgs_LIBRARIES}
+  ${rclcpp_LIBRARIES}
   ${sensor_msgs_LIBRARIES}
   ${tf2_ros_LIBRARIES}
   ${tf2_LIBRARIES}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,37 @@ find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(Eigen3 REQUIRED)
 
+set(includes
+  include
+  ${EIGEN3_INCLUDE_DIRS}
+  ${rclcpp_INCLUDE_DIRS}
+  ${geographic_msgs_INCLUDE_DIRS}
+  ${geometry_msgs_INCLUDE_DIRS}
+  ${diagnostic_msgs_INCLUDE_DIRS}
+  ${diagnostic_updater_INCLUDE_DIRS}
+  ${nav_msgs_INCLUDE_DIRS}
+  ${sensor_msgs_INCLUDE_DIRS}
+  ${tf2_ros_INCLUDE_DIRS}
+  ${tf2_INCLUDE_DIRS}
+  ${tf2_geometry_msgs_INCLUDE_DIRS}
+)
+
+set(libraries
+  rl_lib
+  ${angles_LIBRARIES}
+  ${EIGEN3_LIBRARIES}
+  ${rclcpp_LIBRARIES}
+  ${geographic_msgs_LIBRARIES}
+  ${geometry_msgs_LIBRARIES}
+  ${diagnostic_msgs_LIBRARIES}
+  ${diagnostic_updater_LIBRARIES}
+  ${nav_msgs_LIBRARIES}
+  ${sensor_msgs_LIBRARIES}
+  ${tf2_ros_LIBRARIES}
+  ${tf2_LIBRARIES}
+  ${tf2_geometry_msgs_LIBRARIES}
+)
+
 rosidl_generate_interfaces(${PROJECT_NAME}
   "srv/SetDatum.srv"
   "srv/SetPose.srv"
@@ -36,18 +67,8 @@ rosidl_generate_interfaces(${PROJECT_NAME}
 
 include_directories(SYSTEM ${Eigen_INCLUDE_DIRS})
 include_directories(
-  include
-  ${EIGEN3_INCLUDE_DIRS}
-  ${rclcpp_INCLUDE_DIRS}
-  ${geographic_msgs_INCLUDE_DIRS}
-  ${geometry_msgs_INCLUDE_DIRS}
-  ${diagnostic_msgs_INCLUDE_DIRS}
-  ${diagnostic_updater_INCLUDE_DIRS}
-  ${nav_msgs_INCLUDE_DIRS}
-  ${sensor_msgs_INCLUDE_DIRS}
-  ${tf2_ros_INCLUDE_DIRS}
-  ${tf2_INCLUDE_DIRS}
-  ${tf2_geometry_msgs_INCLUDE_DIRS})
+  ${includes}
+)
 
 add_library(
   rl_lib
@@ -73,19 +94,7 @@ add_executable(
 )
 
 target_link_libraries(
-  rl_lib
-  ${angles_LIBRARIES}
-  ${EIGEN3_LIBRARIES}
-  ${rclcpp_LIBRARIES}
-  ${geographic_msgs_LIBRARIES}
-  ${geometry_msgs_LIBRARIES}
-  ${diagnostic_msgs_LIBRARIES}
-  ${diagnostic_updater_LIBRARIES}
-  ${nav_msgs_LIBRARIES}
-  ${sensor_msgs_LIBRARIES}
-  ${tf2_ros_LIBRARIES}
-  ${tf2_LIBRARIES}
-  ${tf2_geometry_msgs_LIBRARIES}
+  ${libraries}
 )
 
 target_link_libraries(
@@ -187,12 +196,6 @@ install(
   DESTINATION lib/${PROJECT_NAME})
 endif()
 
-ament_export_dependencies(rosidl_default_runtime)
-ament_export_include_directories(include)
-ament_export_libraries(
-  rl_lib
-)
-
 # Install params config files.
 install(DIRECTORY
   params
@@ -200,4 +203,7 @@ install(DIRECTORY
   DESTINATION share/${PROJECT_NAME}
 )
 
+ament_export_dependencies(rosidl_default_runtime)
+ament_export_include_directories(include)
+ament_export_libraries(rl_lib)
 ament_package()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,23 +23,10 @@ find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(Eigen3 REQUIRED)
 
-set(includes
-  include
-  ${EIGEN3_INCLUDE_DIRS}
-  ${rclcpp_INCLUDE_DIRS}
-  ${geographic_msgs_INCLUDE_DIRS}
-  ${geometry_msgs_INCLUDE_DIRS}
-  ${diagnostic_msgs_INCLUDE_DIRS}
-  ${diagnostic_updater_INCLUDE_DIRS}
-  ${nav_msgs_INCLUDE_DIRS}
-  ${sensor_msgs_INCLUDE_DIRS}
-  ${tf2_ros_INCLUDE_DIRS}
-  ${tf2_INCLUDE_DIRS}
-  ${tf2_geometry_msgs_INCLUDE_DIRS}
-)
+set(library_name rl_lib)
 
 set(libraries
-  rl_lib
+  ${library_name} 
   ${angles_LIBRARIES}
   ${EIGEN3_LIBRARIES}
   ${rclcpp_LIBRARIES}
@@ -67,11 +54,13 @@ rosidl_generate_interfaces(${PROJECT_NAME}
 
 include_directories(SYSTEM ${Eigen_INCLUDE_DIRS})
 include_directories(
-  ${includes}
+  include
+  ${EIGEN3_INCLUDE_DIRS}
+  ${diagnostic_updater_INCLUDE_DIRS}
 )
 
 add_library(
-  rl_lib
+  ${library_name}
   src/ekf.cpp
   src/ukf.cpp
   src/filter_base.cpp
@@ -80,7 +69,11 @@ add_library(
   src/ros_filter.cpp
   src/ros_filter_utilities.cpp)
 
-rosidl_target_interfaces(rl_lib
+ament_target_dependencies(${library_name} 
+  ${dependencies}
+)
+
+rosidl_target_interfaces(${library_name}
   ${PROJECT_NAME} "rosidl_typesupport_cpp")
 
 add_executable(
@@ -99,26 +92,27 @@ target_link_libraries(
 
 target_link_libraries(
   se_node
-  rl_lib
+  ${library_name}
 )
 
 target_link_libraries(
   navsat_transform_node
-  rl_lib
+  ${library_name}
+)
+
+ament_target_dependencies(
+  se_node
+  ${library_name}
+  ${dependencies}
 )
 
 install(
   TARGETS
   navsat_transform_node
-  rl_lib
   se_node
+  ${library_name}
   DESTINATION lib/${PROJECT_NAME}
 )
-
-#install(
-#  TARGETS ${PROJECT_NAME}
-#  DESTINATION lib/${PROJECT_NAME}
-#)
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
@@ -130,46 +124,46 @@ if(BUILD_TESTING)
 
   #### FILTER BASE TESTS ####
   ament_add_gtest(filter_base-test test/test_filter_base.cpp)
-  target_link_libraries(filter_base-test rl_lib)
+  target_link_libraries(filter_base-test ${library_name})
 
   # This test uses se_node node for convenience.
   ament_add_gtest(test_filter_base_diagnostics_timestamps
                     test/test_filter_base_diagnostics_timestamps.cpp)
-  target_link_libraries(test_filter_base_diagnostics_timestamps rl_lib)
+  target_link_libraries(test_filter_base_diagnostics_timestamps ${library_name})
   add_dependencies(test_filter_base_diagnostics_timestamps se_node)
 
   #### EKF TESTS ######
   ament_add_gtest(test_ekf test/test_ekf.cpp)
-  target_link_libraries(test_ekf rl_lib)
+  target_link_libraries(test_ekf ${library_name})
 
   ament_add_gtest(test_ekf_localization_node_interfaces
   test/test_ekf_localization_node_interfaces.cpp)
-  target_link_libraries(test_ekf_localization_node_interfaces rl_lib)
+  target_link_libraries(test_ekf_localization_node_interfaces ${library_name})
 
   ament_add_gtest(test_ekf_localization_node_bag1 test/test_localization_node_bag_pose_tester.cpp)
-  target_link_libraries(test_ekf_localization_node_bag1 rl_lib)
+  target_link_libraries(test_ekf_localization_node_bag1 ${library_name})
 
   ament_add_gtest(test_ekf_localization_node_bag2 test/test_localization_node_bag_pose_tester.cpp)
-  target_link_libraries(test_ekf_localization_node_bag2 rl_lib)
+  target_link_libraries(test_ekf_localization_node_bag2 ${library_name})
 
   ament_add_gtest(test_ekf_localization_node_bag3 test/test_localization_node_bag_pose_tester.cpp)
-  target_link_libraries(test_ekf_localization_node_bag3 rl_lib)
+  target_link_libraries(test_ekf_localization_node_bag3 ${library_name})
 
   #### UKF TESTS #####
   ament_add_gtest(test_ukf test/test_ukf.cpp)
-  target_link_libraries(test_ukf rl_lib)
+  target_link_libraries(test_ukf ${library_name})
 
   ament_add_gtest(test_ukf_localization_node_interfaces test/test_ukf_localization_node_interfaces.cpp)
-  target_link_libraries(test_ukf_localization_node_interfaces rl_lib)
+  target_link_libraries(test_ukf_localization_node_interfaces ${library_name})
 
   ament_add_gtest(test_ukf_localization_node_bag1 test/test_localization_node_bag_pose_tester.cpp)
-  target_link_libraries(test_ukf_localization_node_bag1 rl_lib)
+  target_link_libraries(test_ukf_localization_node_bag1 ${library_name})
 
   ament_add_gtest(test_ukf_localization_node_bag2 test/test_localization_node_bag_pose_tester.cpp)
-  target_link_libraries(test_ukf_localization_node_bag2 rl_lib)
+  target_link_libraries(test_ukf_localization_node_bag2 ${library_name})
 
   ament_add_gtest(test_ukf_localization_node_bag3 test/test_localization_node_bag_pose_tester.cpp)
-  target_link_libraries(test_ukf_localization_node_bag3 rl_lib)
+  target_link_libraries(test_ukf_localization_node_bag3 ${library_name})
 
   # This forces cppcheck to consider all files in this project to be C++,
   # including the headers which end with .h, which cppcheck would normally
@@ -179,21 +173,22 @@ if(BUILD_TESTING)
   ament_lint_cmake()
   ament_uncrustify()
 
-install(
-  TARGETS
-  filter_base-test
-  test_filter_base_diagnostics_timestamps
-  test_ekf
-  test_ekf_localization_node_interfaces
-  test_ekf_localization_node_bag1
-  test_ekf_localization_node_bag2
-  test_ekf_localization_node_bag3
-  test_ukf
-  test_ukf_localization_node_interfaces
-  test_ukf_localization_node_bag1
-  test_ukf_localization_node_bag2
-  test_ukf_localization_node_bag3
-  DESTINATION lib/${PROJECT_NAME})
+  install(
+    TARGETS
+    filter_base-test
+    test_filter_base_diagnostics_timestamps
+    test_ekf
+    test_ekf_localization_node_interfaces
+    test_ekf_localization_node_bag1
+    test_ekf_localization_node_bag2
+    test_ekf_localization_node_bag3
+    test_ukf
+    test_ukf_localization_node_interfaces
+    test_ukf_localization_node_bag1
+    test_ukf_localization_node_bag2
+    test_ukf_localization_node_bag3
+    DESTINATION lib/${PROJECT_NAME}
+  )
 endif()
 
 # Install params config files.
@@ -203,7 +198,7 @@ install(DIRECTORY
   DESTINATION share/${PROJECT_NAME}
 )
 
-ament_export_dependencies(rosidl_default_runtime)
 ament_export_include_directories(include)
-ament_export_libraries(rl_lib)
+ament_export_libraries(${library_name})
+ament_export_dependencies(${dependencies})
 ament_package()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,9 +100,9 @@ target_link_libraries(
 
 install(
   TARGETS
+  navsat_transform_node
   rl_lib
   se_node
-  navsat_transform_node
   DESTINATION lib/${PROJECT_NAME}
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ find_package(Eigen3 REQUIRED)
 set(library_name rl_lib)
 
 set(libraries
-  ${library_name} 
+  ${library_name}
   ${angles_LIBRARIES}
   ${diagnostic_msgs_LIBRARIES}
   ${diagnostic_updater_LIBRARIES}
@@ -69,7 +69,7 @@ add_library(
   src/ros_filter.cpp
   src/ros_filter_utilities.cpp)
 
-ament_target_dependencies(${library_name} 
+ament_target_dependencies(${library_name}
   ${dependencies}
 )
 
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -250,6 +250,11 @@ private:
   bool publish_gps_;
 
   /**
+   * @brief Queue size/QoS of publishers and subscribers
+   */
+  int queue_size_;
+
+  /**
    * @brief Transform buffer for managing coordinate transforms
    */
   tf2_ros::Buffer tf_buffer_;
@@ -345,9 +350,19 @@ private:
    */
   bool zero_altitude_;
 
+  /**
+   * @brief Private node member
+   */
   rclcpp::Node::SharedPtr node_;
+
+  /**
+   * @brief IMU Subscription
+   */
   rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub;
-  
+    
+  /**
+   * @brief Navsatfix publisher
+   */
   rclcpp::Publisher<sensor_msgs::msg::NavSatFix>::SharedPtr filtered_gps_pub;
 };
 
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -51,25 +51,25 @@
 namespace robot_localization
 {
 
-class NavSatTransform
+class NavSatTransform : public rclcpp::Node
 {
 public:
   /**
    * @brief Constructor
    */
-  NavSatTransform(std::shared_ptr<rclcpp::Node> node);
+  NavSatTransform(const rclcpp::NodeOptions &);
 
   /**
    * @brief Destructor
    */
   ~NavSatTransform();
 
-  /**
-   * @brief Main run loop
-   */
-  void run();
-
 private:
+  /**
+   * @brief Callback for computing and publish transform
+   */
+  void transformCallback();
+
   /**
    * @brief Computes the transform from the UTM frame to the odom frame
    */
@@ -167,10 +167,30 @@ private:
   bool broadcast_utm_transform_as_parent_frame_;
 
   /**
+   * @brief TimerBase for publish callback
+   */
+  rclcpp::Service<robot_localization::srv::SetDatum>::SharedPtr datum_srv;
+
+  /**
+   * @brief Navsatfix publisher
+   */
+  rclcpp::Publisher<sensor_msgs::msg::NavSatFix>::SharedPtr filtered_gps_pub;
+
+  /**
    * @brief The frame_id of the GPS message (specifies mounting location)
    */
   std::string gps_frame_id_;
 
+  /**
+   * @brief GPS odometry publisher
+   */
+  rclcpp::Publisher<nav_msgs::msg::Odometry>::SharedPtr gps_odom_pub;
+
+  /**
+   * @brief GPS Subscription
+   */
+  rclcpp::Subscription<sensor_msgs::msg::NavSatFix>::SharedPtr gps_sub;
+  
   /**
    * @brief Timestamp of the latest good GPS message
    *
@@ -203,6 +223,11 @@ private:
   bool has_transform_odom_;
 
   /**
+   * @brief IMU Subscription
+   */
+  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub;
+
+  /**
    * @brief Covariance for most recent odometry data
    */
   Eigen::MatrixXd latest_odom_covariance_;
@@ -227,6 +252,11 @@ private:
    * environment.
    */
   double magnetic_declination_;
+
+  /**
+   * @brief Odometry Subscription
+   */
+  rclcpp::Subscription<nav_msgs::msg::Odometry>::SharedPtr odom_sub;
 
   /**
    * @brief Timestamp of the latest good odometry message
@@ -265,9 +295,20 @@ private:
   bool transform_good_;
 
   /**
+   * @brief Timer
+   */
+  rclcpp::TimerBase::SharedPtr timer_;
+
+  /**
    * @brief Latest IMU orientation
    */
   tf2::Quaternion transform_orientation_;
+
+  /**
+   * @brief Parameter that specifies the how long we wait for a transform to
+   * become available.
+   */
+  tf2::Duration transform_timeout_;
 
   /**
    * @brief Holds the UTM pose that is used to compute the transform
@@ -332,33 +373,12 @@ private:
   double yaw_offset_;
 
   /**
-   * @brief Parameter that specifies the how long we wait for a transform to
-   * become available.
-   */
-  tf2::Duration transform_timeout_;
-
-  /**
    * @brief Whether or not to report 0 altitude
    *
    * If this parameter is true, we always report 0 for the altitude of the
    * converted GPS odometry message.
    */
   bool zero_altitude_;
-
-  /**
-   * @brief Private node member
-   */
-  rclcpp::Node::SharedPtr node_;
-
-  /**
-   * @brief IMU Subscription
-   */
-  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub;
-    
-  /**
-   * @brief Navsatfix publisher
-   */
-  rclcpp::Publisher<sensor_msgs::msg::NavSatFix>::SharedPtr filtered_gps_pub;
 };
 
 }  // namespace robot_localization
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -280,11 +280,6 @@ private:
   bool publish_gps_;
 
   /**
-   * @brief Custom Sensor Queue-size
-   */
-  int queue_size_;
-  
-  /**
    * @brief Transform buffer for managing coordinate transforms
    */
   tf2_ros::Buffer tf_buffer_;
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -174,7 +174,7 @@ private:
   /**
    * @brief Navsatfix publisher
    */
-  rclcpp::Publisher<sensor_msgs::msg::NavSatFix>::SharedPtr filtered_gps_pub;
+  rclcpp::Publisher<sensor_msgs::msg::NavSatFix>::SharedPtr filtered_gps_pub_;
 
   /**
    * @brief The frame_id of the GPS message (specifies mounting location)
@@ -184,12 +184,12 @@ private:
   /**
    * @brief GPS odometry publisher
    */
-  rclcpp::Publisher<nav_msgs::msg::Odometry>::SharedPtr gps_odom_pub;
+  rclcpp::Publisher<nav_msgs::msg::Odometry>::SharedPtr gps_odom_pub_;
 
   /**
    * @brief GPS Subscription
    */
-  rclcpp::Subscription<sensor_msgs::msg::NavSatFix>::SharedPtr gps_sub;
+  rclcpp::Subscription<sensor_msgs::msg::NavSatFix>::SharedPtr gps_sub_;
   
   /**
    * @brief Timestamp of the latest good GPS message
@@ -225,7 +225,7 @@ private:
   /**
    * @brief IMU Subscription
    */
-  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub;
+  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub_;
 
   /**
    * @brief Covariance for most recent odometry data
@@ -256,7 +256,7 @@ private:
   /**
    * @brief Odometry Subscription
    */
-  rclcpp::Subscription<nav_msgs::msg::Odometry>::SharedPtr odom_sub;
+  rclcpp::Subscription<nav_msgs::msg::Odometry>::SharedPtr odom_sub_;
 
   /**
    * @brief Timestamp of the latest good odometry message
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -169,7 +169,7 @@ private:
   /**
    * @brief TimerBase for publish callback
    */
-  rclcpp::Service<robot_localization::srv::SetDatum>::SharedPtr datum_srv;
+  rclcpp::Service<robot_localization::srv::SetDatum>::SharedPtr datum_srv_;
 
   /**
    * @brief Navsatfix publisher
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -57,7 +57,7 @@ public:
   /**
    * @brief Constructor
    */
-  NavSatTransform();
+  NavSatTransform(std::shared_ptr<rclcpp::Node> node_);
 
   /**
    * @brief Destructor
@@ -78,10 +78,10 @@ private:
   /**
    * @brief Callback for the datum service
    */
-  void datumCallback(
-    const std::shared_ptr<robot_localization::srv::set_datum::request>
+  bool datumCallback(
+    const std::shared_ptr<robot_localization::srv::SetDatum::Request>
     request,
-    std::shared_ptr<robot_localization::srv::set_datum::response>);
+    std::shared_ptr<robot_localization::srv::SetDatum::Response>);
 
   /**
    * @brief Given the pose of the navsat sensor in the UTM frame, removes the
@@ -107,19 +107,19 @@ private:
    * @brief Callback for the GPS fix data
    * @param[in] msg The NavSatFix message to process
    */
-  void gpsFixCallback(const sensor_msgs::msg::NavSatFix::ConstPtr & msg);
+  void gpsFixCallback(const sensor_msgs::msg::NavSatFix::SharedPtr msg);
 
   /**
    * @brief Callback for the IMU data
    * @param[in] msg The IMU message to process
    */
-  void imuCallback(const sensor_msgs::msg::Imu::ConstPtr & msg);
+  void imuCallback(const sensor_msgs::msg::Imu::SharedPtr msg);
 
   /**
    * @brief Callback for the odom data
    * @param[in] msg The odometry message to process
    */
-  void odomCallback(const nav_msgs::msg::Odometry::ConstPtr & msg);
+  void odomCallback(const nav_msgs::msg::Odometry::SharedPtr msg);
 
   /**
    * @brief Converts the odometry data back to GPS and broadcasts it
@@ -138,14 +138,14 @@ private:
    * transform
    * @param[in] msg The NavSatFix message to use in the transform
    */
-  void setTransformGps(const sensor_msgs::msg::NavSatFix::ConstPtr & msg);
+  void setTransformGps(const sensor_msgs::msg::NavSatFix::SharedPtr & msg);
 
   /**
    * @brief Used for setting the odometry data that will be used to compute the
    * transform
    * @param[in] msg The odometry message to use in the transform
    */
-  void setTransformOdometry(const nav_msgs::msg::Odometry::ConstPtr & msg);
+  void setTransformOdometry(const nav_msgs::msg::Odometry::SharedPtr & msg);
 
   /**
    * @brief Frame ID of the robot's body frame
@@ -345,7 +345,10 @@ private:
    */
   bool zero_altitude_;
 
-  rclcpp::node::Node::SharedPtr node_;
+  rclcpp::Node::SharedPtr node_;
+  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub;
+  
+  rclcpp::Publisher<sensor_msgs::msg::NavSatFix>::SharedPtr filtered_gps_pub;
 };
 
 }  // namespace robot_localization
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -57,7 +57,7 @@ public:
   /**
    * @brief Constructor
    */
-  NavSatTransform(std::shared_ptr<rclcpp::Node> node_);
+  NavSatTransform(std::shared_ptr<rclcpp::Node> node);
 
   /**
    * @brief Destructor
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -280,6 +280,11 @@ private:
   bool publish_gps_;
 
   /**
+   * @brief Custom Sensor Queue-size
+   */
+  int queue_size_;
+  
+  /**
    * @brief Transform buffer for managing coordinate transforms
    */
   tf2_ros::Buffer tf_buffer_;
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -250,11 +250,6 @@ private:
   bool publish_gps_;
 
   /**
-   * @brief Queue size/QoS of publishers and subscribers
-   */
-  int queue_size_;
-
-  /**
    * @brief Transform buffer for managing coordinate transforms
    */
   tf2_ros::Buffer tf_buffer_;
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -57,7 +57,7 @@ public:
   /**
    * @brief Constructor
    */
-  NavSatTransform(const rclcpp::NodeOptions &);
+  explicit NavSatTransform(const rclcpp::NodeOptions &);
 
   /**
    * @brief Destructor
@@ -190,7 +190,7 @@ private:
    * @brief GPS Subscription
    */
   rclcpp::Subscription<sensor_msgs::msg::NavSatFix>::SharedPtr gps_sub_;
-  
+
   /**
    * @brief Timestamp of the latest good GPS message
    *
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -442,11 +442,6 @@ protected:
   //!
   bool publish_transform_;
 
-
-  //! @brief Queue size for publish and subscribe
-  //!
-  int queue_size_;
-
   //! @brief Whether to reset the filters when backwards jump in time is
   //! detected
   //!
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -442,6 +442,11 @@ protected:
   //!
   bool publish_transform_;
 
+
+  //! @brief Queue size for publish and subscribe
+  //!
+  int queue_size_;
+
   //! @brief Whether to reset the filters when backwards jump in time is
   //! detected
   //!
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -324,8 +324,8 @@ protected:
   //! @brief Aggregates all diagnostics so they can be published
   //! @param[in] wrapper - The diagnostic status wrapper to update
   //!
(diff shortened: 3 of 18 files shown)


ayrton04:
Please see my notes on that issue.

test/test_ukf_localization_node_interfaces.cpp
```diff
@@ -583,14 +579,13 @@ TEST(InterfacesTest, ImuTwistBasicIO) {
...
-  rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
-  custom_qos_profile.depth = 5;
+  auto custom_qos_profile = rclcpp::SensorDataQoS();
   // publish and subscribe calls have been changed as per ros2
   auto imuPub = node_->create_publisher<sensor_msgs::msg::Imu>(
     "/imu_input1", custom_qos_profile);
 
   auto filteredSub = node_->create_subscription<nav_msgs::msg::Odometry>(
-    "/odometry/filtered", filterCallback);
+    "/odometry/filtered", 10, filterCallback);
```
SteveMacenski: This callback depth should probably be a parameter if you're going to add it.

src/navsat_transform.cpp
```diff
@@ -132,12 +130,12 @@ void NavSatTransform::run()
   }
 
   auto gps_odom_pub =
-    node_->create_publisher<nav_msgs::msg::Odometry>("odometry/gps");
+    node_->create_publisher<nav_msgs::msg::Odometry>("odometry/gps", 10);
```
SteveMacenski: This callback depth should probably be a parameter if you're going to add it? Same with all similar below.

src/ros_filter.cpp
```diff
@@ -1135,7 +1134,7 @@ void RosFilter::loadParams()
 
         topic_subs_.push_back(node_->create_subscription<
             geometry_msgs::msg::PoseWithCovarianceStamped>(
-            pose_topic, pose_callback));
+            pose_topic, 10, pose_callback));
```
SteveMacenski: I think this should be the `_queue_size` parameter depth. Same for all following uses of the same thing (so I dont make 10 comments of the same)

test/test_ukf_localization_node_interfaces.cpp
SteveMacenski: This is a test, ignore this

src/ros_filter.cpp
klintan: Thanks, yeah I have a bunch of changes coming in, so feel free to put off the review for just a while and I'll make it more complete. The PR was a bit pre-mature

SteveMacenski: Well small changes are good. I think most maintainers would prefer 10 small PRs vs 1 megaPR that does everything under the sun & hard to just visually check and say "yup looks good" vs needing extensive testing for how much content changed.

SteveMacenski: This is just parameterization declaration changes, EZ to just say "yup still compiles **merge**"

klintan: I def agree, so maybe I will rename this PR to reflect what the major thing is: getting navsat_transform_node to work.

klintan:
It now compiles with navsat_transform_node. However there is one major issue, trying to run I get problems with the dynamically linked stuff: 

```
dyld: Library not loaded: @loader_path/libnddsc.dylib
  Referenced from: /ros2/ros2-osx/lib/librosidl_typesupport_connext_c.dylib
  Reason: image not found
```
Which is really strange, because I couldn't see anything in particular in the CMakeLists file that would make a hard dependency on anything from connext (as it is an optional middleware).

I'll continue my research, but any input on that is welcome :). However I think its ready for some major PR feedback now.

Edit: Fixed!

src/navsat_transform.cpp
klintan: Sounds good, I'll fix it.

src/ros_filter.cpp
klintan: > I think this should be the `_queue_size` parameter depth. Same for all following uses of the same thing (so I dont make 10 comments of the same)

I'll get it fixed, thanks!

CMakeLists.txt
```diff
@@ -5,9 +5,13 @@ if(NOT WIN32)
...
 endif()
 
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
-#find_package(geographic_msgs REQUIRED)
+find_package(geographic_msgs REQUIRED)
```
ayrton04, line 14: Happy to see this was released.

CMakeLists.txt
```diff
@@ -81,10 +93,16 @@ target_link_libraries(
...
+target_link_libraries(
+  navsat_transform_node
+  rl_lib
+)
+
 install(
   TARGETS
   rl_lib
   se_node
+  navsat_transform_node
```
ayrton04, line 111: Nit: please alphabetise.

include/robot_localization/navsat_transform.hpp
```diff
@@ -57,7 +57,7 @@ class NavSatTransform
   /**
    * @brief Constructor
    */
-  NavSatTransform();
+  NavSatTransform(std::shared_ptr<rclcpp::Node> node_);
```
ayrton04: What do the ROS2 coding standards say about parameter variables names? Should this have a trailing underscore?

src/navsat_transform.cpp
```diff
@@ -51,22 +51,21 @@
 
 using std::placeholders::_1;
 using std::placeholders::_2;
+using namespace std::chrono_literals;
 
 namespace robot_localization
 {
-NavSatTransform::NavSatTransform()
+NavSatTransform::NavSatTransform(std::shared_ptr<rclcpp::Node> node_)
```
ayrton04: We'll leave this as-is for now, but in a future PR, I'd like to have these be one initialisation per line.

src/navsat_transform.cpp
```diff
@@ -271,8 +269,8 @@ void NavSatTransform::computeTransform()
 }
 
 bool NavSatTransform::datumCallback(
-  robot_localization::srv::set_datum::request & request,
-  robot_localization::srv::set_datum::response &)
+  std::shared_ptr<robot_localization::srv::SetDatum::Request> request,
```
ayrton04: Potentially stupid question: did the service types not expose a `::SharedPtr` alias?

src/navsat_transform_node.cpp
```diff
@@ -37,7 +37,9 @@ int main(int argc, char ** argv)
 {
   rclcpp::init(argc, argv);
 
-  robot_localization::NavSatTransform trans;
+  auto node_ = rclcpp::Node::make_shared("navsat_transform_node");
```
ayrton04: Same comment re: non-member variables. Do the new standards suggest that this should have a trailing underscore?

SteveMacenski: No, I think it was just a copy pasta error

include/robot_localization/navsat_transform.hpp
SteveMacenski: And I'm also looking at the [cpp file](https://github.com/cra-ros-pkg/robot_localization/blob/3fed5670769ae397222d64ac3c6f129e2c358140/src/navsat_transform.cpp#L58) and maybe someone can school me on smart pointers a little here, but I don't actually see where the `node_` that is an input (that should be `node`) is given to the `node_` that is the member variable (or does C++11 auto-copy when you make them the same name O.o that seems bad)

rotu:
> It now compiles with navsat_transform_node. However there is one major issue, trying to run I get problems with the dynamically linked stuff:

I'm experiencing these issues as well with an unrelated package. I suspect this is caused by something upstream.

CMakeLists.txt
klintan: I don't think it's released yet though. I saw the comment in the issue. Right now it needs to be cloned manually and activate that workspace. This is kind of important as well https://github.com/ros-geographic-info/geographic_info/pull/26 which hasn't been approved yet :(

SteveMacenski: uh, then we cant merge this @klintan, it'll not build in CI or on releasing debians

include/robot_localization/navsat_transform.hpp
klintan: I just thought it should be seen as the private nodehandle in ROS1. I'll rename it. 

Should the private node class member be called node_ ?  

Regarding the pointer, you are right it is not explicitly assigned anywhere will fix it. Not sure what the default behavior would be, it compiled because its declared, but I guess running it would have resulted in an exception ? (I'm fairly new to C++)

src/navsat_transform_node.cpp
klintan: Thanks, will fix it.

src/navsat_transform.cpp
klintan: > > It now compiles with navsat_transform_node. However there is one major issue, trying to run I get problems with the dynamically linked stuff:
> 
> I'm experiencing these issues as well with an unrelated package. I suspect this is caused by something upstream.

Let me know if you find the/a solution, I'm going to try just removing the dynamic links using otool/install_name_tool . I'll let you know if I figure it out.

klintan:
> I'm experiencing these issues as well with an unrelated package. I suspect this is caused by something upstream.

Let me know if you find the/a solution, I'm going to try just removing the dynamic links using otool/install_name_tool . I'll let you know if I figure it out.

@rotu the issue was indeed upstream, it seems that geographic_msgs was exporting all of these dependencies 
```
ament_export_dependencies(ament_cmake)
ament_export_dependencies(std_msgs)
ament_export_dependencies(geometry_msgs)
ament_export_dependencies(unique_identifier_msgs)
ament_export_dependencies(rosidl_default_generators)
ament_export_dependencies(rosidl_default_runtime)
```
Removing all of them except 
`ament_export_dependencies(rosidl_default_runtime)`
fixes that.

CMakeLists.txt
klintan: No worries, I'll try to push to get the geographic_msgs released, maybe we can keep this PR open until then (hopefully won't be months) and as soon as it's there, this will be mergeable.

edit: Sidenote, recently ROS2 converted ros-driver package for velodyne depends on the `angles` package https://github.com/ros/angles.git which is also not a part of the official release yet. But I get that different packages see this differently (of course regardless the geographic_info package at least needs the latest fixes in the ros2 branch)

src/navsat_transform.cpp
klintan: It did :) fixed!

include/robot_localization/navsat_transform.hpp
SteveMacenski: So member's have a trailing `_` in this code style, and inputs to members do not. So the constructor would look like `Name(type name)` and not `Name(type name_)` but the variable `name` is stored in is `name_` in the class

IanTheEngineer:
To test this PR, I made a vcstools file to include the unreleased upstream dependencies and open PR's:

`navsat_transform_node.repos :`
```YAML
repositories:
  diagnostics:
    type: git
    url: git@github.com:ros/diagnostics.git
    version: ros2-devel
  geographic_info:
    type: git
    url: git@github.com:PickNikRobotics/geographic_info.git
    version: unique_identifier_msgs
  robot_localization:
    type: git
    url: git@github.com:klintan/robot_localization.git
    version: dashing-devel
```
After installing ROS Dashing, and `python3-vcstools`,
```
$ mkdir -p ~/ros2_ws/src
$ cd ~/ros2_ws
$ vcs import src < navsat_transform_node.repos
$ source /opt/ros/dashing/setup.bash
$ colcon build

**Comment by IanTheEngineer** [comment]
The `package.xml` isn't in this PR, but it needs to have [geographic_msgs uncommented](https://github.com/klintan/robot_localization/blob/dashing-devel/package.xml#L21) for this PR to begin to build with colcon : 
```xml
-  <!--<depend>geographic_msgs</depend>-->
+  <depend>geographic_msgs</depend>
```

klintan:
> To test this PR, I made a vcstools file to include the unreleased upstream dependencies and open PR's:
> 
> `navsat_transform_node.repos :`
> 
> ```yaml
> repositories:
>   diagnostics:
>     type: git
>     url: git@github.com:ros/diagnostics.git
>     version: ros2-devel
>   geographic_info:
>     type: git
>     url: git@github.com:PickNikRobotics/geographic_info.git
>     version: unique_identifier_msgs
>   robot_localization:
>     type: git
>     url: git@github.com:klintan/robot_localization.git
>     version: dashing-devel
> ```
> 
> After installing ROS Dashing, and `python3-vcstools`,
> 
> ```
> $ mkdir -p ~/ros2_ws/src
> $ cd ~/ros2_ws
> $ vcs import src < navsat_transform_node.repos
> $ source /opt/ros/dashing/setup.bash
> $ colcon build
> ```

That's awesome, thanks!

CMakeLists.txt
SteveMacenski: *sigh* ok, now, weeks later, I have lobbied all the parties required to merge the ROS2 port PR, it will be released in the next dashing sync assuming I didnt mess it up. But now I have a new shiny GitHub organization on my profile, so not a total loss of time :-)  
https://github.com/ros/rosdistro/pull/22324

ayrton04:
Wow, thanks for all the effort here. I'll keep an eye out for the Dashing sync and merge afterwards, assuming nothing else needs to be done here.

SteveMacenski:
I still see some issues, will do another pass through this afternoon.

CMakeLists.txt
```diff
@@ -35,6 +39,7 @@ include_directories(
   include
```
SteveMacenski, line 57: I'm going to point you to one of my other packages for ros2 cmakelist best practices, this is really not easily maintainable with all the include / link dirs. If you set a variable with all the dependencies and you use ament, you can get rid of most of this code.

https://github.com/ros-planning/navigation2/blob/master/nav2_planner/CMakeLists.txt

include/robot_localization/navsat_transform.hpp
```diff
@@ -57,7 +57,7 @@ class NavSatTransform
   /**
    * @brief Constructor
    */
-  NavSatTransform();
+  NavSatTransform(std::shared_ptr<rclcpp::Node> node);
```
SteveMacenski: Best practices for this is that the NavSatTransform node should derive from rclcpp::Node, and when you make the shared_ptr of it in `main` that will deal with this, then all of your calls are working with methods available vs storing your `node_`. The ros nodehandle methodology is depreciated in ros2, its preferred to derive from the node to best make use of lifecycle and components, which is important for this package in particular

src/navsat_transform.cpp
```diff
@@ -80,7 +94,7 @@ void NavSatTransform::run()
   double transform_timeout = 0.0;
 
   // Load the parameters we need
-  magnetic_declination_ = node_->get_parameter("magnetic_declination_radians");
+  magnetic_declination_ = node_->declare_parameter("magnetic_declination_radians", 0.0);
```
SteveMacenski: You added the declare parameter to a couple, but you didn't update any of the `get_parameter_or` which don't declare variables, they all need to be updated

include/robot_localization/ros_filter.hpp
```diff
@@ -442,6 +442,11 @@ class RosFilter
   //!
   bool publish_transform_;
 
+
+  //! @brief Queue size for publish and subscribe
+  //!
+  int queue_size_;
```
SteveMacenski: why was this added into ros filter?

src/navsat_transform.cpp
```diff
@@ -153,7 +168,7 @@ void NavSatTransform::run()
 
       if (transform_good_ && !use_odometry_yaw_ && !use_manual_datum_) {
         // Once we have the transform, we don't need the IMU
-        imu_sub.reset(nullptr);
+        imu_sub = nullptr;
```
SteveMacenski: this is leaking memory, please reset the smart pointer

src/navsat_transform.cpp
```diff
@@ -253,7 +268,7 @@ void NavSatTransform::computeTransform()
     // it.
     if (broadcast_utm_transform_) {
       geometry_msgs::msg::TransformStamped utm_transform_stamped;
-      utm_transform_stamped.header.stamp = rclcpp::Time::now();
+      utm_transform_stamped.header.stamp = rclcpp::Clock().now();
```
SteveMacenski: You have a `node_`, you should use it

src/navsat_transform.cpp
```diff
@@ -283,15 +298,15 @@ bool NavSatTransform::datumCallback(
...
 
   sensor_msgs::msg::NavSatFix fix;
-  fix.latitude = request.geo_pose.position.latitude;
-  fix.longitude = request.geo_pose.position.longitude;
-  fix.altitude = request.geo_pose.position.altitude;
-  fix.header.stamp = rclcpp::Time::now();
+  fix.latitude = request->geo_pose.position.latitude;
+  fix.longitude = request->geo_pose.position.longitude;
+  fix.altitude = request->geo_pose.position.altitude;
+  fix.header.stamp = rclcpp::Clock().now();
```
SteveMacenski: use the clock in your `node_`

src/ros_filter.cpp
```diff
@@ -1017,13 +1029,12 @@ void RosFilter::loadParams()
             pose_callback_data, twist_callback_data);
 
         topic_subs_.push_back(
-          node_->create_subscription<nav_msgs::msg::Odometry>(odom_topic,
+          node_->create_subscription<nav_msgs::msg::Odometry>(odom_topic, queue_size_,
```
SteveMacenski: These topics have `_queue_size` parameters which should be used, like the differential / relative parameters you update below

src/ros_filter.cpp
```diff
@@ -1737,6 +1748,7 @@ void RosFilter::run()
     // their received measurements
     rclcpp::spin_some(node_);
     cur_time = node_->now();
+    std::cout << "running node spin some" << std::endl;
```
SteveMacenski: Delete this print out, if its necessary then it should use ros logging.

src/ros_filter.cpp
```diff
@@ -1826,11 +1838,12 @@ void RosFilter::run()
       }
 
       // Fire off the position and the transform
+      std::cout << "publish" << std::endl;
```
SteveMacenski: As well on the print outs

src/navsat_transform.cpp
```diff
@@ -93,9 +107,11 @@ void NavSatTransform::run()
   node_->get_parameter_or("frequency", frequency, 10.0);
   node_->get_parameter_or("delay", delay, 0.0);
   node_->get_parameter_or("transform_timeout", transform_timeout, 0.0);
+  queue_size_ = node_->declare_parameter("queue_size", 10);
```
SteveMacenski: You should make clear what this queue size is for, since there's alot of parameterizable queue sizes in this package. 

Edit: looking down the review I think this is mostly a miscommunication on how the `XYZ_queue_size` parameter being used. In any case, all the member variables `queue_size_` introduced should be removed

SteveMacenski (changes requested):
Biggest issues around the queue_size & ros2 conventions

include/robot_localization/navsat_transform.hpp
klintan: I actually tried to use that convention here as well, but I couldn't get it to work, always got the bad_weak_ptr exception during runtime because of the TF2 package (I think). The initializer lists wants a shared, and since there is now owner when it's passed I got that exception. However it could have been the way I used image_transport the "old" way, instead of the `image_transport::create_subscription/create_publisher` that made it fail, I can't fully recall I'll need to retest it.

include/robot_localization/ros_filter.hpp
klintan: Good question, will fix!

src/navsat_transform.cpp
klintan: Thanks, good catch. Not even sure why I changed it.. ?

src/navsat_transform.cpp
klintan: Thanks, yes good catch, this could probably introduce the "different clock source error" as well.

src/navsat_transform.cpp
klintan: ok will remove

klintan:
I thought some of the changes from the last PR was going to conflict with this one? Regardless I'll try to adopt the same style as that one for some of the changes here as well (and of course all of the review remarks)

Thanks for the very thorough review, I appreciate it and I appreciate your patience.

include/robot_localization/navsat_transform.hpp
SteveMacenski: If you `make_shared` the NavSatTransform object in `main` that will fix that. You get an issue with shared_ptrs if there's not already the reference counter set

SteveMacenski:
> thorough

I think some of my colleagues would use another, less kind, word in its place, but thanks for baring with me :smile_cat:

src/navsat_transform.cpp
klintan: Fixed!

klintan:
Fixed a bunch of stuff, I will give the inherit from `rclcpp::Node` another try as well, hopefully I'll have time to get to that tomorrow night.

klintan:
@SteveMacenski Now compiling and working using the ROS2 node-style. 

Made most of the std::outs to RCLCPP style logging. On of the old std::outs outputted the pose though, which is kind of tricky to log using RCLCPP style. I don't think that one is necessary though, but maybe long term for debugging purposes/brevity we want to get it in there.

Let me know if you find anything, any and all feedback is welcome :)

src/navsat_transform_node.cpp
```diff
@@ -37,11 +37,14 @@ int main(int argc, char ** argv)
...
-
-  robot_localization::NavSatTransform trans(node);
-
-  trans.run();
+  rclcpp::executors::SingleThreadedExecutor exec;
+  
+  const rclcpp::NodeOptions options;
+  auto navsat_transform_node = std::make_shared<robot_localization::NavSatTransform>(options);
+  
+  exec.add_node(navsat_transform_node);
```
SteveMacenski: if its just 1 node to spin you can `rclcpp::spin(navsat_transform_node->get_node_base_interface());` without the executor unless you want multi-threaded

src/navsat_transform.cpp
```diff
@@ -67,52 +68,46 @@ NavSatTransform::NavSatTransform(std::shared_ptr<rclcpp::Node> node) :
...
-  transform_timeout = node_->declare_parameter("transform_timeout", 0.0);
+  magnetic_declination_ = this->declare_parameter("magnetic_declination_radians", 0.0);
+  yaw_offset_ = this->declare_parameter("yaw_offset", 0.0);
+  broadcast_utm_transform_ = this->declare_parameter("broadcast_utm_transform", false);
+  broadcast_utm_transform_as_parent_frame_ = this->declare_parameter("broadcast_utm_transform_as_parent_frame", false);
+  zero_altitude_ = this->declare_parameter("zero_altitude", false);
+  publish_gps_ = this->declare_parameter("publish_filtered_gps", true);
+  use_odometry_yaw_ = this->declare_parameter("use_odometry_yaw", false);
+  use_manual_datum_ = this->declare_parameter("wait_for_datum", false);
+  frequency = this->declare_parameter("frequency", 10.0);
```
SteveMacenski: You have frequency defined above, use that variable. Same with delay/timeout.

ei `   frequency = this->declare_parameter("frequency", frequency); `

src/navsat_transform.cpp
```diff
@@ -134,54 +129,57 @@ void NavSatTransform::run()
...
   }
 
   // Sleep for the parameterized amount of time, to give
   // other nodes time to start up (not always necessary)
   rclcpp::sleep_for(std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::duration<double>(delay)));
 
-  rclcpp::Rate rate(frequency);
-  while (rclcpp::ok()) {
-    rclcpp::spin_some(node_);
+  int interval = (1/frequency) * 1000;
```
SteveMacenski: you have frequency defaulted to `10`, so this would be... 100 hz? I think you mean to move `s` to `ms` but the variable name didnt make that clear, can you change the name or do the ms conversion in the chrono call? Or better yet, use `chrono::seconds(interval)`

SteveMacenski (approved):
Some small fish changes needed

src/navsat_transform.cpp
klintan: Yes, good catch. I'll keep the variable name, but I like the approach of  `chrono::seconds(interval)`
to make it more clear.

klintan:
@SteveMacenski fixed the things you noted (I hope :) ). However I noticed another issue when fixing that, line 954 in ros_filter.cpp (same thing in 2 other places in ros_filter), perhaps you can take an extra look at that so it looks sane. I'm open to any other approach.

Sorry for the delay.

reso1:
Hello @klintan, I've been using your "beta version" of the navsat_transform node, and I found a problem when using a **dual_ekf_navsat** configurations, a local ekf node and a global ekf node with GPS (following the yaml below: [dual_ekf_navsat_example.yaml](https://github.com/klintan/robot_localization/blob/dashing-devel/params/dual_ekf_navsat_example.yaml)

SO here is the problem, the **navsat_transform** node subscribes to **"/odometry/filtered" topic msgs**, and because both "ekf_se_odom"(local) node and "ekf_se_map"(global) node will publish "/odometry/filtered" topic msg, it will cause the **"/odometry/gps" topic msg** published by navsat_transform node jump between 2 values, each calculated based on the msg published by "ekf_se_odom"(local) node or and "ekf_se_map"(global) node, respectively.
 
image
> from this workflow figure, the **navsat_transfrom_node** only takes the  **"/odometry/filtered"** msg published by **”*kf_localization_node(map)“** into account.

But anyway I've made a workaround, by only executing "odomCallback" function when the frame_id of the "/odometry/filtered" msg header eqauls to "map"(i.e. published by "ekf_se_map"(global) node).

I'm wondering that whether I'm not correctly using the "dual_ekf_navsat" configurations with GPS info, or is there a better solution for this problem? Thanks!

ayrton04:
@reso1, the two EKFs should *not* be publishing on the same topic. Remap the tier 1 (odom) frame one to something else.

SteveMacenski:
@klintan I’ll take a look this afternoon. 

I (still, frustratingly, annoyingly, {other more colorful words}) do not have a robot to meaningfully test this on, so by in large I’m trusting beyond the code semantics that this actually works. 

I’m not sure if Tom has a ROS2-able robot to have a quick third party check, or @reso1 other than your topic remapping (see Tom’s comment), is this working for you? Have you tried all the params getting to you & outputs as expected?

CMakeLists.txt
```diff
@@ -19,37 +23,60 @@ find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(Eigen3 REQUIRED)
 
+set(includes
```
SteveMacenski: This isn't necessary anymore, use `ament_target_dependencies` see examples here https://github.com/ros-planning/navigation2/blob/master/nav2_planner/CMakeLists.txt

CMakeLists.txt
```diff
@@ -19,37 +23,60 @@ find_package(tf2_geometry_msgs REQUIRED)
...
+  ${diagnostic_updater_INCLUDE_DIRS}
+  ${nav_msgs_INCLUDE_DIRS}
+  ${sensor_msgs_INCLUDE_DIRS}
+  ${tf2_ros_INCLUDE_DIRS}
+  ${tf2_INCLUDE_DIRS}
+  ${tf2_geometry_msgs_INCLUDE_DIRS}
+)
+
+set(libraries
+  rl_lib
```
SteveMacenski: As well here -- you're getting closer and closer on the current ROS2 cmake practices. I know it was a pain in earlier releases, but its pretty streamlined now.

CMakeLists.txt
```diff
@@ -170,17 +196,14 @@ install(
...
-)
-
 # Install params config files.
 install(DIRECTORY
   params
   launch
   DESTINATION share/${PROJECT_NAME}
 )
 
+ament_export_dependencies(rosidl_default_runtime)
```
SteveMacenski: To be safe, you should just export all the dependencies, again referencing to the `nav2_planner` I wrote

include/robot_localization/navsat_transform.hpp
```diff
@@ -166,11 +166,31 @@ class NavSatTransform
    */
   bool broadcast_utm_transform_as_parent_frame_;
 
+  /**
+   * @brief TimerBase for publish callback
+   */
+  rclcpp::Service<robot_localization::srv::SetDatum>::SharedPtr datum_srv;
```
SteveMacenski: need the trailing `_` for class variables by the projects style. Same with `filtered_gps_pub`, `gps_odom_pub` `gps_sub` `imu_sub` `odom_sub`

src/navsat_transform.cpp
```diff
@@ -51,55 +51,63 @@
...
+  yaw_offset_ = this->declare_parameter("yaw_offset", 0.0);
+  broadcast_utm_transform_ = this->declare_parameter("broadcast_utm_transform", false);
+  broadcast_utm_transform_as_parent_frame_ = this->declare_parameter("broadcast_utm_transform_as_parent_frame", false);
+  zero_altitude_ = this->declare_parameter("zero_altitude", false);
+  publish_gps_ = this->declare_parameter("publish_filtered_gps", true);
+  use_odometry_yaw_ = this->declare_parameter("use_odometry_yaw", false);
+  use_manual_datum_ = this->declare_parameter("wait_for_datum", false);
+  frequency = this->declare_parameter("frequency", frequency);
+  delay = this->declare_parameter("delay", delay);
+  transform_timeout = this->declare_parameter("transform_timeout", 0.0);
```
SteveMacenski: transform_timeout default is defined above.

src/navsat_transform.cpp
```diff
@@ -110,66 +118,68 @@ void NavSatTransform::run()
...
+    request->geo_pose.orientation = tf2::toMsg(quat);
+    auto response = std::make_shared<robot_localization::srv::SetDatum::Response>();
     datumCallback(request, response);
   }
 
-  auto odom_sub = node_->create_subscription<nav_msgs::msg::Odometry>(
-    "odometry/filtered", std::bind(&NavSatTransform::odomCallback, this, _1));
-  auto gps_sub = node_->create_subscription<sensor_msgs::msg::NavSatFix>(
-    "gps/fix", std::bind(&NavSatTransform::gpsFixCallback, this, _1));
+  odom_sub = this->create_subscription<nav_msgs::msg::Odometry>(
```
SteveMacenski: The ROS1 version of this has the queue size set to `1`, I'd follow in line with that rather than `10` for this, gpd sub and IMU sub

src/navsat_transform.cpp
```diff
@@ -110,66 +118,68 @@ void NavSatTransform::run()
...
   if (publish_gps_) {
     filtered_gps_pub =
-      node_->create_publisher<sensor_msgs::msg::NavSatFix>("gps/filtered");
+      this->create_publisher<sensor_msgs::msg::NavSatFix>("gps/filtered", rclcpp::QoS(10));
   }
 
   // Sleep for the parameterized amount of time, to give
   // other nodes time to start up (not always necessary)
-  rclcpp::utilities::sleep_for(std::chrono::seconds(delay));
+  rclcpp::sleep_for(std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::duration<double>(delay)));
```
SteveMacenski: You changed this from `seconds` to `nanoseconds` for the unit on `delay`, is that intentional? I think seconds was the more logical unit. Open for comments

src/navsat_transform.cpp
```diff
@@ -110,66 +118,68 @@ void NavSatTransform::run()
...
 
-  rclcpp::rate::Rate rate(frequency);
-  while (rclcpp::ok()) {
-    rclcpp::spin_some(node_);
+  int interval = (1/frequency);
```
SteveMacenski: `static_cast<int>(1.0/frequency)`

src/navsat_transform.cpp
```diff
@@ -242,18 +250,13 @@ void NavSatTransform::computeTransform()
...
-      "\n";
-
     transform_good_ = true;
 
     // Send out the (static) UTM transform in case anyone else would like to use
     // it.
     if (broadcast_utm_transform_) {
       geometry_msgs::msg::TransformStamped utm_transform_stamped;
-      utm_transform_stamped.header.stamp = rclcpp::Time::now();
+      utm_transform_stamped.header.stamp = this->get_clock()->now();
```
SteveMacenski: You can actually just do `this->now()`

src/navsat_transform.cpp
```diff
@@ -283,15 +286,15 @@ bool NavSatTransform::datumCallback(
...
 
   sensor_msgs::msg::NavSatFix fix;
-  fix.latitude = request.geo_pose.position.latitude;
-  fix.longitude = request.geo_pose.position.longitude;
-  fix.altitude = request.geo_pose.position.altitude;
-  fix.header.stamp = rclcpp::Time::now();
+  fix.latitude = request->geo_pose.position.latitude;
+  fix.longitude = request->geo_pose.position.longitude;
+  fix.altitude = request->geo_pose.position.altitude;
+  fix.header.stamp = this->get_clock()->now();
```
SteveMacenski: `this->now()`

src/ros_filter.cpp
```diff
@@ -634,22 +644,21 @@ void RosFilter::loadParams()
   twist_var_counts[StateMemberVyaw] = 0;
 
   // Determine if we'll be printing diagnostic information
-  node_->get_parameter_or("print_diagnostics", print_diagnostics_, false);
+  print_diagnostics_ = node_->declare_parameter("print_diagnostics", false);
 
   // Check for custom gravitational acceleration value
-  node_->get_parameter("gravitational_acceleration",
-    gravitational_acceleration_);
+  gravitational_acceleration_ = node_->declare_parameter("gravitational_acceleration",
```
SteveMacenski, line 651: This can change back to `gravitational_acceleration_` since that's set to 9.8... in the constructor

src/ros_filter.cpp
```diff
@@ -909,7 +910,7 @@ void RosFilter::loadParams()
   // Create a subscriber for manually setting/resetting pose
   set_pose_sub_ =
     node_->create_subscription<geometry_msgs::msg::PoseWithCovarianceStamped>(
-    "set_pose",
+    "set_pose", 10,
```
SteveMacenski: ROS1 was `1`, I'd go back to that

src/ros_filter.cpp
```diff
@@ -950,17 +951,21 @@ void RosFilter::loadParams()
     ss << "odom" << topic_ind++;
     std::string odom_topic_name = ss.str();
     std::string odom_topic;
-    more_params = node_->get_parameter(odom_topic_name, odom_topic);
+    node_->declare_parameter(odom_topic_name);
+
+    rclcpp::Parameter parameter;
```
SteveMacenski, line 961: Can you explain what's happening here? I see some gets but no declares and the 2 `get_parameter` calls look identical

src/ros_filter.cpp
```diff
@@ -1017,13 +1018,12 @@ void RosFilter::loadParams()
             pose_callback_data, twist_callback_data);
 
         topic_subs_.push_back(
-          node_->create_subscription<nav_msgs::msg::Odometry>(odom_topic,
+          node_->create_subscription<nav_msgs::msg::Odometry>(odom_topic, 10,
```
SteveMacenski: Isn't this a parameter for the `odom_N: queue_size`?

src/ros_filter.cpp
```diff
@@ -1135,7 +1138,7 @@ void RosFilter::loadParams()
 
         topic_subs_.push_back(node_->create_subscription<
             geometry_msgs::msg::PoseWithCovarianceStamped>(
-            pose_topic, pose_callback));
+            pose_topic, 10, pose_callback));
```
SteveMacenski: As well here, you have declares for other related parameters `_relative`, you need one for `_queue_size`

src/ros_filter.cpp
```diff
@@ -1212,7 +1221,7 @@ void RosFilter::loadParams()
 
         topic_subs_.push_back(node_->create_subscription<
             geometry_msgs::msg::TwistWithCovarianceStamped>(
-            twist_topic, twist_callback));
+            twist_topic, 10, twist_callback));
```
SteveMacenski: Queue size stuff

src/ros_filter.cpp
```diff
@@ -1362,7 +1371,7 @@ void RosFilter::loadParams()
             twist_callback_data, accel_callback_data);
 
         topic_subs_.push_back(node_->create_subscription<sensor_msgs::msg::Imu>(
```
SteveMacenski, line 1411: queue size

src/ros_filter.cpp
```diff
@@ -1436,7 +1445,7 @@ void RosFilter::loadParams()
       deceleration_gains);
 
     control_sub_ = node_->create_subscription<geometry_msgs::msg::Twist>(
-      "cmd_vel",
+      "cmd_vel", 10,
```
SteveMacenski: queue size

src/ros_filter.cpp
SteveMacenski: Extra print still?

SteveMacenski (changes requested):
Themes:
- still need queue sizes
- A few more updates to CMakeList still necessary 
- Member variable naming 
- Just making sure defaults are same as in ROS1

reso1:
@ayrton04 @SteveMacenski 
Thanks for your advices, I've tried the remapping solution and now it works fine :)

src/ros_filter.cpp
klintan: This parameter would be outside the scope of odom, twist, pose and imu. Do we want to declare a separate parameter for this one?

CMakeLists.txt
klintan: Hm I think I might have missed something, because couldn't get it to find some includes (eigen3 and diagnostics_updated without adding them). 

See my changes, I think I need some help on that one.

SteveMacenski: Sure let me know

src/ros_filter.cpp
SteveMacenski: Oh sorry this one is fine. I went a little trigger happy when I saw the number `10` :wink: Though I think `1` is also the ROS1 default, if you see how its used, there's no benefit to storing anything except the most recent one

CMakeLists.txt
SteveMacenski: Of everything I'm the least concerned by the CMake stuff, give it another shot and if not I can take care of it for you after the fact

src/ros_filter.cpp
klintan: So my thinking was this: 
`get_parameter` does not cast to bool anymore, and we also need to `declare_parameter` before using it. 

The code 
```
node_->declare_parameter(odom_topic_name);

    rclcpp::Parameter parameter;
    if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(odom_topic_name, parameter)) {
      more_params = true;
      odom_topic = node_->get_parameter(odom_topic_name, parameter);
    };
```
will first declare an empty `odom_topic_name` which will get `rclcpp::PARAMETER_NOT_SET` if not provided in the config. 
The conditional checks if it's set, if so it will set `more_params` to `true` and also assign the string `odom_topic` the name of the topic for `odom_x`.

There might be a nicer way to get rid of one `get_parameter` call, not sure how.

Ok a tiny bit convoluted explanation, but in short it's an attempt to reproduce the assigning to bool if a parameter is set or not.

SteveMacenski: Well the only bad thing is you're now declaring the `_N+1`th parameter when it doesnt exist. I'd say you could make an `else` to `undeclare` it, but that's not good posture either. 

There are documented parameters for `  node_options.allow_undeclared_parameters(true);` in the node options object to grab parameters from a file before declaring them, maybe this is a good use-case for it. I'd leave it up to @ayrton04 if he thinks that declaring, checking, and then undeclaring is an acceptable workaround. It seems a little hacky for my blood but I'd accept it to get this PR merged with a ticket to fix later

src/ros_filter.cpp
klintan: haha :) ok I'll change this to 1 as well.

src/ros_filter.cpp
klintan: I tend to agree that this feels way to hacky, although I also did not want to use the `node_options.allow_undeclared_parameters(true);` as it seems like a good candidate for future deprecation. 
I'm fine either way, we'll see what Tom says, perhaps a later ticket for fix is good enough for now.

SteveMacenski: Actually from a discussion in another thread with Tully, Ithink that undeclared parameters thing is going to stay for good. They claim there's a few good reasons for it. 

If we keep it this way, if its unset, you should `undeclare_parameter` that parameters so that when the user goes `ros2 topic list` it doesnt appear

klintan:
> @ayrton04 @SteveMacenski
> Thanks for your advices, I've tried the remapping solution and now it works fine :)

Happy to hear it's working for you :)

SteveMacenski:
Wow that was quick, ok I think the last thing is just the conversation on the parameters for checking `_N+1`

I'm not buying this `set(libraries ...)` call as necessary,but if that's the worst thing in here, I can live with that.

src/ros_filter.cpp
klintan: I think I might have come across that thread, at least a very similar discussion.

I added the else condition for undeclaring the parameters.

SteveMacenski (approved):
With that, I give this my (fairly meaningless) thumbs up.

Lets see what Tom says on the `undeclare` thing. Otherwise, assuming that you've sufficiently tested this since it doesn't appear the automated testing pipeline is in good shape, I would be OK merging this. 

Thanks with sticking with my rounds-and-rounds of nitpicks

src/ros_filter.cpp
```diff
@@ -1849,11 +1881,11 @@ void RosFilter::run()
 
     double diag_duration = (cur_time - last_diag_time).nanoseconds();
     if (print_diagnostics_ &&
-    		(diag_duration >= diagnostic_updater_.getPeriod() ||
-    				diag_duration < 0.0))
+        (diag_duration >= diagnostic_updater_.getPeriod() ||
```
blacksoul000: diagnostic_updater_.getPeriod().nanoseconds()
?

klintan: getPeriod() is already returning nanoseconds :)

edit: as seen below, I didn't have the latest diagnostics_updater, thanks for pointing this out!

klintan:
> With that, I give this my (fairly meaningless) thumbs up.
> 
> Lets see what Tom says on the `undeclare` thing. Otherwise, assuming that you've sufficiently tested this since it doesn't appear the automated testing pipeline is in good shape, I would be OK merging this.
> 
> Thanks with sticking with my rounds-and-rounds of nitpicks

Thank you for your patience. Sounds good, @reso1 seems to have successfully gotten everything going. And I did try it out before, but I haven't tested all functions after these latest changes.

I'll try to get it back up and running, maybe someone else gets it running for testing.

CMakeLists.txt
```diff
@@ -19,13 +23,31 @@ find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(Eigen3 REQUIRED)
 
+set(library_name rl_lib)
+
+set(libraries
```
ayrton04, line 28: Non-blocking: maybe alphabetise these?

include/robot_localization/navsat_transform.hpp
```diff
@@ -51,25 +51,25 @@
 namespace robot_localization
 {
 
-class NavSatTransform
+class NavSatTransform : public rclcpp::Node
```
ayrton04, line 54: I need to brush up on my ROS2, but it's interesting that the class now inherits from `rclcpp::Node`.

include/robot_localization/ukf.hpp
```diff
@@ -64,8 +64,8 @@ class Ukf : public FilterBase
    * args[2] contains the beta parameter.
    */
   explicit Ukf(
-    const bool alpha = 0.001, const bool kappa = 0.0,
-    const bool beta = 2.0);
+    const double alpha = 0.001, const double kappa = 0.0,
```
ayrton04, line 67: Yeaaaah, I fixed this in the ROS1 branches, I believe. I hang my head in shame.

src/navsat_transform.cpp
```diff
@@ -51,55 +51,63 @@
...
 using std::placeholders::_2;
+using namespace std::chrono_literals;
 
 namespace robot_localization
 {
-NavSatTransform::NavSatTransform()
-: magnetic_declination_(0.0), utm_odom_tf_yaw_(0.0), yaw_offset_(0.0),
-  transform_timeout_(0.0), broadcast_utm_transform_(false),
+NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options) : 
+  Node("navsat_transform_node", options),
```
ayrton04: Much more readable, thank you.

src/navsat_transform.cpp
```diff
@@ -51,55 +51,63 @@
...
-  node_->get_parameter_or("broadcast_utm_transform_as_parent_frame",
-    broadcast_utm_transform_as_parent_frame_, false);
-  node_->get_parameter_or("zero_altitude", zero_altitude_, false);
-  node_->get_parameter_or("publish_filtered_gps", publish_gps_, false);
-  node_->get_parameter_or("use_odometry_yaw", use_odometry_yaw_, false);
-  node_->get_parameter_or("wait_for_datum", use_manual_datum_, false);
-  node_->get_parameter_or("frequency", frequency, 10.0);
-  node_->get_parameter_or("delay", delay, 0.0);
-  node_->get_parameter_or("transform_timeout", transform_timeout, 0.0);
+  magnetic_declination_ = this->declare_parameter("magnetic_declination_radians", 0.0);
```
ayrton04, line 92: So parameters are one place I am _really_ out of the loop for ROS2. I added the ROS2 feature to support parameter arrays, but lost touch with it after that. I know ROS2 handles parameters at the node level, so I'm guessing this is the canonical way to define what parameters a node supports.

src/ros_filter.cpp
```diff
@@ -950,17 +951,23 @@ void RosFilter::loadParams()
...
     std::string odom_topic;
-    more_params = node_->get_parameter(odom_topic_name, odom_topic);
+    node_->declare_parameter(odom_topic_name);
+
+    rclcpp::Parameter parameter;
+    if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(odom_topic_name, parameter)) {
+      more_params = true;
+      odom_topic = node_->get_parameter(odom_topic_name, parameter);
+    } else {
+        node_->undeclare_parameter(odom_topic_name);
```
ayrton04: I feel like we're missing a `more_params = false;` statement here. The previous code would switch the state of `more_params`, but in this loop, once `more_params` is set to _true_, I don't see how it gets set back to false. If I'm right, you'll want to double-check the others.

ayrton04 (approved):
First, thank you very much for taking the time to do this, and to all involved in reviewing it. I haven't had any cycles to dedicate to this package, which pains me, but it's nice to see people making progress with it.

My approval is dependent on the issue I raised with `more_params`. If I'm off base on that, consider it approved.

include/robot_localization/navsat_transform.hpp
SteveMacenski: Yeah, ROS2 becomes pretty intrusive unfortunately.

src/navsat_transform.cpp
SteveMacenski: Yeah, so you have to declare variables and their types (or implied by defaults) for them to be get/set-able.

src/ros_filter.cpp
Jconn: I am looking at the diagnostic [repo](https://github.com/ros/diagnostics), and getPeriod() returns a `rcl_duration_value_t` in the head of the crystal branch , but returns a `rclcpp::Duration` at the head of the dashing branch.  `rclcpp::Duration` does need a `.nanoseconds()` to get the types matching

Jconn: it looks like the type change came in with `c272eb276d` of the diagnostic repo dashing branch and is in the released debian for dashing

src/navsat_transform.cpp
```diff
@@ -110,66 +118,68 @@ void NavSatTransform::run()
...
 
   // Sleep for the parameterized amount of time, to give
   // other nodes time to start up (not always necessary)
-  rclcpp::utilities::sleep_for(std::chrono::seconds(delay));
+  rclcpp::sleep_for(std::chrono::duration_cast<std::chrono::seconds>(std::chrono::duration<double>(delay)));
 
-  rclcpp::rate::Rate rate(frequency);
-  while (rclcpp::ok()) {
-    rclcpp::spin_some(node_);
+  int interval = static_cast<int>(1.0/frequency);
```
Jconn: setting interval as an int here will round it to zero for all frequencies > 1

I think a better type is   
`auto interval = std::chrono::duration<double>(1.0/frequency)
`

and then the wall timer declaration becomes

  `timer_ = this->create_wall_timer(interval, std::bind(&NavSatTransform::transformCallback, this));
`


an interval of 0 nanoseconds here had my wall timer starving out all of the subscriptions such that none of my navsat subscription callbacks were firing. Neat failure signature

klintan: oh nice catch, thank you! will change,

src/ros_filter.cpp
klintan: I'll take an extra looks, thanks for your comments, really appreciate it.

src/ros_filter.cpp
klintan: aah ok, yes I did not pull the absolute latest, I'll do that and re-run. I think the one I had was at least dashing, so maybe that commit was aimed at one of the patch-releases.

include/robot_localization/navsat_transform.hpp
blacksoul000: You don't need to inherit rclcpp::Node, aggregation will be enough.

SteveMacenski: Not if you want to make use of composition or all the powerful tools that are important to ROS2

Jconn:
I am testing this branch out with the dual ekf + navsat with gps and imu in gazebo.

I'm running opensplice as my dds, and I had some trouble with the ekf subscriptions not getting messages. I changed all subscription types to sensor QoS and that resolved that issue. Anyone else hitting this?

Everything "seems" sane with my `odom->base_link` tf, but my `map->odom` link goes crazy if I cross from >0 radians latitude to <0 radians lattiude, 
By crazy I mean the transform jumps from: 
```
    translation:
      x: -0.17590036410552923
      y: 1.5005250803106018
      z: 0.0

```
to

```
    translation:
      x: 429025.9890931018
      y: 9990792.751806233
      z: 0.0

```

~~I set my gazebo model to different lat/long such that I'm not crossing from positive to negative, and now `map->odom` "seems" okay.~~ I've just been teleop'ing around and seeing what my transforms do right now, I'm gonna start testing with the nav stack 


Edit:
Still having trouble with `map->odom` going crazy regardless of my lat/long starting point, I was wrong about it being coupled to sign change in latitude or longitude

src/ros_filter.cpp
klintan: It is set to `false` again after each `do - while`-statement before the next round,  so regardless which statement is hit here, it will be set to false before it is used again (see line 1087).

klintan:
> I am testing this branch out with the dual ekf + navsat with gps and imu in gazebo.
> 
> I'm running opensplice as my dds, and I had some trouble with the ekf subscriptions not getting messages. I changed all subscription types to sensor QoS and that resolved that issue. Anyone else hitting this?
> 
> Everything "seems" sane with my `odom->base_link` tf, but my `map->odom` link goes crazy if I cross from >0 radians latitude to <0 radians lattiude,
> By crazy I mean the transform jumps from:
> 
> ```
>     translation:
>       x: -0.17590036410552923
>       y: 1.5005250803106018
>       z: 0.0
> ```
> 
> to
> 
> ```
>     translation:
>       x: 429025.9890931018
>       y: 9990792.751806233
>       z: 0.0
> ```
> 
> ~I set my gazebo model to different lat/long such that I'm not crossing from positive to negative, and now `map->odom` "seems" okay.~ I've just been teleop'ing around and seeing what my transforms do right now, I'm gonna start testing with the nav stack
> 
> Edit:
> Still having trouble with `map->odom` going crazy regardless of my lat/long starting point, I was wrong about it being coupled to sign change in latitude or longitude

Thanks again for testing out. Hm ok interesting, do you/we know if this is something that is only affecting this ROS2 port of navsat_transform (Just to to get it out of the way, that this is where this behavior originated, and it's not present in ROS1)? 

edit: All other requested changes should be in there.

SteveMacenski:
> I'm running opensplice as my dds, and I had some trouble with the ekf subscriptions not getting messages. I changed all subscription types to sensor QoS and that resolved that issue. Anyone else hitting this?

Good catch. @klintan you need to change all the subscribers for data streams (imu, gps, odom) to use the `SensorDataQoS()` with the `_queue_size` we talked about. QoS is important in ROS2, sorry I should have caught that earlier. 

@jconn do you think these issues are related to this PR, if so any thoughts?

klintan:
> > I'm running opensplice as my dds, and I had some trouble with the ekf subscriptions not getting messages. I changed all subscription types to sensor QoS and that resolved that issue. Anyone else hitting this?
> 
> Good catch. @klintan you need to change all the subscribers for data streams (imu, gps, odom) to use the `SensorDataQoS()` with the `_queue_size` we talked about. QoS is important in ROS2, sorry I should have caught that earlier.

Oh no worries at all, I'll have to take an additional look at the profiles, I thought the queue depth was one of the only differences between the different default profiles included, but I guess there might be other things as well? I'll change, I'll nudge you for a final look after that on that issue.

Jconn:
@klintan  @SteveMacenski  I think my problem was with my config, sorry for the false alarm. I copied the gazebo sensor configs and robot loc configs posted here: https://github.com/nickcharron/waypoint_nav and this package is stable and my transforms look good.  

I've got some minor things to figure out that I think have to do with how I'm using robot loc, not this PR. Once those QoS changes come in I'm comfortable merging

klintan:
> @klintan @SteveMacenski I think my problem was with my config, sorry for the false alarm. I copied the gazebo sensor configs and robot loc configs posted here: https://github.com/nickcharron/waypoint_nav and this package is stable and my transforms look good.
> 
> I've got some minor things to figure out that I think have to do with how I'm using robot loc, not this PR. Once those QoS changes come in I'm comfortable merging

Just to clarify: change them for the `navsat_transform` node subscriptions ? (subscribers for data streams (imu, gps, odom) ) 
and/or in the `ros_filter.cpp` ?

SteveMacenski:
All subscribers to any sensor data

klintan:
> All subscribers to any sensor data

Ok should be added for all. For the custom queue-size and using SensorDataQos wasn't awesome, since  it needed custom initialization etc.  Maybe there is a better way?

Another question: for the custom `queue_size` for navsat_transform, do you want to have a separate queue_size parameter for each sensor type? Right now I just defined one for all so the granularity is lost, but in ROS1 I guess you had neither so.

src/navsat_transform.cpp
```diff
@@ -100,6 +100,7 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options) :
   frequency = this->declare_parameter("frequency", frequency);
   delay = this->declare_parameter("delay", delay);
   transform_timeout = this->declare_parameter("transform_timeout", transform_timeout);
+  queue_size_ = this->declare_parameter("queue_size", 5);
```
SteveMacenski: Ahhhh just when I thought we were done ;-)

Default in ROS1 was `1`, can you make this here too? Also making the queue_size more specific.

... Or just make 1 hardcoded in them and call this done like they were before. If you read the code 1 is the only reasonable number for these anyhow. I'd actually recommend this. No need to add a literally useless parameterization

src/navsat_transform.cpp
```diff
@@ -130,14 +131,14 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options) :
   }
 
   odom_sub_ = this->create_subscription<nav_msgs::msg::Odometry>(
-    "odometry/filtered", 1, std::bind(&NavSatTransform::odomCallback, this, _1));
+    "odometry/filtered", rclcpp::SensorDataQoS(rclcpp::QoSInitialization(RMW_QOS_POLICY_HISTORY_KEEP_LAST, queue_size_)), std::bind(&NavSatTransform::odomCallback, this, _1));
```
SteveMacenski: Cleaner way https://github.com/ros-planning/navigation2/blob/master/nav2_costmap_2d/plugins/obstacle_layer.cpp#L181

```
    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_sensor_data;
    custom_qos_profile.depth = queue_size;
    ...
    create_sub<>(..., SensorDataQoS(custom_qos_profile), ...);
```

You can make 1 profile and just update the depth for each of the subscribers. It's easier on the eyes and sends people in 1 direction to figure out what's going on instead of 5.

SteveMacenski (changes requested):

src/ros_filter.cpp
ayrton04: Right, but my concern is that we'll never break _out_ of this do-while. For example, we set `more_params` to false before the loop here:

https://github.com/cra-ros-pkg/robot_localization/blob/70cd3dd54dc017b0a9167f85ce92f1711df7b0b9/src/ros_filter.cpp#L944

Then, if `odom0` is defined, we set `more_params` to true here:

https://github.com/cra-ros-pkg/robot_localization/blob/70cd3dd54dc017b0a9167f85ce92f1711df7b0b9/src/ros_filter.cpp#L958

Eventually, we reach the loop condition check here:

https://github.com/cra-ros-pkg/robot_localization/blob/70cd3dd54dc017b0a9167f85ce92f1711df7b0b9/src/ros_filter.cpp#L1083

`more_params` is true, so we go back to the beginning of the loop.

We don't have `odom1` defined, so we'll hit this block:

https://github.com/cra-ros-pkg/robot_localization/blob/70cd3dd54dc017b0a9167f85ce92f1711df7b0b9/src/ros_filter.cpp#L961

...but `more_params` is still true. So we'll reach the loop condition check at the end and keep going. I am not currently seeing how this doesn't end in an infinite loop, but I'm assuming people are actually running this code, so I'm clearly missing something.

In the original ROS1 code, we are updating `moreParams` at every loop cycle:

https://github.com/cra-ros-pkg/robot_localization/blob/3bd285e6fc400447086490e833a6f0181284b4d8/src/ros_filter.cpp#L1034

In the new code, we set it to _true_ if it exists, but don't set it to _false_ if it doesn't.

Jconn: @ayrton04 is right, this `more_params` is missing a clear, I needed to add this to get things working

src/navsat_transform.cpp
klintan: I'll just hard-code 1 for those.

src/navsat_transform.cpp
klintan: Thanks, much nicer. Actually did it this way somewhere else, but for some reason I did it this way instead. 

Hm ok, take a look at how I did, it's a bit different the navsat_transform node and the ros_filter.cpp. Sorry I don't think I fully got what you meant :/ 

edit: seems `rclcpp::QoS` and similar only takes `QoSInitialization` 
```
SensorDataQoS::SensorDataQoS(const QoSInitialization & qos_initialization)
```
will see if there are some other signatures to use.
An alternative is: 
`rclcpp::SensorDataQoS(QoSInitialization::from_rmw(custom_qos_profile))`
but not sure how much better it is. 

In navsat_transform my guess was that I could use the same profile for each, and the queue_size is the same for all. 

For ros_filter, I could do it the way you described, just perhaps create one profile early and before each subscription set the `depth` with `queue_size`  I guess it will not reallocate, which saves memory and speeds it up a tiny bump, on the other hand, I was trying to get google the programming principle of keeping things together that is used together haha but I couldn't find it. My concern was that putting the profile creation early early and then just reassigning the depth, that it might be very far from where it's used. But maybe I'm just overthinking this and that it will be a smoother experience (if that is what you meant).

Sorry for going back and forth on this one, but since we've kept at it this long, we might as well get it right, and I'm learning a lot :D at least I think so :P

src/ros_filter.cpp
klintan: Gah 🤦‍♂, ok thanks for taking another look, your def right.

src/navsat_transform.cpp
SteveMacenski: Thanks!

src/navsat_transform.cpp
SteveMacenski: Your suggestion to use `from_rmw` seems fine. I want to drop the `RMW_QOS_POLICY_HISTORY_KEEP_LAST` because 

1) I'm not sure what effects that has
2) It might actually change the sensor data QoS too much to not connect correctly

https://github.com/ros-planning/navigation2/blob/master/nav2_costmap_2d/plugins/voxel_layer.cpp#L84

Another navigation example where we use a custom QoS for a publisher.

Jconn:
The issue I mentioned with my `map->odom` transform was this issue: https://github.com/cra-ros-pkg/robot_localization/issues/222

So I haven't seen any regressions using this PR

src/navsat_transform.cpp
klintan: Ok thanks, I was also worried about 2. or at least wondering what in the profile that actually made it work. Ok i'll use from_rmw. 

I just moved back from NYC to SF, so haven't been able to get to this, but will fix it asap so that we can get this merged :)

klintan: OK this is now fixed, lemme know what you think. I went for this https://github.com/ros-planning/navigation2/blob/master/nav2_costmap_2d/plugins/voxel_layer.cpp#L84 approach. It's pretty explicit and short. Only downside is it's not clear that it's the sensor profile (with depth 1 for some though).
@Jconn could you confirm that these are similar to what you already did and worked?

src/navsat_transform.cpp
```diff
@@ -129,15 +129,17 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options) :
     datumCallback(request, response);
   }
 
+  auto custom_qos = rclcpp::QoS(rclcpp::KeepLast(1)).durability_volatile().best_effort();
```
SteveMacenski: Please use the sensor data QoS, don't specify all of this. The sensor data QoS is just given to you for free

SteveMacenski (changes requested):

src/navsat_transform.cpp
klintan: yeah I have no excuses, fixed now :)

SteveMacenski (approved):

SteveMacenski:
@klintan @Jconn any other things you want to do here before merging?

Jconn:
I needed just a few changes to head of @klintan's branch in ros_filter.cpp for param getters. I didn't catch these earlier with all of my other local changes. I made a [PR](https://github.com/klintan/robot_localization/pull/1/files) to save you the trouble @klintan 
```
     rclcpp::Parameter parameter;
     if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(odom_topic_name, parameter)) {
       more_params = true;
-      odom_topic = node_->get_parameter(odom_topic_name, parameter);
+      odom_topic = parameter.as_string();
     } else {
         more_params = false;
         node_->undeclare_parameter(odom_topic_name);
@@ -1205,7 +1205,7 @@ void RosFilter::loadParams()
     rclcpp::Parameter parameter;
     if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(twist_topic_name, parameter)) {
       more_params = true;
-      twist_topic = node_->get_parameter(twist_topic_name, parameter);
+      twist_topic = parameter.as_string();
     } else {
         more_params = false;
         node_->undeclare_parameter(twist_topic_name);
@@ -1281,7 +1281,7 @@ void RosFilter::loadParams()
     rclcpp::Parameter parameter;
     if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(imu_topic_name, parameter)) {
       more_params = true;
-      imu_topic = node_->get_parameter(imu_topic_name, parameter);
+      imu_topic = parameter.as_string();
     } else {
         more_params = false;
         node_->undeclare_parameter(imu_topic_name);
@@ -2215,6 +2215,7 @@ std::vector<bool> RosFilter::loadUpdateConfig(const std::string & topic_name)
   const std::string topc_config_name = topic_name + "_config";
 
   node_->declare_parameter(topc_config_name, update_vector);
+  node_->get_parameter(topc_config_name, update_vector);

```

SteveMacenski:
... what

I just changed the base to Dashing (because that's LTS and Crystal may as well be Hydro by now).

klintan:
> I needed just a few changes to head of @klintan's branch in ros_filter.cpp for param getters. I didn't catch these earlier with all of my other local changes. I made a [PR](https://github.com/klintan/robot_localization/pull/1/files) to save you the trouble @klintan
> 
> ```
>      rclcpp::Parameter parameter;
>      if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(odom_topic_name, parameter)) {
>        more_params = true;
> -      odom_topic = node_->get_parameter(odom_topic_name, parameter);
> +      odom_topic = parameter.as_string();
>      } else {
>          more_params = false;
>          node_->undeclare_parameter(odom_topic_name);
> @@ -1205,7 +1205,7 @@ void RosFilter::loadParams()
>      rclcpp::Parameter parameter;
>      if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(twist_topic_name, parameter)) {
>        more_params = true;
> -      twist_topic = node_->get_parameter(twist_topic_name, parameter);
> +      twist_topic = parameter.as_string();
>      } else {
>          more_params = false;
>          node_->undeclare_parameter(twist_topic_name);
> @@ -1281,7 +1281,7 @@ void RosFilter::loadParams()
>      rclcpp::Parameter parameter;
>      if (rclcpp::PARAMETER_NOT_SET != node_->get_parameter(imu_topic_name, parameter)) {
>        more_params = true;
> -      imu_topic = node_->get_parameter(imu_topic_name, parameter);
> +      imu_topic = parameter.as_string();
>      } else {
>          more_params = false;
>          node_->undeclare_parameter(imu_topic_name);
> @@ -2215,6 +2215,7 @@ std::vector<bool> RosFilter::loadUpdateConfig(const std::string & topic_name)
>    const std::string topc_config_name = topic_name + "_config";
>  
>    node_->declare_parameter(topc_config_name, update_vector);
> +  node_->get_parameter(topc_config_name, update_vector);
> ```

NIce ones! and thanks. Seems that is a typo as well, but it was in the old file, so we might as well have another PR for that to try to get this one merged as it's been dragging on for a while. 
@SteveMacenski thanks for updating, didn't even realize it was made on the crystal branch. Anything else we need now before merging?

edit: whoops, see now that we have a bunch of conflicts so we'll need a rebase.

SteveMacenski:
Resolving merge conflicts. I glanced through about half of them, they're really easy, just alot of them

klintan:
> Resolving merge conflicts. I glanced through about half of them, they're really easy, just alot of them

Thanks for your patience, took a while to knock it out. Fixed a bunch of linting issues that failed in the tests. Let me know if you see anything I missed after/during the rebase.

SteveMacenski:
@jconn are your concerns resolved?

Jconn (approved):

Jconn:
> @Jconn are your concerns resolved?

No more concerns, good work @klintan

SteveMacenski:
I dont believe it. Its finally in!

ayrton04:
Epic. Thanks to everyone involved!

PR #478, opened 2019-08-09
Changes needed for Dashing

rotu:
Declare parameters
Name transform listener node to prevent node name collision
Reuse executor (rclcpp::spin_some creates an executor, which takes 36 ms, causing significant issues).

### DIFF
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -653,6 +653,8 @@ protected:
   //!
   rclcpp::Node::SharedPtr node_;
 
+  rclcpp::QoS qos_;
+
   //! @brief Transform buffer for managing coordinate transforms
   //!
   tf2_ros::Buffer tf_buffer_;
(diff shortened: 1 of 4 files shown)


SteveMacenski:
Does this substantively different than what's in #476 ? I see alot of the same "stuff" in different places. 

Only off hand review question beyond trying to rectify these PRs together is I see you've hardcoded `odom0` and `imu0`, that needs to not be like that. One of the benefits of `r_l` is that it is generalized enough to handle `N` sources

rotu:
@SteveMacenski It's pretty similar, though that PR looks like it may be in a better state.
I'm ambivalent which one should be merged in.

@cottsay did most of the work on this one, so I already feel a little weird creating a PR of someone else's work.

rotu:
yielding to #476. I'll make any needed changes on top of that PR.

SteveMacenski:
@rotu eh, he works for open robotics, I think its ok ;-) they're busy enough.

If this adds something more to that PR says so and we can work out a plan to integrate, just from a quick scan it seemed similar

PR #483, opened 2019-08-25
use correct NodeHandle instances and node names for diagnostic updater

dseifert:
When running a robot_localization filter, the diagnostic output is using the default node handle (an instance of `ros::NodeHandle()`) and the node's name (`ros::this_node::getName()`). However this is not ideal for nodelets, as both refer to the nodelet manager but not the nodelet. Running a single robot_localization nodelet, this can be somewhat ignored, however running two or more will cause issues as it can not be differentiated which nodelet is sending what diagnostic output (or even how many nodelets are sending).

The issue lies in the creation of `diagnostic_updater::Updater`. Here, the proper nodehandle must be given. This is easy to accomplish as this is already available in the `RosFilter` class. However for differentiating the output, the node's name as derived from `ros::this_node::getName()` is used to label the diagnostic update (field `name` in [DiagnosticStatus](http://docs.ros.org/api/diagnostic_msgs/html/msg/DiagnosticStatus.html)) and is ultimately what the output can (must) be differentiated by. Supporting this properly is more difficult, as (a) `RosFilter` is already unaware of whether it is running as node or nodelet and (b) it is not possible to derive the node(let)'s name from a given `NodeHandle` instance.

The solution presented here is expanding the constructor to also pass in the node's name (either (default value) `ros::this_node::getName()` or `Nodelet::getName()`) to `RosFilter`. To avoid API break the constructor is split into two. However this will / may break ABI?

Before applying this patch, running e.g. two ekf_localization nodelets would result in the nodelet manager publishing the diagnostic output on /diagnostic with apparently twice the expected rate (as there are two nodelets) with no possibility to differentiate which particular diagnostic message belongs to which nodelet. With this patch, this will be easy to differentiate as the nodelet's name is used as the sender of the message.

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -111,6 +111,16 @@ template<class T> class RosFilter
     //! The RosFilter constructor makes sure that anyone using
     //! this template is doing so with the correct object type
     //!
+    explicit RosFilter(ros::NodeHandle nh,
+                       ros::NodeHandle nh_priv,
+                       std::string node_name,
+                       std::vector<double> args = std::vector<double>());
+
+    //! @brief Constructor
+    //!
+    //! The RosFilter constructor makes sure that anyone using
+    //! this template is doing so with the correct object type
+    //!
     explicit RosFilter(ros::NodeHandle nh, ros::NodeHandle nh_priv, std::vector<double> args = std::vector<double>());
 
     //! @brief Destructor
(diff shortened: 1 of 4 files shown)


ayrton04:
Just a note that I haven't had a chance to look this over yet. Thanks for submitting it!

ayrton04 (approved):
Thanks for submitting this. I'm actually on the fence about worrying about breaking ABI, as I don't know who would be linking against the `RosFilter` class. But that's not a good reason for breaking it. :) And I *think* adding a new non-virtual method should be fine from an ABI perspective.

PR #487, opened 2019-09-04
Fix pub sub deprecation warnings and diagnostics errors for Dashing

mabelzhang:
Syncs to latest Dashing, to build without warnings and errors.

[Diagnostics](https://github.com/ros/diagnostics/tree/dashing) was released yesterday for Dashing. The change in the [Updater::Updater() constructor](https://github.com/ros/diagnostics/pull/112/files#diff-20c9ee76647764cf9f41b2260853ebb1L374) caused an error in `ros_filter.cpp` about implicitly using a no-arg constructor. Another change is that `diagnostic_updater.getPeriod()` now returns `rclcpp::Duration`, which caused the other error in the same file.

All the rest of the changes are fixing the deprecation warnings about `create_subscription()` and `create_publisher()`, which now require a `rclcpp::QoS`. These are in `ros_filter.cpp` and 3 files in `test/`.


src/ros_filter.cpp
```diff
@@ -1018,6 +1026,7 @@ void RosFilter::loadParams()
 
         topic_subs_.push_back(
           node_->create_subscription<nav_msgs::msg::Odometry>(odom_topic,
+          rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_default)),
```
SteveMacenski: I believea ll of these subscribers have a parameter for queue size that QoS should use

mabelzhang: You were right! Corrected in 0423eaf and 73611be. Thanks!

SteveMacenski: I meant that `rclcpp::SystemDefaultsQoS()` doesnt specify the queue size, but the parameters for r_l will set them `odomN_queue_size: 2`, so this will ignore the parameterizations for all the subs of data

mabelzhang: Okay. I see what you mean. I looked at the [melodic-devel version](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ros_filter.cpp#L1071) and tried to mirror the reading of `odom`, `pose`, `twist`, and `imu` `_queue_size`s from parameters. `SystemDefaultsQoS()` is eliminated in daba06a.

SteveMacenski:
The only thing that looks odd to me is the lack of `declare_parameter` calls, I thought in Dashing you had to declare your parameters in order to get them? But I'll be frank, my knowledge on the parameter situation is perhaps not the best

@ayrton04 I see nothing else worth blocking over

src/ros_filter.cpp
SteveMacenski: Awesome thanks for that

mabelzhang:
You’re right! I’m not very familiar with changes in Dashing. `declare_parameter()` added to all parameters in eec99d2 and simplified in 8b76030.

Before:
```
$ ros2 param list
/se_node:
  diagnostic_updater.period
  use_sim_time
/transform_listener_impl:
  use_sim_time
```

After declarations:
```
$ ros2 param list
/se_node:
  base_link_frame
  control_timeout
  debug
  diagnostic_updater.period
  dynamic_process_noise_covariance
  filter_type
  frequency
  gravitational_acceleration
  history_length
  imu0
  initial_estimate_covariance
  initial_state
  map_frame
  odom0
  odom_frame
  pose0
  print_diagnostics
  process_noise_covariance
  publish_acceleration
  publish_tf
  reset_on_time_jump
  sensor_timeout
  smooth_lagged_data
  tf_prefix
  transform_time_offset
  transform_timeout
  twist0
  two_d_mode
  use_control
  use_sim_time
  world_frame
/transform_listener_impl:
  use_sim_time
```

SteveMacenski:
Much better.

SteveMacenski (approved):

cr01927:
RE: Declaring parameters, to load parameters from the YAML file, I had to create and assign a NodeOptions with `automatically_declare_parameters_from_overrides` set to `true`, and then wrap the `declare_parameter` calls inside a check to `has_parameter`. I'm not sure if I missed a much simpler way to load the YAML file.

ETA or just a simpler way to configure the "dynamic" parameters (odom0, imu0, etc.)

SteveMacenski:
@mabelzhang have you checked this passes tests and works (able to set all the parameters to 42 or something ;-) )?

@ayrton04 I think this is a good PR to get ROS2 limbing again. After October I'll plan to sit down and deal with some of the ROS2 issues we discussed but it would be great to get it moved forward a little bit

ayrton04:
Sorry, all. Been a little busy. Thanks very much for the PR!

PR #490, opened 2019-09-19
remove a couple unnecessary lines that clang was complaining about

nickolai-voyage:
These were highlighted by clang static analyzer. The exact message for each case was "Value stored to 'variable_name' is never read."


ayrton04:
Yeah, these look safe. Thanks.

PR #495, opened 2019-10-01
Remove unused variable utm_odom_tf_yaw_ for clang compatibility

timonegk:
Building `robot_localization` with CLang currently fails because the variable `utm_odom_tf_yaw_` in `navsat_transform` is unused. CLang seems to be more restrict than GCC here, and due to the `-Werror` flag, the code does not compile. This PR removes the unused variable.

CLang output on branch `melodic_devel`:
```
In file included from /tmp/robot_localization/src/navsat_transform.cpp:33:
.../robot_localization/include/robot_localization/navsat_transform.h:218:12: error: private field 'utm_odom_tf_yaw_' is not used [-Werror,-Wunused-private-field]
    double utm_odom_tf_yaw_;
           ^
1 error generated.
```

### DIFF
--- a/include/robot_localization/navsat_transform.h
+++ b/include/robot_localization/navsat_transform.h
@@ -213,10 +213,6 @@ class NavSatTransform
     //!
     double utm_meridian_convergence_;
 
-    //! @brief Stores the yaw we need to compute the transform
-    //!
-    double utm_odom_tf_yaw_;
-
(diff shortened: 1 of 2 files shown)


twdragon:
I am ready to support this pull request. The new compilers are stopping build on this field.

ayrton04:
Sorry for the delay, all.

PR #497, opened 2019-10-19
Fix IMU update vector

### LINKED ISSUE #432: [EKF] Parameter  `imuN` contains and uses fields for x,y,z, vx,vy,vz
The fields for `x,y,z, vx,vy,vz` in the parameter `imuN` first of all make no sense, because an Imu message does not contain these; even worse though, they are actually used by the filter.
Hence, if somebody was to set the field of `vx` to `true` the filter will accept this and introduce a velocity measurement of value 0.0 with 0.0 covariance, which is highly undesired.

The parameters should either be removed or at least it should be ensured, that `true` entries of non-existing data does not have an impact.

johuber:
Closes #432.

Previously if the config entries for position or linear_speed update from IMU topic were set to true:
*  measurements with values 0.0 were introduced into state estimation

Since sensor_msgs/Imu does not contain any infos about these fields they should never be read. Now they are masked and ignored, additionally there is a `ROS_WARN` output if they are set true, to alert the user that this does not have any effect.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1367,6 +1367,26 @@ namespace RobotLocalization
         // update configuration (as this contains pose information)
         std::vector<int> updateVec = loadUpdateConfig(imuTopicName);
 
+        // sanity checks for update config settings
+        std::vector<int> positionUpdateVec(updateVec.begin() + POSITION_OFFSET,
+                                           updateVec.begin() + POSITION_OFFSET + POSITION_SIZE);
+        int positionUpdateSum = std::accumulate(positionUpdateVec.begin(), positionUpdateVec.end(), 0);
+        if (positionUpdateSum > 0)
+        {
+          ROS_WARN_STREAM("Warning: Some position entries in parameter " << imuTopicName << "_config are listed true, "
+                          "but sensor_msgs/Imu contains no information about position");
+        }
+        std::vector<int> linearVelocityUpdateVec(updateVec.begin() + POSITION_V_OFFSET,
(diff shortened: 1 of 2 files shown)


ayrton04:
Thanks for fixing this! I'll take a look later today or tomorrow.

src/ros_filter.cpp
```diff
@@ -1367,7 +1367,31 @@ namespace RobotLocalization
...
         std::vector<int> updateVec = loadUpdateConfig(imuTopicName);
 
+        // sanity checks for update config settings
+        std::vector<int> positionUpdateVec(updateVec.begin() + POSITION_OFFSET,
+                                           updateVec.begin() + POSITION_OFFSET + POSITION_SIZE);
+        int positionUpdateSum = std::accumulate(positionUpdateVec.begin(), positionUpdateVec.end(), 0);
+        if (positionUpdateSum > 0)
+        {
+          ROS_WARN_STREAM("Warning: Some position entries in parameter " << imuTopicName << "_config are listed true, "
+                          "but sensor_msgs/Imu e contains no information about position");
```
ayrton04: Extraneous 'e' here.

ayrton04 (approved):
Thanks for this! Please fix the warning I mentioned, but the rest LGTM.

src/ros_filter.cpp
johuber: fixed

ayrton04 (approved):

ayrton04:
Thanks again. Will merge once the build farm gets done.

PR #505, opened 2019-11-20
Use ament_target_dependencies correctly

jacobperron:
This change fixes a CMake error when building against the latest version of ROS 2.
The command replaces the use of target_link_libraries for ament packages.
I also removed the empty CMake variable 'dependencies' and added a missing dependency on rcl.


SteveMacenski (approved):

ayrton04 (approved):

ayrton04:
And this is still valid for Dashing? Or does it need to go into Eloquent?

SteveMacenski:
Valid in Dashing

jacobperron:
Yes, this change is valid for Dashing and Eloquent.

PR #509, opened 2019-12-04
Dashing cleanup

SteveMacenski:
This is a little bit of a monster PR. I would have separated it but this is something like ripping off a band aid and it could have been 10 PRs outright and that would take forever.

What happened:
- Separating out UKF and EKF nodes
- Separating classes into individual files for headers (measurement, state)
- Changed the `ImuAccBasicIO` test ref value from 1.2 -> 1.8 (see comment below)
- Retemplated RosFilter
- pointerized all ROS assets, added global assets as required
- Removed Run() method for the timer based solution
- Fixed typos
- Node-ified the Rosfilter base. Its possible to register as a component now but I did not do that.
- Exposed the filter to the ros filter so we can set UKF constants (or other filter constants) without the vector of doubles hack
- Changed namespaces for parameter yaml files. I noticed there's a dual version but no launch file for that... so if it didn't work before, it doesn't work now still. If it worked before, then it works now as well. 
- Launch files: only actual launch files. And ones that actually use the parameters. Because, that seems obvious. 
- Fixed diagnostics test 
- fixed interface tests
- Enable all the tests to run through `colcon build`

What didnt happen
- CI
- Bag tests still in an unusable state

"What didnt happen" is my future work. More than happy to let someone else do the tests stuff however ;-)

This has **not** been tested at all. @klintan if you're an RL user, I'd appreciate a few moments as a checkbox that "this works" I have no physical assets to test on. Nothing functionally changed so if it doesnt crash then we're good.

@mabelzhang FYI (also if you have a robot in ROS2 to test with)
@rotu this would be the first official release of this package as debians, so I'd appreciate if you took a look too with Rover since it'll affect you downstream. I went through your forks and added the relevant items


mabelzhang:
At the moment we're trying to make our stuff more stable, so we won't be testing this immediately.

Just curious, why are the tests removed in CMakeLists, do they not pass? I'm doing some ports of the diffs between melodic-devel and dashing-devel, and it would be helpful to know if I shouldn't be running those tests in ROS 2 as an indication of whether I ported things correctly.

SteveMacenski:
> At the moment we're trying to make our stuff more stable, so we won't be testing this immediately.

My goal here too

> Just curious, why are the tests removed in CMakeLists, do they not pass?

That testing framework with shell scripts is really unideal. The CI for RL in ROS1 was way better and I'm not sure what's even going on here. Rather than trying to fix it, I'd rather just start over from ROS1. Why did I delete them from Cmake? Frankly because I did that to compile some code and forgot to add it back :laughing: but then just owned it since I am planning on fixing CI later.

SteveMacenski:
@mabelzhang added the tests back in to the CMakeLists & updated

SteveMacenski:
For later reference this commit: https://github.com/cra-ros-pkg/robot_localization/pull/509/commits/b31b9bb34795be8353d707295ff6336959edf37e

This is (minus the places I had to fix things that were clearly wrong) a goodish example of how to make proper tests in ROS2 using the test fixtures and launch system without shell scripts. 

I'm not sure I'll be updating _all_ the tests, but at least a few of them so its a bit cleaner

(sorry, I'm also getting more done on this today than I expected so I keep creeping more and more in this PR...)

mabelzhang:
Cool. I wouldn't complain about that :D Thank you!

SteveMacenski:
Status update:
- pretty sure it works, still need some verification
- The base filter tests pass and EKF/UKF interface tests (I'd argue the best smoke tests for the fact that its probably working as stated above) are working except the IMU tests which are reporting PI with it should be reporting 0... that stinks of something fishy but working through it. I suspect given what I know about this code that its either a) broken before I got here b) the test is broken but the code is OK c) I really somehow messed up the IMU math inside a refactor, which would be impressive

SteveMacenski:
Here's a good question @ayrton04 isn't [this](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/test/test_ekf_localization_node_interfaces.cpp#L636) test a singularity? That would explain the test failure that it should be 0 but is reporting PI. That's the only test failing after fixing some ros2-y things in the interface tests for EKF/UKF.

ayrton04:
Hrm, it ought to be, yeah, but I'm curious why it used to work. I once started to go through the process of switching everything to quaternions (_that_ was a fun Jacobian), but it never went anywhere.

In any case, feel free to give the angular y velocity a different value that won't cause it to pass through -pi/2.

SteveMacenski:
Immediately on changing PI/2 to PI/3 we're off to the races. I stared at that for a solid 5 minutes straight before I realized it was the classic singularity condition. It may have worked in ROS1 since its ill-defined, some library change in TF or below TF resulted in changes in the singularity outcome. 

(I'm surprised RL doesnt already use quaternions, I'm surprised you couldn't just steal that math from another less-general E/UKF github project. Remind me to not use RL if I start a highspeed drone company)

SteveMacenski:
@mabelzhang I fixed up all the tests except the bag tests. For that, I spent a *bunch* of time just getting the code to work but I suspect the bag isn't being played. I at least fixed all the parameter issues.

I'm going to un-scope fixing the bag tests and comment them out from the CMakeLists so they dont run on `colcon test`. If you have cycles / interest for that, please do!

Otherwise, all tests are now working (except 1, working on it now). I'm confident this works as expected from a CI level. Still needs hardware validation. 

`colcon test` doesn't work because of the launch files I only migrated the diagnostics launch file to the proper format. But all the non-commented-out tests succeed in launching them `ros2 launch ...`. Migrating the rest will be part of setting up CI. I'm pretty burned out of automated testing for the week... maybe I'll pick that back up in the new year.

mabelzhang:
Awesome!! Thank you for doing all this!

SteveMacenski:
@mabelzhang @ayrton04 

Alright, passes tests. There's a couple of asterisks I wanted to bring up explicitly.

I had to add a 10% margin to some tests because they were reallly close but not quite on the ball. This is due to the creation of a node in each test so it takes a bit for the network traffic to make connections and I'll typically see about 10% of messages dropped. I think that would account for this change required.

Another thing is the `ImuAccBasicIO` test **which may indicate an algorithmic issue in the filters or I'm not sure how it passed in ROS1**. What I'm seeing is that the actual value of 
```

  EXPECT_LT(::fabs(filtered_.pose.pose.position.x - 1.2), 0.4);
  EXPECT_LT(::fabs(filtered_.pose.pose.position.y + 1.2), 0.4);
  EXPECT_LT(::fabs(filtered_.pose.pose.position.z - 1.2), 0.4);
```

required to be changed to 

```

  EXPECT_LT(::fabs(filtered_.pose.pose.position.x - 1.8), 0.4);
  EXPECT_LT(::fabs(filtered_.pose.pose.position.y + 1.8), 0.4);
  EXPECT_LT(::fabs(filtered_.pose.pose.position.z - 1.8), 0.4);
```

to pass. I consistently saw this and even after pulling out a bunch of ROS2 traffic metrics to see if this was a case where also some margin was required, but no. Even when I changed the QoS profiles to promise delivery and give adequate time to process, I'm seeing it hover around 1.8. See math below, but overall I'm seeing both the filters seemingly "lead" the signal which I wouldn't normally expect.

Assuming prefect kinematics, v_f = v_i + a * t. For the first part of this test we apply a 1 m/s2 acceleration over 1 second from stop. That would give us a 1m/s v_f. In reality I see something near 1m/s on occasion but its usually around 1.14m/s. And that's with some dropped messages, which frankly baffles me with constant acceleration being published faster than filter update rate. Either way, the test for the speed:

``` 
  EXPECT_LT(::fabs(filtered_.twist.twist.linear.x - 1.0), 0.41);
  EXPECT_LT(::fabs(filtered_.twist.twist.linear.y + 1.0), 0.41);
  EXPECT_LT(::fabs(filtered_.twist.twist.linear.z - 1.0), 0.41);
```

Passes without an issue, 1, 1.14, all good. Now the next step is where we see the problem. Going back to our kinematics: dx = v_o * t + 0.5 * a * t * t. If we have our initial velocity at 1 m/s and 0 m/s2 acceleration, we should move forward 1m. Knowing dx = (dv/2) * t, we know that the first step's change in pose was 0.5m. So that's 1.5m, in theory. So... why was the test for `1.2m`? Because a filter lags the signal. Makes sense.

So what we're actually seeing, with those numbers, is 1.14m/s which gives us a displacement in the 0m/s2 acceleration step of 1.14m. Knowing the start and end velocity of the acceleration phase, we get 0.57 as our displacement, totalling to 1.71m displacement. Which is _less_ than what I'm seeing at 1.8m but its in the `0.4` error tolerance the test started within.

So that indicates to me that its leading the signal (?) which is odd. So overall for this particular test, I don't really understand where `1.2` came from, I think it should be `1.5` in a perfect world and the `0.4` tolerance band would still cover `1.2` from `1.5` as the setpoint if we're accurately tracking our signal. I don't know why we're leading the signal, but again if we had set `1.5` as our setpoint, `1.8` would fall in the tolerance band.

Given that I see this identically in both the EKF and the UKF this gives me a big "phew" feeling that I don't have to look at transfer functions, **but if this behavior is different than what you see in ROS1, this may be indicative of an issue**. However its consistent. 

```
s.macenski@reese:~/Documents/dashing_ws$ ros2 launch robot_localization test_ekf_localization_node_interfaces.launch.py
[INFO] [launch]: All log files can be found below /home/s.macenski/.ros/log/2019-12-06-16-21-22-150139-reese-5466
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [ekf_node-1]: process started with pid [5483]
[INFO] [test_ekf_localization_node_interfaces-2]: process started with pid [5484]
[ekf_node-1] [WARN] [rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
[test_ekf_localization_node_interfaces-2] [==========] Running 10 tests from 1 test case.
[test_ekf_localization_node_interfaces-2] [----------] Global test environment set-up.
[test_ekf_localization_node_interfaces-2] [----------] 10 tests from InterfacesTest
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.OdomPoseBasicIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.OdomPoseBasicIO (1536 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.OdomTwistBasicIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.OdomTwistBasicIO (82595 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.PoseBasicIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.PoseBasicIO (1605 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.TwistBasicIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.TwistBasicIO (82594 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.ImuPoseBasicIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.ImuPoseBasicIO (3106 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.ImuTwistBasicIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.ImuTwistBasicIO (6010 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.ImuAccBasicIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.ImuAccBasicIO (4012 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.OdomDifferentialIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.OdomDifferentialIO (6525 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.PoseDifferentialIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.PoseDifferentialIO (6604 ms)
[test_ekf_localization_node_interfaces-2] [ RUN      ] InterfacesTest.ImuDifferentialIO
[test_ekf_localization_node_interfaces-2] [       OK ] InterfacesTest.ImuDifferentialIO (12612 ms)
[test_ekf_localization_node_interfaces-2] [----------] 10 tests from InterfacesTest (207200 ms total)
[test_ekf_localization_node_interfaces-2] 
[test_ekf_localization_node_interfaces-2] [----------] Global test environment tear-down
[test_ekf_localization_node_interfaces-2] [==========] 10 tests from 1 test case ran. (207200 ms total)
[test_ekf_localization_node_interfaces-2] [  PASSED  ] 10 tests.
```

Ah. This is what I've been waiting a week to see.

include/robot_localization/ros_filter.hpp
```diff
@@ -93,17 +94,16 @@ using MeasurementQueue =
 using MeasurementHistoryDeque = std::deque<MeasurementPtr>;
 using FilterStateHistoryDeque = std::deque<FilterStatePtr>;
 
-class RosFilter
+template<class T>
```
ayrton04, line 97: I think I was the one who actually removed the templating, as it seemed that having a pointer to the core filter made more sense than templating. But I'm not married to it. I was also aiming for having a single state estimation node, with the algorithm being configurable. But again, this is fine.

include/robot_localization/ros_filter.hpp
```diff
@@ -299,7 +310,7 @@ class RosFilter
   //! older measurements come in.
   //! @param[in] filter - The filter base object whose state we want to save
   //!
-  void saveFilterState(FilterBase::UniquePtr & filter);
+  void saveFilterState(T & filter);
```
ayrton04, line 313: I need to revisit the definition, but is there any reason we can't make this parameter `const`?

include/robot_localization/ros_filter_utilities.hpp
```diff
@@ -82,7 +82,7 @@ double getYaw(const tf2::Quaternion quat);
 //! value of @p targetFrameTrans and returns true, otherwise it returns false.
 //!
 bool lookupTransformSafe(
-  const tf2_ros::Buffer & buffer,
+  const tf2_ros::Buffer * buffer,
```
ayrton04, line 85: Should we just update these methods to take unique_ptrs directly? Or even just const references?

src/ekf.cpp
```diff
@@ -304,78 +304,78 @@ void Ekf::predict(
   // Prepare the transfer function Jacobian. This function is analytically
   // derived from the transfer function.
   double x_coeff = 0.0;
-  double yCoeff = 0.0;
+  double y_coeff = 0.0;
```
ayrton04, line 307: D'oh!

ayrton04:
This is really great, thanks. I'm slowly working my way through it. In the meantime...

> I had to add a 10% margin to some tests because they were reallly close but not quite on the ball. This is due to the creation of a node in each test so it takes a bit for the network traffic to make connections and I'll typically see about 10% of messages dropped. I think that would account for this change required.

Does ROS2 have the ability to wait until all the connections have been established before commencing playback?

The IMU test changes are a little suspicious. Perhaps we'll need to enable debug mode, comment out all the other tests, and run it, and then repeat with the ROS1 versions, just to see what shakes out.

The ridiculous tolerances don't say much for the quality of the tests to begin with. It might make more sense to have a test fixture that inherits from the `RosFilter` and lets us manually fire the callbacks. That would make everything a little more deterministic. But that's outside scope here. I just want to make sure that nothing was broken between ROS1 and ROS2.

Reading over this old code is humbling. There are a lot of things I'd do (very) differently now. :)

SteveMacenski:
> Does ROS2 have the ability to wait until all the connections have been established before commencing playback?

I could do something like that, I did a rough test adding sleeps after the node was created to let everything settle down, but it ended up messing up the filter more. That's certainly a "future work" thing to have a single node given to tests vs creating their own, and creating subscribers / publishers / service clients on the fly constantly, that's got to take a toll. 

> The IMU test changes are a little suspicious. Perhaps we'll need to enable debug mode, comment out all the other tests, and run it, and then repeat with the ROS1 versions, just to see what shakes out.

I'm not, not suspicious of it either, one reason I highlight it. 

> But that's outside scope here.

Agreed. They pass, they pass, but I also like keeping the server separated for my personal testing style. I like to poke the bear as the bear is seen in the wild. 

On the test side of this PR: I'd say there's certainly lots of room for improvement, but the goal here was really to clean up RL functional code, and the changes I made to the tests were singularly with the goal of showing that the functional code changes I made didnt break anything. Test refactoring is out of scope for my attempt of this PR. Though the IMU accelerometer test is showing some confusing results that probably need more investigation. However I'm pretty confident in saying nothing in my refactor caused that so its a feature of the ROS2 port to begin with.

SteveMacenski:
Colcon test now works (except for bags, I know I keep saying I'm not going to do something, then I do it,  but I'm _really_ not going to do the bag file ones, also commented out so it doesnt run default). CI will wait until we're ready to release so we can make use of the ROS PR builder

On my venture debugging I think I can see maybe how the 1.2 -> 1.8 fixes things. See [exhibit A](https://github.com/cra-ros-pkg/robot_localization/blob/dashing-devel/src/ros_filter.cpp#L1967) and [exhibit B](https://github.com/cra-ros-pkg/robot_localization/blob/dashing-devel/src/ros_filter.cpp#L199) that may potentially be causing a problem. There are 3 places (2 things) that are removed from the code. I don't think they cause this problem since we still query `last_set_pose_time_` param on the callback trigger and throw out anything stamped as old, though obviously clearing the queue would be more ideal. 

I do notice however that when I run the test by itself, it passes with `1.2`. Maybe something is wrong with the resetting, acceleration, or IMU callbacks? If I move it to the first test, it passes but then the second test fails. If I mess with some params there, I can get all the tests to pass. That leads me to believe that the state on `set_pose` isn't being fully reset. 

Also after getting everything running and playing nice, I was able to remove all the 10% margins. So that's no longer something to think about :fireworks: 

This is definitely something that should be looked at. I'm not sure I'd block this PR for that however -- I just tested against current `dashing-devel` which fails every test so its definitely an improvement. I opened #511 on this IMUAccel test that we had to change to 1.8m when running after other tests.

SteveMacenski:
@klintan @mabelzhang any thoughts on this?

klintan:
Haven't gotten to this, I'll take a look tonight :)

JWhitleyWork:
@SteveMacenski - Are there items still to complete for a "complete" Dashing port after this MR? I'd be happy to help with them (yes, including linting).

SteveMacenski:
2 things I know are still missing:
- CI for ROS2
- tests using the bag files are very broken

Mabel mentioned on the ROS2 port ticket she’s been delayed in cherry picking all the recent commits in RL from ROS1 that have happened over the last couple years so there’s some there too.

JWhitleyWork:
Cool, thanks. I can try adding the CI for it. It looks like there isn't any for ROS1 either, is that the case? If so, I would be willing to set it up with CircleCI, but I would need to be added to the managing group on Github.

SteveMacenski:
Mhm, that does appear to be the case. 

https://circleci.com/add-projects/gh/cra-ros-pkg Looks like I don’t have admin to enable it. I was also planning on setting up the ROS build farm PR builder for CI when I released dashing and eloquent on their devel branches. 

We should continue this in the ros2 port ticket

include/robot_localization/filter_state.hpp
```diff
@@ -0,0 +1,87 @@
...
+  // The time stamp of the most recent control term
+  rclcpp::Time latest_control_time_;
+
+  // We want the queue to be sorted from latest to earliest timestamps.
+  bool operator()(const FilterState & a, const FilterState & b)
+  {
+    return a.last_measurement_time_ < b.last_measurement_time_;
+  }
+};
+using FilterStatePtr = std::shared_ptr<FilterState>;
```
klintan, line 83: A question/comment here since I don't know the standard:  since this is a shared ptr does it make more sense to be called `FilterStateSharedPtr` for instance the shortcuts in ROS2 for messages are always <>::SharedPtr, and Ptr. This name would imply `unique_ptr`.  This is extreme nitpicking though, so feel free to ignore this and I think the verbosity of the longer name might also be undesired.

src/ekf_node.cpp
```diff
@@ -45,31 +42,12 @@
...
-
-    filter = std::make_unique<robot_localization::Ekf>();
-  }
-
-  robot_localization::RosFilter ros_filter(node, filter);
-  ros_filter.run();
-
+  rclcpp::NodeOptions options;
+  options.arguments({"ekf_filter_node"});
+  std::shared_ptr<robot_localization::RosEkf> filter =
```
klintan, line 47: Could this be an `auto` without loosing too much readability? Maybe I'm missing something on why you would like it to be an explicit type here? (again this is nitpicking so feel free to ignore)

klintan:
Went through everything, looks awesome! I like the new changes and making the ros filter templated! I didn't see anything, had some very minor comments/questions.  

My IMU is unfortunately broken :/ so I don't have the ability to test this until I get a new one.

src/ekf_node.cpp
SteveMacenski: I hate auto, and its really not used in the styling of RL. I know there is some in there now, but I don't think that's something to be aimed for.

include/robot_localization/filter_state.hpp
SteveMacenski: There's 2-3 places where the Ptr is used to describe a shared ptr. I dont think what you describe would hurt though. 

I never read a `Ptr` as a unique ptr, I dont think by any standard I'm aware of you can assume the type of pointer a `Ptr` type is. I'd actually assume it would a raw pointer unless otherwise mentioned. Smart pointers are still relatively new to C++

ayrton04:
> I like the new changes and making the ros filter templated!

This was actually the way it was originally written, but for one reason or another I convinced myself that it wasn't the right approach, so I changed it for ROS2. Happy for the templates to return, though. I'm not that fussed either way.

Is there anything else we want to sort out before we merge this?

SteveMacenski:
> I convinced myself that it wasn't the right approach

I think its as good as anything. I think the power of templates is better displayed when there are many, many implementations and needs. Something with 20 implementations the "if/else" on a parameter string gets exhaustive. Pluginlib is good for something like this as well but frankly it didn't come to mind at the time. Templates make it easier to have a bunch of code run and make different executable for each but reduce code redundancy to near-zero. I didn't know it was intentional to remove the templates and create 1 executable for all the filters. I probably wouldn't have done it if I knew that's what you were going for. I thought of R_L in terms of `ukf` node and `ukf` node, but I suppose I could also get behind a single node with a parameter on each. I haven't gone through the forks recently, has anyone implemented another filter that could be added upstream?

I don't think there's anything I think I need to do before merging. I was hoping for someone independently to give it a whirl and report back, but I think its ready to go.

JWhitleyWork (approved):
LGTM. Did not test on-vehicle but the code looks sound.

ayrton04 (approved):

ayrton04:
Thanks for all the effort!

include/robot_localization/ros_filter_utilities.hpp
SteveMacenski: Passing smart pointers to functions is considered against the best practices for C++. Not to say I didn’t do that because of that reason, I just did what required the least amount of effort :-) 

With that said, the smart pointers best practices are still developing.

include/robot_localization/ros_filter.hpp
SteveMacenski: It can be, yes

PR #514, opened 2019-12-12
Set the filtered gps frame_id to base_link

Timple:
This is actually the frame_id of the information, not the 'gps' frame which might not even exist.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -690,7 +690,7 @@ namespace RobotLocalization
 
       filtered_gps.position_covariance_type = sensor_msgs::NavSatFix::COVARIANCE_TYPE_KNOWN;
       filtered_gps.status.status = sensor_msgs::NavSatStatus::STATUS_GBAS_FIX;
-      filtered_gps.header.frame_id = "gps";
+      filtered_gps.header.frame_id = base_link_frame_id_;
       filtered_gps.header.stamp = odom_update_time_;
 
       // Mark this GPS as used


ayrton04:
Thanks for the PR. However, I'm not yet convinced this is the right move. The frame of the GPS data is actually the geographic coordinate system.

If we look at the Kalman filters, they are producing a state estimate in, e.g., the *odom* frame. In the output message, you have a header with a `frame_id` of *odom*, and a `child_frame_id` of *base_link*. This means we are providing the pose of the *base_link* frame within the *odom* coordinate frame.

The `NavSatFix` message has no `child_frame_id` field, but the data is reported in geographic coordinates. It is the position of the `base_link` frame _within_ the geographic frame, but the data is not given in the *base_link* frame, but in geographic coordinates.

This does present a bit of a conundrum, though: the GPS device will be placed at some offset on the robot. That offset is usually given by, e.g., a *base_link*->*gps* transform, making the *gps* frame relative to *base_link* in the ROS `tf` tree, *even though the data that we receive is actually in the geographic coordinate frame*. What this PR is (mostly correctly) doing is observing that, if the original data was given in the *gps* frame, then the output should be in the *base_link* frame, as though we have some new GPS device that is reporting our lat/long at the robot's centroid.

I could go either way on this one. Thoughts, @SteveMacenski?

Timple:
Thanks for the comment, your first sentence doubts the `base_link` frame is the right one. But your last paragraph explains exactly what is happening and why `base_link` should be used. 
The data represented in the `gps/filtered` topic is a representation of `base_link`. And thus indeed behaves like a fictional gps device at `base_link` when converted back to the geographic coordinate frame. The coordinates won't closely match the original gps device coordinates.

Alternative is a much larger PR where a TF transformation back to the gps device location is made and the `gps_frame_id` from the original gps message is used (not just 'gps'). 
Can be done but usually these transformations are left to the client side of these messages in ROS.

SteveMacenski:
Admittedly I haven't spent much (any) time with the nav sat transform, however it sounds like there's 2 problems being brought up:

- the "gps" string is hardcoded and therefore might not exist or may be remapped from the data
- Is the base link the correct replacement for a hardcoded string

This line (https://github.com/cra-ros-pkg/robot_localization/blob/791953a0fb347100d126f4e934ab249f2818be13/src/navsat_transform.cpp#L532) shows the assignment of the gps frame ID to that in which the sensor data provides. 

The point of the `prepareFilteredGps()` method is to give output GPS data from the fusion, who's publisher's docs state:

```
If true, navsat_transform_node will also transform your robot’s world frame (e.g., map) position back to GPS coordinates, and publish a sensor_msgs/NavSatFix message on the /gps/filtered topic.
```

So that makes me think that it should be in the original gps frame, so I'd go with `gps_frame_id_` rather than `"gps"`. That way the input `gps/fix` and output `gps/filtered` messages have the same interfaces / TF transforms so they can be 1:1 swapped out by a client application. I don't think that should be `base_link`. That seems to me to be the solution of the spirit of the 2 problems.

Timple:
```
If true, navsat_transform_node will also transform your robot’s world frame (e.g., map) position back to GPS coordinates, and publish a sensor_msgs/NavSatFix message on the /gps/filtered topic.
```
This says the robot's world frame position, not the gps sensor's world frame position.

In the code the sensor position is transformed into the base_link frame before being output. I do not see this translation back to the sensor position in de code. Hence the position from (probably) robot_localization is not transformed back to the sensor position but  just converted to geograpic coordinates.

(Writing this from my mind, will verify my code statements on Monday. But either I'm right about the code or your quoted documentation line is wrong in that is should state 'sensor position')

Timple:
I just checked the code. [This](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/navsat_transform.cpp#L204) statements leads to the execution of [`prepareFilteredGps`](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/navsat_transform.cpp#L656) which takes [`latest_world_pose_.getOrigin()`](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/navsat_transform.cpp#L662) and converts is to LL and publishes this without any conversions remaining.

So the questions is which frame `latest_world_pose_.getOrigin()` is in. This variable is set [here](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/navsat_transform.cpp#L641) in the [`odomCallback`](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/navsat_transform.cpp#L631) which subscribes to the [`odometry/filtered`](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/navsat_transform.cpp#L156) topic. 
And this odometry topic is in the base_link frame as also assumed [here](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/navsat_transform.cpp#L634).

Therefore the `gps/filtered` topic is in the `base_link` frame. And can (thus) not directly be compared to the `gps/fix` topic unless this sensor is placed at the origin of this vehicle.

So I guess my original PR still holds.

ayrton04:
Yes, agreed. If we're following the standard of "gps sensor frame is its offset from the robot's origin," then the PR makes sense. We're not correcting for that offset when we publish the filtered GPS message.

ayrton04:
Thanks again.

Timple:
Without being too forward, is this change and all others in the last 5 months enough to warrant a new debian release? :-)

ayrton04:
It can be. I honestly just don't get the cycles to work on this package, and my work ROS setup makes it hard to use bloom. I'll try to get to it soon.

Issue #520, opened 2020-01-11
Matrix size error in RosRobotLocalizationListener::getState()


mabelzhang:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - From source
- Version or commit hash:
  - melodic-devel ba4010aa44d532ac509cbeabc7b95c71e785cdc9

### Steps to reproduce issue

This bug was found while porting classes to Dashing. The test gets an Eigen run-time error in Dashing, but doesn't get an error in Melodic, for whatever reason, though the matrix size is just as wrong in Melodic.

The test `LocalizationListenerTest.GetStateOfRelatedFrame` in `test_ros_robot_localization_listener.cpp` exposes a matrix dimensions error in matrix multiplication in `ros_robot_localization_listener.cpp` that isn't caught by Eigen for some reason.

The offending line is https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ros_robot_localization_listener.cpp#L504

It multiplies a `rot_6d` matrix which is 6 x 6, by a covariance matrix which in the test is 15 x 15, by the transpose of the 6 x 6 matrix. I added these 4 lines to inspect the matrices:
```
std::cerr << rot_6d << std::endl;
std::cerr << estimator_state.covariance.eval() << std::endl;
std::cerr << (rot_6d * estimator_state.covariance.eval()) << std::endl;
std::cerr << (rot_6d * estimator_state.covariance.eval() * rot_6d.transpose()) << std::endl;
```

Then run the test:
```
$ rosrun robot_localization test_ros_robot_localization_listener_publisher
$ rosrun robot_localization test_ros_robot_localization_listener
```

This prints 4 matrices with size 6 x 6, 15 x 15, 6 x 15 (!!), 6 x 6 (!!).

### Expected behavior
Test fails. The covariance matrix's size depends on the `STATE_SIZE`, which is declared in `include/robot_localization/filter_common.h`:
```
const int STATE_SIZE = 15;
```

### Actual behavior
Test passes.

### Additional information

In Dashing, I'm not sure what is different, but running the test will result in run-time error from Eigen, which is actually correct:
```
[...]

PR #521, opened 2020-01-11
Port to Dashing: robot_localization_estimator and ros_robot_localization_listener

mabelzhang:
As mentioned [here](https://github.com/cra-ros-pkg/robot_localization/issues/473#issuecomment-561822472), we're trying to decrease the differences between the Melodic branch and the Dashing branch, to increase our confidence in running the Dashing branch on our robot.

This PR ports commits 11e0da9..e912e40 in https://github.com/cra-ros-pkg/robot_localization/compare/dashing-devel...melodic-devel to Dashing.

Those are 71 commits that show up as diffs between `melodic-devel` and `dashing-devel`.
61 of those commits consist of adding these new files:
```
robot_localization_estimator.*
ros_robot_localization_listener.*
robot_localization_listener_node.*
test_robot_localization_estimator.*
test_ros_robot_localization_listener.*
test_ros_robot_localization_listener_publisher.*
```

At the end of it I realized the last few commits also include the enable service for `RosFilter`, `enableFilterSrvCallback()` and some minor changes. Didn't realize that before I started - I just grabbed everything from that date. Sorry it isn't as clean as it can be. Let me know if I need to break this up.

During the process, a bug in `ros_robot_localization_listener` is exposed and filed here https://github.com/cra-ros-pkg/robot_localization/issues/520


include/robot_localization/robot_localization_estimator.hpp
```diff
@@ -0,0 +1,224 @@
...
+{
+  ExtrapolationIntoFuture = 0,
+  Interpolation,
+  ExtrapolationIntoPast,
+  Exact,
+  EmptyBuffer,
+  Failed
+};
+}  // namespace EstimatorResults
+typedef EstimatorResults::EstimatorResult EstimatorResult;
```
SteveMacenski: You have `EstimatorResult` inside a namespace, then immediately `typedef` it away of the same name. I think you should either `using namespace EstimatorResults` or not include a namespace for this enum at all. 

Same below with filter types.

include/robot_localization/robot_localization_estimator.hpp
```diff
@@ -0,0 +1,224 @@
...
+{
+public:
+  //! @brief Constructor for the RobotLocalizationListener class
+  //!
+  //! @param[in] args - Generic argument container (not used here, but needed so that the ROS filters can pass arbitrary
+  //! arguments to templated filter types).
+  //!
+  explicit RobotLocalizationEstimator(unsigned int buffer_capacity,
+                                      FilterType filter_type,
+                                      const Eigen::MatrixXd& process_noise_covariance,
```
SteveMacenski: I think you forgot to lint your changes `const Eigen::MatrixXd& `?

include/robot_localization/robot_localization_estimator.hpp
```diff
@@ -0,0 +1,224 @@
...
+  //!
+  //! @brief The buffer holding the system states that have come in. Interpolation and extrapolation is done starting
+  //! from these states.
+  //!
+  boost::circular_buffer<EstimatorState> state_buffer_;
+
+  //!
+  //! @brief A pointer to the filter instance that is used for extrapolation
+  //!
+  FilterBase* filter_;
```
SteveMacenski: Can we wrap this as a smart pointer (or template to just be the base object type, but that's more work)?

include/robot_localization/ros_robot_localization_listener.hpp
```diff
@@ -0,0 +1,181 @@
...
+  //! @param[in] accel message
+  //!
+  void odomAndAccelCallback(
+    const std::shared_ptr<nav_msgs::msg::Odometry const>& odom,
+    const std::shared_ptr<geometry_msgs::msg::AccelWithCovarianceStamped const>& accel);
+
+  //! @brief The core state estimator that facilitates inter- and
+  //! extrapolation between buffered states.
+  //!
+  RobotLocalizationEstimator* estimator_{nullptr};
```
SteveMacenski: Can we wrap this as a unique_ptr?

src/robot_localization_estimator.cpp
```diff
@@ -0,0 +1,224 @@
...
+  else if ( filter_type == FilterTypes::UKF )
+  {
+    if ( filter_args.size() < 3 )
+    {
+      filter_ = new Ukf();
+    }
+    else
+    {
+      filter_ = new Ukf();
+      static_cast<Ukf*>(filter_)->setConstants(filter_args[0], filter_args[1], filter_args[2]);
```
SteveMacenski: I think this should be `dynamic_cast` to ensure we correctly cast these changes.

This is actually one area where the templating is nice since you can just work with the object and not require type casting

src/ros_filter.cpp
```diff
@@ -1793,6 +1804,15 @@ void RosFilter<T>::initialize()
 template<typename T>
 void RosFilter<T>::periodicUpdate()
 {
+  // Wait for the filter to be enabled
+  if (!enabled_)
+  {
+    RCLCPP_INFO_ONCE(this->get_logger(),
```
SteveMacenski, line 1809: We got `XYZ_ONCE` in ROS2?!? :heart:

src/ros_robot_localization_listener.cpp
```diff
@@ -0,0 +1,548 @@
...
+    return FilterTypes::UKF;
+  }
+  else
+  {
+    return FilterTypes::NotDefined;
+  }
+}
+
+RosRobotLocalizationListener::RosRobotLocalizationListener(
+  rclcpp::Node::SharedPtr node=rclcpp::Node::make_shared("robot_localization")):
```
SteveMacenski: I'm super duper not a fan of having nodes being implicitly created given how much overhead they create in ROS2. Is there a reason we need this default argument?

SteveMacenski (changes requested):
Pretty small bones:

- I think you need to run a linter over this PR
- Converting the few raw pointers introduced into smart pointers (or templating classes, which, I'm not necessarily advocating for)
- Couple one-liner changes for safety

include/robot_localization/robot_localization_estimator.hpp
mabelzhang: Thank you! Fix linter errors in ceceac6

src/ros_robot_localization_listener.cpp
mabelzhang: I am not sure whether it is needed. The original [constructor](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ros_robot_localization_listener.cpp#L69) in Melodic constructs a `NodeHandle` with `robot_localization`, and I don't know if anyone using it assumes that namespace in the topic names. So I just kept it. I'm constructing it with a different name in test and using that name in the test launch. The only other use case in the code is in `ros_robot_localization_listener.cpp`, and that uses a different name too.

It needs to be a valid node pointer, because `odom_sub_` and `accel_sub_` are constructed using it. I don't know whether it fails safely if the pointer is invalid.

I removed the default argument in 926a4e4.

src/ros_filter.cpp
mabelzhang: Yeah, those are in [Dashing](http://docs.ros2.org/dashing/api/rclcpp/logging_8hpp.html). In [Eloquent](http://docs.ros2.org/latest/api/rclcpp/logging_8hpp.html) there is also `*_THROTTLE` and `*_STREAM`. I thought of simply porting to Eloquent so I didn't have to reformat all the `_STREAM` calls, but just stuck with Dashing... It would have been easier to port to Eloquent.

src/robot_localization_estimator.cpp
mabelzhang: Changed in 926a4e4
I did not do the extra work for templating...

include/robot_localization/ros_robot_localization_listener.hpp
mabelzhang: 926a4e4

include/robot_localization/robot_localization_estimator.hpp
mabelzhang: 926a4e4
Thanks for catching that. I wrapped one or two but didn't see these.

include/robot_localization/robot_localization_estimator.hpp
mabelzhang: 926a4e4
I removed the typedef and just typed out the `EstimatorResults::EstimatorResult` in the one or two places that used it. Same for `FilterTypes::FilterType`.

mabelzhang:
Should I disable the failing test in #520 here?

src/ros_robot_localization_listener.cpp
SteveMacenski: Makes sense. Nodehandles in ROS1 (to my understanding, which I'm opening myself up to being completely wrong) had trivial overhead. I'd like make'm to grab a param and trash them. ROS2 Nodes are a different beast creating their own DDS participants and coming with alot of online networking traffic. 

But it sounds like you just removed the default argument, so that resolves my concern.

src/ros_filter.cpp
SteveMacenski: WE GOT `THROTTLE` TOO?!

You're speaking my language

SteveMacenski:
Not sure on your question -  I'll throw that football to Tom.

Let me know if you want me to take another pass again if you're done with the changes

src/ros_robot_localization_listener.cpp
mabelzhang: Interesting. The Melodic version did actually have a second node in the constructor doing exactly what you just said, grabbing a few parameters, heh... I guess that isn't weird.

mabelzhang:
Yep done with changes for now. Please take another pass. Thank you for the quick reviews!

SteveMacenski (approved):
LGTM

Still need discussion on the tests.

ayrton04:
So this node was [a user contribution](https://github.com/cra-ros-pkg/robot_localization/pull/324), and while I ran it and reviewed the code before merging, I am not especially familiar with it. Might be worth running it by @rokusottervanger.

mabelzhang:
731b3fb added fix from https://github.com/cra-ros-pkg/robot_localization/pull/525
That fixed the test in Dashing.

SteveMacenski:
@rokusottervanger  doesn't look like he's been active in awhile. I'd air on the side of "merge it" in that case if this is just a cherry pick anyhow. 

@mabelzhang anything else before it goes in?

mabelzhang:
(Eh @rokusottervanger responded in #520 and that resulted in #525)

Nothing else from my side.

SteveMacenski:
Ah ok - lets wait a bit for a response then

mabelzhang:
Actually I have a question, is the changelog maintained in `dashing-devel`?

The changelog in melodic-devel reflects some of the changes in my 2 PRs https://github.com/cra-ros-pkg/robot_localization/commit/93c53a8a0f2b7bba5e947a55534f0d14a1be8d1a
I'm not sure whether they should be added to the changelog in dashing-devel. If so, should the dates be 2020, instead of the original dates?

SteveMacenski:
I'll say for sure I didn't update it. Some tools autogenerate it on release. 

Up to you what you'd like to do, unless @ayrton04 has a preference.

mabelzhang:
I probably would not touch it then, since other stuff aren't in it. It would not be good for people to see recent stuff in there and assume it is complete, when many things aren't in it. Better to be consistent.

mabelzhang:
I'm re-reading #520 and looks like @reinzor took over the issue from @rokusottervanger and implemented that bug fix addressing the test issue here. I would think since we closed that issue and merged that fix into melodic, the same fix here should also be good to go. It's a pretty straight-forward fix - takes upper left 6 x 6 of the 15 x 15 matrix.

mabelzhang:
Ohhh, I am now seeing that fix in a later commit in `melodic-devel` https://github.com/cra-ros-pkg/robot_localization/commit/7f55b81d35e112334e2c7bc20a8dbdf9aad9962b

I am just now getting to that commit in my porting. So I think that is definitely the intended fix for Dashing as well.

mabelzhang:
I just noticed that same fix in a later commit 7f55b81 in `melodic-devel`. I'm just getting to that point in the porting. So that's definitely the intended fix for Dashing as well. Also updated a few todo comments according to that commit.

SteveMacenski:
.... so merge it? Or wait?

mabelzhang:
I think it's okay to merge. You can wait if you want to.

PR #522, opened 2020-01-16
Fix frame id of imu in differential mode, closes #482.

### LINKED ISSUE #482: A possible invalid assumption regarding frame id of IMU in differential mode
I've experienced some problems with the state estimation _only_ when I use (the orientation of) an IMU in `differential` mode. If I set `differential` mode to `false`, everything is fine. This line https://github.com/cra-ros-pkg/robot_localization/blob/086bcfcc81d422d357e2ecabbadcca68624a0062/src/ros_filter.cpp#L2534 seems suspicious to me. I would expect that it should use the `frame_id` of the IMU message. If I replace that line with `poseTmp.frame_id_ = msg->header.frame_id;` everything looks fine. Any idea why the frame id should be set to `finalTargetFrame` if differential mode is used?

RonaldEnsing:
This should fix #482

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2531,7 +2531,7 @@ namespace RobotLocalization
     {
       // Otherwise, we should use our target frame
       finalTargetFrame = targetFrame;
-      poseTmp.frame_id_ = (differential ? finalTargetFrame : msg->header.frame_id);
+      poseTmp.frame_id_ = msg->header.frame_id;
     }
 
     RF_DEBUG("Final target frame for " << topicName << " is " << finalTargetFrame << "\n");
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2531,7 +2531,7 @@ namespace RobotLocalization
     {
       // Otherwise, we should use our target frame
       finalTargetFrame = targetFrame;
-      poseTmp.frame_id_ = msg->header.frame_id;
+      poseTmp.frame_id_ = (differential && !imuData ? finalTargetFrame : msg->header.frame_id);
     }
 
     RF_DEBUG("Final target frame for " << topicName << " is " << finalTargetFrame << "\n");


ayrton04:
Just letting you know I haven't forgotten about this. I need to test it with some non-IMU data. I'm worried that this change will fix your IMU case, but break others. It may be that the correct solution is to transform the data before differentiating it, as we discussed before. I'll test it soon.

SteveMacenski:
@ayrton04 friendly ping :-)

ayrton04:
Yeah, sorry, I need to test this with some sample data of differential poses and make sure it doesn't break how it's intended to work in non-IMU cases. My plan was to play it safe and just differentiate the pose after we transform it into the target frame. I feel like that will definitely cover all the use cases. I'll try to get to this before the end of next week (truly).

ayrton04:
So I am working on this now, and yeesh, I am not happy about the way I am doing a bunch of things here. Looking at old code is humbling.

Anyway, my new strategy was going to be to first transform the pose into the target frame, and then differentiate it using the previous (already transformed) pose, but a quick whiteboard proved that I was being dumb:

I was proposing doing `(target_frame_trans * previous_pose)^-1 * (target_frame_trans * current_pose)`. But that becomes `previous_pose^-1 * target_frame_trans^-1 * target_frame_trans * current_pose`, which is just `previous_pose^-1 * current pose`.

Anyway, I'm going to address this. If I had more time, I'd probably rewrite the entire `preparePose` method. Alas.

ayrton04:
Actually, scratch that. That again assumes that the IMU was mounted neutrally. That differentiation gets us the rotational velocity in the frame of the IMU, not the robot frame. So we have to transform it into the body frame of the robot.

ayrton04:
OK, after diving a bit deeper, I understand the issue. Your fix will correctly handle IMU data, but I think it would break normal pose data differentiation.

The trouble is on [this line](https://github.com/cra-ros-pkg/robot_localization/blob/799ad4481b8b16b12a50c34725dbbe4d82b4a84a/src/ros_filter.cpp#L2804). For differential mode, we first do `previous_pose^-1 * current_pose`. Consider a standard pose message that gets reported in the `odom` frame. If we differentiate two of those, that data is _already_ in the robot's body frame. (Note that this is also true if the data is reported in some world-fixed frame _relative_ to odom: `pose_1^-1 * pose2` will always get you a body-frame relative pose). So we don't need to perform any further transformations for differential mode. That's why [this line](https://github.com/cra-ros-pkg/robot_localization/blob/799ad4481b8b16b12a50c34725dbbe4d82b4a84a/src/ros_filter.cpp#L2576) was the way it was: if we're in differential mode, we say that the target frame and the frame the data is reported in are the same, which will prevent any data getting transformed _after_ the differentiation.

But IMUs are a special case. After the differentiation process, we now have velocity data in the _IMU_'s body frame, and not the robot's.

So then what we really want is:

1. If the data is normal pose data, when we differentiate it, we don't want to apply any further transforms
2. If the data is IMU data, then after differentiation, we need to rotate it into the `base_link` frame.

With that, I will propose a small tweak to your PR that should fix this. Even if I'm wrong, it limits your change to IMU data only, which was the purpose of the PR.

src/ros_filter.cpp
```diff
@@ -2531,7 +2531,7 @@ namespace RobotLocalization
     {
       // Otherwise, we should use our target frame
       finalTargetFrame = targetFrame;
-      poseTmp.frame_id_ = (differential ? finalTargetFrame : msg->header.frame_id);
+      poseTmp.frame_id_ = msg->header.frame_id;
```
ayrton04:
```suggestion
      poseTmp.frame_id_ = (differential && !imuData ? finalTargetFrame : msg->header.frame_id);
```

ayrton04: @RonaldEnsing sorry for the ping, especially as I took forever to get to this PR, but would you mind updating the PR with the code above so I can merge? If you don't have the cycles, let me know, and I'll take care of it. Thanks!

RonaldEnsing: Sure. I committed your suggestion to the branch of this PR.

PR #525, opened 2020-01-20
fix(robot_localization_listener): Only rotate upper-left covariance part

reinzor:
Solves https://github.com/cra-ros-pkg/robot_localization/issues/520


SteveMacenski (approved):

SteveMacenski:
Thanks Rein

Issue #527, opened 2020-01-24
[Windows] unresolved rclcpp operator+


jnoyola:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - Source
- Version or commit hash:
  - 534a719246f3c6ac8e8aa7d705a494b79aca6610

#### Steps to reproduce issue
Build on Windows 10.

#### Expected behavior
Building should succeed.

#### Actual behavior
Linking fails with "unresolved external symbol" due to uses ([1](https://github.com/cra-ros-pkg/robot_localization/blob/534a719246f3c6ac8e8aa7d705a494b79aca6610/src/ros_filter.cpp#L1835), [2](https://github.com/cra-ros-pkg/robot_localization/blob/534a719246f3c6ac8e8aa7d705a494b79aca6610/src/ros_filter.cpp#L1898)) of an rclcpp [`operator+` that is not exported](https://github.com/ros2/rclcpp/blob/34dae0a8c9ca0da8310ebc9a343d4d356aedcde4/rclcpp/include/rclcpp/time.hpp#L126).

#### Additional information
There is a similar [`operator+` that is exported](https://github.com/ros2/rclcpp/blob/34dae0a8c9ca0da8310ebc9a343d4d356aedcde4/rclcpp/include/rclcpp/time.hpp#L91) that could easily be used instead by reversing the lhs and rhs. I'm not sure if there's some difference in behavior of the [visibility control](https://github.com/ros2/rclcpp/blob/eloquent/rclcpp/include/rclcpp/visibility_control.hpp) between Windows and Linux, but I assume it's possible this is only a Windows issue and no one's run into it yet. I can probably submit a small PR tomorrow, but I just wanted to mention this in case I'm missing something.

SteveMacenski:
@jnoyola PRs appreciated. I don't think Tom or I have a windows machine we use to even know if we fixed the issue.

PR #528, opened 2020-01-25
Fix unresolved rclcpp operator+

jnoyola:
Addresses #527 for Windows.
[`operator+(Duration, Time)`](https://github.com/ros2/rclcpp/blob/34dae0a8c9ca0da8310ebc9a343d4d356aedcde4/rclcpp/include/rclcpp/time.hpp#L126) is not exported.
[`operator+(Time, Duration)`](https://github.com/ros2/rclcpp/blob/34dae0a8c9ca0da8310ebc9a343d4d356aedcde4/rclcpp/include/rclcpp/time.hpp#L91) is exported, but now requires an explicit cast since the rhs was previously implicitly casted.


SteveMacenski (approved):
LGTM

PR #526, opened 2020-01-22
Port to Dashing: melodic-devel 2.5.1 to 2.6.2

mabelzhang:
In sequence of #521 and #523, this PR ports commits in the diffs of https://github.com/cra-ros-pkg/robot_localization/compare/dashing-devel...melodic-devel.

Specifically, this PR ports commits in the melodic [CHANGELOG](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/CHANGELOG.rst) items of 2.5.1 to 2.6.2, inclusive. Rough equivalent of 
```
git cherry-pick 6dbbf3e^..82a1ad1
```
but skipping the `CHANGELOG` and version bump commits.

I'm grouping these together because the next commit in sequence, https://github.com/cra-ros-pkg/robot_localization/commit/58bef056cf84856eb5e91a89f643017c448669ff, changes actual math, and I don't want to mess Dashing users up in case that changes behavior. I'll put that in a separate PR later.

We are actively testing `dashing-devel` branch on a robot as things are merged. If there are problems in the porting, hopefully we spot them.

PR #523, opened 2020-01-18
Port to Dashing: Bug fixes from melodic-devel Dec 07 2017

mabelzhang:
This PR continues to port diffs between https://github.com/cra-ros-pkg/robot_localization/compare/dashing-devel...melodic-devel.
Specifically, it ports the 11 commits from Dec 07 2017, corresponding to
```
$ git cherry-pick 312b52a^..f7f98d0
```

They are a series of bug fixes, small diffs but potentially important bugs.

The fixes being ported correspond to this subset in the change log:
```
 2.5.0 (2017-12-15)
------------------
* Fixing datum precision
* Fixing timing variable
* Fixing state history reversion
* Fixing critical bug with dynamic process noise covariance
* Fix typo in reading Mahalanobis thresholds.
* Zero out rotation in GPS to base_link transform
* Update xmlrpcpp includes for Indigo support
* Removing lastUpdateTime
* Fixing timestamps in map->odom transform
```

Porting of `312b52a` seems to have fixed 2 tests in `dashing-devel`, `FilterBaseDiagnosticsTest.TimestampsBeforeSetPose` and `FilterBaseDiagnosticsTest.TimestampsBeforePrevious`.

PR #530, opened 2020-01-29
Port to Dashing: melodic-devel 2.6.3 to current

mabelzhang:
In sequence of #526, this PR ports commits in the diffs of https://github.com/cra-ros-pkg/robot_localization/compare/dashing-devel...melodic-devel.

Specifically, this PR ports commits in the melodic [CHANGELOG](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/CHANGELOG.rst) items of 2.6.3 to latest commits not included in the changelog, up to and including ba4010a.

Very rough equivalent of 
```
git cherry-pick 9678d25^..ba4010a
```
with exceptions in the breakdown of commits below:

- 21	Ported
- 13	Ported - merges of commits ported
- 3	Already in Dashing
- 3	Not ported - ROS 1-specific
- 1	Not ported - merges of commits not ported
- 6	Not ported - CHANGELOG, version bumps
- 1	Not ported - github administrative
- 48	Total

This concludes my series of ports from Melodic to Dashing. Final stats of the number of commits in https://github.com/cra-ros-pkg/robot_localization/compare/dashing-devel...melodic-devel ported in #521, #523, #526 and this PR are as follows:

- 118	Ported
- 3	Already in Dashing
- 14	Not ported - ROS 1-specific, e.g. nodelets, CMakeLists, XmlRpc
- 33	Not ported - CHANGELOG, version bumps, merge, github administrative
- 168	Cumulative total number of commits looked at

All tests in Dashing are passing. The `FilterBaseDiagnosticsTest.EmptyTimestamps` test that was failing in #521 had been fixed in either #523 or #526.

After this PR is merged, this should be in a more complete state to be released to Dashing.

Regarding Eloquent - There are some printouts that use `_STREAM`, `_ONCE`, `_THROTTLE`, that are commented out. They now exist in Eloquent and can be brought back. It would be nice if an `eloquent-devel` branch can be opened so that those printouts can be cleaned up and whatever warnings can be fixed in Eloquent. Though I just ported stuff to Dashing, we're actually testing in Eloquent. I can open a separate issue for that.

I'm also still thinking of adding whatever items I ported into the `CHANGELOG`... That way whoever looking at this later won't be confused about what items have been ported and which few have been skipped.


CMakeLists.txt
```diff
@@ -36,6 +41,8 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "srv/SetDatum.srv"
   "srv/SetPose.srv"
   "srv/ToggleFilterProcessing.srv"
+  "srv/FromLL.srv"
+  "srv/ToLL.srv"
```
JWhitleyWork, line 45: Would you mind alpha-ordering these? It's easier to see changes/missing values/added values in an alphabetized list.

include/robot_localization/navsat_conversions.hpp
```diff
@@ -204,13 +204,17 @@ static inline char UTMLetterDesignator(double Lat)
...
+ *
+ * @param[out] gamma meridian convergence at point (degrees).
  *
  * Written by Chuck Gantz- chuck.gantz@globalstar.com
  */
 static inline void LLtoUTM(
   const double Lat, const double Long,
   double & UTMNorthing, double & UTMEasting,
-  std::string & UTMZone)
+  std::string & UTMZone, double & gamma)
```
JWhitleyWork, line 217: [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Output_Parameters) (ROS2-recommended guidelines) says that an output parameter like this should be a pointer to a non-`const` rather than a reference. This should be true for ROS1 as well. I know this is a bigger issue that I'll address in a different ticket but the question is: Are we treating the ROS2 version of this package as an extension of the ROS1 package (meaning I should create the MR against ROS1) or treating them separately but cross-applying MRs that make sense (meaning I should create two MRs - one for each version)?

include/robot_localization/navsat_conversions.hpp
```diff
@@ -253,7 +257,9 @@ static inline void LLtoUTM(
 
   // Compute the UTM Zone from the latitude and longitude
   char zone_buf[] = {0, 0, 0, 0};
-  snprintf(zone_buf, sizeof(zone_buf), "%d%c", ZoneNumber,
+  // We &0x3fU to let GCC know the size of ZoneNumber. In this case, it's under
+  // 63 (6bits)
+  snprintf(zone_buf, sizeof(zone_buf), "%d%c", ZoneNumber & 0x3fU,
```
JWhitleyWork, line 262: Again, the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Streams) says to use streams where appropriate (I would say they are appropriate here) instead of the C equivalents. See question above about where to submit this change.

src/ros_filter.cpp
```diff
@@ -1400,18 +1407,48 @@ void RosFilter<T>::loadParams()
...
+        pose_update_vec.begin() + POSITION_A_OFFSET + ACCELERATION_SIZE, 0);
 
       std::vector<bool> twist_update_vec = update_vec;
+      // IMU message contains no information about linear speeds, filter
+      // everything except angular velocity
       std::fill(twist_update_vec.begin() + POSITION_OFFSET,
         twist_update_vec.begin() + POSITION_OFFSET + POSE_SIZE, 0);
-      std::fill(
-        twist_update_vec.begin() + POSITION_A_OFFSET,
+      std::fill(twist_update_vec.begin() + POSITION_V_OFFSET,
```
SteveMacenski, line 1449: Do you know why we're changing the fill here? This has a pretty different effect, just making sure that its what we should be doing.

JWhitleyWork (commented):
See feedback.

SteveMacenski (commented):
My general sentiment here is "I don't really understand everything that's happening in this PR". 

But, since it passes tests, you can show it's working in your application, and its just a cherry pick of prior accepted PRs, I'm willing to go with it.

I went through everything and don't see anything that I take specific issue with.

include/robot_localization/navsat_conversions.hpp
SteveMacenski: I think that if there's other instances of this, we shouldn't block a cherry-pick PR from this codebase.

include/robot_localization/navsat_conversions.hpp
SteveMacenski: Similar comment as above: If there's other situations where we do this, I don't think its worth changing at this stage

include/robot_localization/navsat_conversions.hpp
JWhitleyWork: I agree, I just wanted clarification on how the ROS1/ROS2 thing was being handled to know where to file the MR. Not intending to hold approval just for this.

SteveMacenski: Pedantic, but IMO references are fine and are preferred, google doesn't even follow google's style guide in many cases. I avoid passing by pointers where feasible, I think that creates opportunity for people not used to pointers to royally mess up. Readability / understand-ability to the largest audience reasonable is king.

src/ros_filter.cpp
mabelzhang: I don't really know what's going on with most of these ports either. This one comes from this commit https://github.com/cra-ros-pkg/robot_localization/commit/0d32daf42d4b019a2d180ac53e01582cf22165b7

SteveMacenski (approved):
I approve this, but would like @ayrton04 to take a full look.

Don't mean to pile onto his time, but there's enough happening here that he might be able to just look at and say "oh yeah, I remember that PR, all good".

After this though, I think Dashing is complete and we can run without as much oversight :cactus: 

_onto eloquent_

CMakeLists.txt
mabelzhang: Done in 04684be

JWhitleyWork (approved):
LGTM.

mabelzhang:
@SteveMacenski Could we get a `eloquent-devel` branch opened in this repo? I compiled the latest Dashing branch in Eloquent and brought back some of the ROS_ macros that are now in Eloquent. I'll open a PR targeting that branch once it exists.

SteveMacenski:
Shouldn’t we wait for this to be in so we don’t introduce fun merge conflicts?

include/robot_localization/navsat_conversions.hpp
ayrton04: This header comes from another package whose standards don't match the rest of the package anyway. If we want to do a PR later to bring everything in line with standards, that's fine, but I'm fine with leaving this as-is for now.

src/ros_filter.cpp
ayrton04: This was a fix for an issue whereby we were fusing non-existent data in IMU messages:

https://github.com/cra-ros-pkg/robot_localization/issues/432

ayrton04 (approved):
Yes, I remember these changes.

I can't thank everyone enough for all the attention this package is getting!

SteveMacenski:
I’ll open an eloquent branch when I get to the office.

SteveMacenski:
https://github.com/cra-ros-pkg/robot_localization/tree/eloquent-devel

mabelzhang:
Thank you all for the quick feedback!

PR #536, opened 2020-02-10
Ensure robot_localization filters use ROS 2 time properly.

hidmic:
Precisely what the title says, though properly is a bit of an overstatement. Necessary to operate in simulation.

### DIFF
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -771,16 +771,6 @@ protected:
   //!
   rclcpp::TimerBase::SharedPtr timer_;
 
-  //! @brief Time bookkeeping for periodic filter updates
-  //!
-  //! @details This is necessary to properly operate in simulation
-  //!
-  rclcpp::Time last_update_time_{0, 0, RCL_ROS_TIME};
-
-  //! @brief Filter updates timespan
-  //!
-  rclcpp::Duration update_timespan_{0};
-
   //! @brief optional signaling diagnostic frequency
   //!
   std::unique_ptr<diagnostic_updater::HeaderlessTopicDiagnostic> freq_diag_;
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -771,6 +771,16 @@ protected:
   //!
   rclcpp::TimerBase::SharedPtr timer_;
 
+  //! @brief Time bookkeeping for periodic filter updates
+  //!
+  //! @details This is necessary to properly operate in simulation
+  //!
+  rclcpp::Time last_update_time_{0, 0, RCL_ROS_TIME};
+
+  //! @brief Filter updates timespan
+  //!
+  rclcpp::Duration update_timespan_{0};
+
   //! @brief optional signaling diagnostic frequency
   //!
   std::unique_ptr<diagnostic_updater::HeaderlessTopicDiagnostic> freq_diag_;
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1874,9 +1874,10 @@ void RosFilter<T>::initialize()
   }
 
   double timespan = 1.0 / frequency_;
-  int timespan_ms = static_cast<int>(timespan * 1000);
+  update_timespan_ = rclcpp::Duration(
+    static_cast<rcl_duration_value_t>(std::floor(timespan * 1e9)));
   timer_ = this->create_wall_timer(
-    std::chrono::milliseconds(timespan_ms),
+    update_timespan_.to_chrono<std::chrono::milliseconds>(),  // assumes RTF >= 1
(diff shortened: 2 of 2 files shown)


src/ros_filter.cpp
```diff
@@ -1893,6 +1894,13 @@ void RosFilter<T>::periodicUpdate()
 
   rclcpp::Time cur_time = this->now();
 
+  const rclcpp::Duration update_timespan_lower_bound = update_timespan_ * 0.999;
+  if (update_timespan_lower_bound > cur_time - last_update_time_) {
```
SteveMacenski: This is being called from a wall timer, why add this? 

If you're concerned about using a timer based on wall time vs simulation time, then we should be making a `timer_` based on the current clock. Checking this is an unacceptable hack if its goal is to deal with simulation time not running at 1.0x speed.

include/robot_localization/ros_filter.hpp
```diff
@@ -771,6 +771,16 @@ class RosFilter : public rclcpp::Node
   //!
   rclcpp::TimerBase::SharedPtr timer_;
 
+  //! @brief Time bookkeeping for periodic filter updates
+  //!
+  //! @details This is necessary to properly operate in simulation
+  //!
+  rclcpp::Time last_update_time_{0, 0, RCL_ROS_TIME};
```
SteveMacenski: You're hardcoding a clock here? Shouldn't this be decided at runtime based on the node's clock?

SteveMacenski (changes requested):

src/ros_filter.cpp
hidmic: > If you're concerned about using a timer based on wall time vs simulation time

That's correct. Specifically, using wall time may result in multiple `odometry/filtered` messages being published at the same simulation time, which may (and in some cases will) cause trouble downstream. Some SLAM packages are not particularly happy with this.

> then we should be making a timer_ based on the current clock. Checking this is an unacceptable hack if its goal is to deal with simulation time not running at 1.0x speed.

I absolutely agree with you, but unfortunately that's not possible in Dashing nor Eloquent, and it may or may not be possible in Foxy either -- a fix is currently in the [roadmap](https://index.ros.org/doc/ros2/Roadmap/#new-features) (see second bullet). `rclcpp` API is a bit deceiving in this regard, because while you can instantiate a timer that uses the current clock and [computes durations using that source](https://github.com/ros2/rcl/blob/41f4fba9664a3ed4a0ddc9985d510f04e10418a4/rcl/src/rcl/wait.c#L568-L570), under the hood it sleeps as if [that duration was specified in system time](https://github.com/ros2/rcl/blob/41f4fba9664a3ed4a0ddc9985d510f04e10418a4/rcl/src/rcl/wait.c#L606-L613). So the `RTF <= 1.0` precondition remains. FWIW, [ROS1 solution wasn't much better](https://github.com/ros/ros_comm/blob/c16649772847c360d6b5275f2d54a2a728f49d8d/clients/roscpp/include/ros/timer_manager.h#L587-L590).

Having said that, I can change the code to use a proper timer ~so that when the feature lands we can drop these additional checks. WDYT?~ On a second look, for timers these checks may be redundant. Rates are the ones that need them. Changing.

SteveMacenski: I think revert and wait for the feature to be properly supported. Its annoying  but if we merge this, that hack will never go away because functionally we'll all forget and then there will be a weird cause of non-determinism in simulation

hidmic: Done in https://github.com/cra-ros-pkg/robot_localization/pull/536/commits/d5629d4a1571205df7412d219e4b953862c78818.

include/robot_localization/ros_filter.hpp
hidmic: This is no longer necessary, but `robot_localization` ROS filter nodes always use `RCL_ROS_TIME` clocks (i.e. system time or simulation time depending on `use_sim_time` parameter value). It's true it's way better to query than to duplicate though.

hidmic:
Thanks for the review @SteveMacenski !

src/ros_filter.cpp
hidmic: Oh, I just see your last comment. The last commit is as correct as it can be, and it'll remain correct once the feature is properly supported. You do have to bear in mind not to use `RTF >= 1.0` (less than, not greater, my bad), but that's true for every node.

SteveMacenski (approved):
Boom, that looks like it works.

Did you test that you get the right rates in simulation and in real time?

hidmic:
Well, it looks like I never posted my last comment.

When configuring `robot_localization` to publish at a `frequency` of  50Hz, I get ~48Hz when `use_sim_time` is `true` and ~50Hz when `use_sim_time` is `false`. It's probably skipping a couple iterations every now and then because of latency in clock publications, which is expected considering the caveats in Dashing and Eloquent implementation described above.

PR #539, opened 2020-02-14
Fix #538: Add launch_ros exec_depend to Dashing

### LINKED ISSUE #538: Dashing debians failed to process due to missing launch_ros dependency
@mabelzhang  - can you PR this?

mabelzhang:
Fixes #538 for Dashing release

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -40,6 +40,7 @@
   
   <exec_depend>rclcpp</exec_depend>
   <exec_depend>rmw_implementation</exec_depend>
+  <exec_depend>launch_ros</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>


package.xml
```diff
@@ -40,6 +40,7 @@
   
   <exec_depend>rclcpp</exec_depend>
   <exec_depend>rmw_implementation</exec_depend>
+  <exec_depend>launch_ros</exec_depend>
```
SteveMacenski, line 43: `test_depend` perhaps?

mabelzhang: I tried that, it doesn't compile with that even with launch_ros installed. I consulted @clalancette and he said launch_ros needs to be exec_depend here. It compiles with that for both of us. Should we give that a try?

SteveMacenski: Ah, no I think you're right. I was a little quick on the trigger there.

PR #541, opened 2020-02-18
Fix FilterBaseDiagnosticsTest.EmptyTimestamps test on amd64

mabelzhang:
Testing buildfarm behavior on EmptyTimestamps test...

### DIFF
--- a/test/test_filter_base_diagnostics_timestamps.cpp
+++ b/test/test_filter_base_diagnostics_timestamps.cpp
@@ -156,7 +156,7 @@ public:
     rclcpp::SystemDefaultsQoS qos = rclcpp::SystemDefaultsQoS();
     diagnostic_sub_ =
       node_->create_subscription<diagnostic_msgs::msg::DiagnosticArray>(
-      "/diagnostics", qos.keep_all(), 
+      "/diagnostics", qos.keep_all(),
       [&](diagnostic_msgs::msg::DiagnosticArray::UniquePtr msg) {
         diagnostics.push_back(*msg);
       });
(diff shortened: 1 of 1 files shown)


SteveMacenski:
ah _ha_ joke's on you, I don't have superadmin rights to enable the `build.ros2.org` webhook. No dashing CI for you!

(@ayrton04 can you enable to ROS2 build farm web hook. Its the same as ROS1 setup but now `ros2` vs `ros`.)

mabelzhang:
I looked into this more and it actually was an intermittent failure that failed on my laptop about 1/10 times. That explains why I didn't notice the failure previously. I changed the QoS to `keep_all`, and this now passes 30/30 times on my laptop.

So I think it's worth trying it on the buildfarm again, if you deem it safe to merge.

ayrton04:
Added the web hook for the ROS2 build farm.

ayrton04 (approved):

mabelzhang:
Looks like the tests passed this time

SteveMacenski:
Well I dont think this is representative. this isn't on an ARM chipset. Its x86 which wasn't in question

mabelzhang:
Wasn't the buildfarm failing on amd64? This is the one I was trying to fix http://build.ros2.org/view/Ddev/job/Ddev__robot_localization__ubuntu_bionic_amd64/
which with this PR is green here http://build.ros2.org/view/Dpr/job/Dpr__robot_localization__ubuntu_bionic_amd64/

The arm64 one I thought was a different issue. That fails on all 3 tests in filter_base. The amd64 one was only failing on the EmptyTimestamps test.

SteveMacenski:
I'm sorry, we were on different pages. I thought you were solving the ARM issue. OK yeah, this makes sense.

SteveMacenski (approved):

Issue #543, opened 2020-02-21
PkgConfig Usage


jnoyola:
Is there a reason we [use PkgConfig just for yaml-cpp](https://github.com/cra-ros-pkg/robot_localization/blob/eloquent-devel/CMakeLists.txt#L34-L35)? It looks like it was a `find_package` in dashing at some point, but maybe it was merged into melodic, which already used PkgConfig, and then merged back to dashing. I'm not positive on that.

Anyway, building on Windows in particular would be simpler and more vanilla without that single use of PkgConfig, but I'm not sure if there's a specific reason for it.

jnoyola:
This just became an issue with the recent merge into the ROS2 branches, which I'm dying to pull 😃

jnoyola:
Assuming it's ok to remove, I opened this tiny PR. `eloquent-devel` builds on Windows now.
#544

PR #544, opened 2020-02-22
remove PkgConfig

jnoyola:
PkgConfig isn't standard on Windows, so I'm removing it since it's only used for a single package that works just as well with `find_package`.

Addresses #543

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,9 +30,7 @@ find_package(tf2_eigen REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(Eigen3 REQUIRED)
-
-find_package(PkgConfig REQUIRED)
-pkg_check_modules(YAML_CPP yaml-cpp)
+find_package(yaml-cpp REQUIRED)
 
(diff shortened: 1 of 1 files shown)


SteveMacenski:
I think this should probably be targeted to Dashing. 

The eloquent branch is behind and we should cherry pick them over to eloquent. Plus Dashing is LTS so this may be useful for more users there.

jnoyola:
> I think this should probably be targeted to Dashing.

Done.

SteveMacenski:
I think what you actually want is this: https://github.com/ros/rosdistro/blob/master/dashing/distribution.yaml#L3452

`yaml_cpp` isn't a valid rosdep key, so its cannot be added to the package.xml which is required.

jnoyola:
Ah I didn't realize the current PkgConfig dep had an entry in the packages.xml. I just built the non-ros package.

I can probably update that tonight.

SteveMacenski:
Update?

jnoyola:
Ah sorry. Between the chaos of FRC season getting cut short and working on other projects, I unfortunately haven't had time to take another look at getting this to work on Windows. Feel free to cancel this PR and I can open another one in the future if needed.

PR #547, opened 2020-03-20
Fix for 'expected [double] got [integer]' error while running navsat …

ashwinsushil:
Upon running the roslaunch file using 
`ros2 launch robot_localization navsat_transform.launch.py`

The error shown below is thrown,

image

Leading to the node dying.

### DIFF
--- a/params/dual_ekf_navsat_example.yaml
+++ b/params/dual_ekf_navsat_example.yaml
@@ -1,7 +1,7 @@
 # For parameter descriptions, please refer to the template parameter files for each node.
 
 ekf_filter_node_odom:
-  frequency: 30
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

mabelzhang:
Hmm this is an instance that indicates we should probably do a fast-forward merge of eloquent from dashing... This had been fixed in the dashing branch.

SteveMacenski:
Yeah I know -- I figured it unblocks the user and we can fast forward later when we're ready. It seemed unfair to say something like "we're still working on eloquent, ignore that branch" when folks want to and can use it.

PR #549, opened 2020-03-25
Modified to get datum array param

routiful:
Hi :)

I couldn't get `datum` parameters when I had set true to `use_manual_datum`param since it was not delcared.
I replaced `get_parameter` to `declare_parameter` and it works.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -113,8 +113,10 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
   from_ll_srv_ = this->create_service<robot_localization::srv::FromLL>(
     "fromLL", std::bind(&NavSatTransform::fromLLCallback, this, _1, _2));
 
-  std::vector<double> datum_vals;
-  if (use_manual_datum_ && this->get_parameter("datum", datum_vals)) {
+  if (use_manual_datum_) {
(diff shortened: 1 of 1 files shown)


src/navsat_transform.cpp
```diff
@@ -113,8 +113,10 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
   from_ll_srv_ = this->create_service<robot_localization::srv::FromLL>(
     "fromLL", std::bind(&NavSatTransform::fromLLCallback, this, _1, _2));
 
-  std::vector<double> datum_vals;
-  if (use_manual_datum_ && this->get_parameter("datum", datum_vals)) {
+  if (use_manual_datum_) {
+    std::vector<double> datum_vals(3, 0);
```
SteveMacenski: Lets leave this as the default constructor. By doing this, you create and destroy a few objects that aren't required.  The rest looks good.

SteveMacenski (changes requested):

src/navsat_transform.cpp
routiful: I reverted the vector to leave on default constructor.

SteveMacenski (approved):

SteveMacenski:
Thanks for the fix!

PR #552, opened 2020-04-08
Fix get parameters

routiful:
I couldn't use control term since other parameters was not set.
```
use_control is set to true, but control_config is missing. No control term will be used.
```
After I was debugged, I could found that below if statement are always return false (https://github.com/cra-ros-pkg/robot_localization/blob/700ea7efb4a7e38c14a343a541f5d738453f0d5f/src/ros_filter.cpp#L858)

So, I have deleted initialization on related vectors and finally it works well.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -938,7 +938,7 @@ void RosFilter<T>::loadParams()
   filter_.setUseDynamicProcessNoiseCovariance(
     dynamic_process_noise_covariance);
 
-  std::vector<double> initial_state(STATE_SIZE, 0.0);
+  std::vector<double> initial_state;
   this->declare_parameter("initial_state");
   if (this->get_parameter("initial_state", initial_state)) {
     if (initial_state.size() != STATE_SIZE) {
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -883,6 +883,7 @@ void RosFilter<T>::loadParams()
     } else {
       std::cerr << "use_control is set to true, but acceleration_limits is "
         "missing. Will use default values.\n";
+      acceleration_limits.resize(TWIST_SIZE, 1.0);
     }
 
     this->declare_parameter("acceleration_gains");
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -844,11 +844,11 @@ void RosFilter<T>::loadParams()
 
   // Determine if we're using a control term
   double control_timeout = sensor_timeout;
-  std::vector<bool> control_update_vector(TWIST_SIZE, 0);
-  std::vector<double> acceleration_limits(TWIST_SIZE, 1.0);
-  std::vector<double> acceleration_gains(TWIST_SIZE, 1.0);
-  std::vector<double> deceleration_limits(TWIST_SIZE, 1.0);
-  std::vector<double> deceleration_gains(TWIST_SIZE, 1.0);
+  std::vector<bool> control_update_vector;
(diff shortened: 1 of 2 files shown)


SteveMacenski:
I don't know much on this one -- @ayrton04 any objections? I don't know if that specific initialization was for something in particular.

ayrton04:
It looks like we're already [checking the dimensions after init anyway](https://github.com/cra-ros-pkg/robot_localization/blob/dashing-devel/src/ros_filter.cpp#L891), so this LGTM.

SteveMacenski:
@routiful are we sure its not effecting tests? http://build.ros2.org/job/Ddev__robot_localization__ubuntu_bionic_amd64/ The last dev build on master was successful but this PR is failing on a few.

I'll admit we have some issues with tests at the moment but I'm not sure that extends to PRs vs dev builds. Can you take a look?

routiful:
@SteveMacenski Okay, I will take a look as soon as possible.

src/ros_filter.cpp
```diff
@@ -844,11 +844,11 @@ void RosFilter<T>::loadParams()
 
   // Determine if we're using a control term
   double control_timeout = sensor_timeout;
-  std::vector<bool> control_update_vector(TWIST_SIZE, 0);
-  std::vector<double> acceleration_limits(TWIST_SIZE, 1.0);
-  std::vector<double> acceleration_gains(TWIST_SIZE, 1.0);
-  std::vector<double> deceleration_limits(TWIST_SIZE, 1.0);
-  std::vector<double> deceleration_gains(TWIST_SIZE, 1.0);
+  std::vector<bool> control_update_vector;
+  std::vector<double> acceleration_limits;
```
ayrton04, line 848: So re-reading this, I feel like this changes the filter's behaviour (at least the ROS1 behaviour).

Before, if the user failed to specify, e.g., `acceleration_limits`, then we would just use the default values that were already here (a vector of six 1.0 values). [Now we just have an empty vector](https://github.com/cra-ros-pkg/robot_localization/blob/b70d03157c092871308b95c8d7de091521f2dfd5/src/ros_filter.cpp#L884), no? I'm not familiar with ROS2's parameter declaration scheme, but I'm guessing that if the user turns on `use_control`, but fails to specify any limits, these vectors will all be empty. I feel like the `else` clauses for failed parameter loading need to initialize these, e.g., you need to add this [here](https://github.com/cra-ros-pkg/robot_localization/blob/b70d03157c092871308b95c8d7de091521f2dfd5/src/ros_filter.cpp#L884):

```
acceleration_limits.resize(TWIST_SIZE, 1.0);
```

...but again, this all depends on whether `get_parameter` returns true, even if the user didn't specify the parameter.

routiful: But CI still failed..

ayrton04: Oh sorry, I wasn't suggesting that this was causing the CI failure. I was just pointing it out as a change in behaviour.

Are you able to run these tests locally? And can you run them with and without your changes?

routiful: @ayrton04 
I could use colcon_test to locally test its and finally solved the problem.
Since `control_update_vector` will be used whether it was initialized, the tests were not passed. I add resize to vector in `else` statement that will be activated when `use_control` param was set false.

SteveMacenski:
@ayrton04 are you happy with this now?

ayrton04 (approved):

SteveMacenski (approved):

PR #553, opened 2020-04-09
Unsilent

florianspy:


params/ekf_template.yaml
```diff
@@ -2,7 +2,7 @@
 # computation until it receives at least one message from one of the inputs. It will then run continuously at the
 # frequency specified here, regardless of whether it receives more measurements. Defaults to 30 if unspecified.
 frequency: 30
-
+silentchecktf: false
```
ayrton04: Can we perhaps rename this parameter? Something like `silent_tf_failure` would be more in line with the parameter scheme, I think.

src/ros_filter.cpp
```diff
@@ -2401,12 +2401,14 @@ namespace RobotLocalization
...
     tf2::Transform targetFrameTrans;
+    bool silentchecktf;
+    nh_.getParam("/silentchecktf",silentchecktf);
     bool canTransform = RosFilterUtilities::lookupTransformSafe(tfBuffer_,
                                                                 targetFrame,
                                                                 msgFrame,
                                                                 msg->header.stamp,
                                                                 tfTimeout_,
-                                                                targetFrameTrans);
+                                                                targetFrameTrans,silentchecktf);
```
ayrton04: Would you mind following the existing style for the other parameters (line break, put in line with the other params)?

florianspy: i change the files now is that now the correct format?

src/ros_filter.cpp
```diff
@@ -2401,12 +2401,15 @@ namespace RobotLocalization
     // It's unlikely that we'll get a velocity measurement in another frame, but
     // we have to handle the situation.
     tf2::Transform targetFrameTrans;
+    bool silentchecktf;
```
ayrton04:
```suggestion
    bool silent_tf_failure;
```

src/ros_filter.cpp
```diff
@@ -2401,12 +2401,15 @@ namespace RobotLocalization
...
+    nh_.getParam("/silent_tf_failure",silentchecktf);
```
ayrton04:
```suggestion
    nh_.getParam("/silent_tf_failure", silent_tf_failure);
```

src/ros_filter.cpp
```diff
@@ -2401,12 +2401,15 @@ namespace RobotLocalization
...
     bool canTransform = RosFilterUtilities::lookupTransformSafe(tfBuffer_,
                                                                 targetFrame,
                                                                 msgFrame,
                                                                 msg->header.stamp,
                                                                 tfTimeout_,
-                                                                targetFrameTrans);
+                                                                targetFrameTrans,
+								silentchecktf);
```
ayrton04:
```suggestion
                                                                silent_tf_failure);
```

ayrton04 (commented):
I made suggestions for what I'd like to fix. Sorry to nitpick. Just trying to keep the package from being even messier. Thanks for the PR!

src/ros_filter.cpp
ayrton04: I added some suggestions. Would you mind incorporating them into the PR? And then, once it passes CI, we can merge.

src/ros_filter.cpp
```diff
@@ -2401,15 +2401,15 @@ namespace RobotLocalization
     // It's unlikely that we'll get a velocity measurement in another frame, but
     // we have to handle the situation.
     tf2::Transform targetFrameTrans;
-    bool silentchecktf;
-    nh_.getParam("/silent_tf_failure",silentchecktf);
+    bool silent_tf_failure;
+    nh_.getParam("/silent_tf_failure",silent_tf_failure);
```
ayrton04: The build is reporting failure because it needs to be linted:

http://build.ros.org/job/Mpr__robot_localization__ubuntu_bionic_amd64/25/testReport/junit/roslint/LintCheck/roslint_robot_localization/

You need a space after the comma:

`nh_.getParam("/silent_tf_failure", silent_tf_failure);`

src/ros_filter.cpp
```diff
@@ -47,9 +47,9 @@ namespace RobotLocalization
 {
   template<typename T>
   RosFilter<T>::RosFilter(ros::NodeHandle nh,
-                          ros::NodeHandle nh_priv,
-                          std::string node_name,
-                          std::vector<double> args) :
+                           ros::NodeHandle nh_priv,
```
ayrton04: Er, not sure what happened with this commit, but

(a) it didn't address [my previous comment](https://github.com/cra-ros-pkg/robot_localization/pull/553/commits/520af0707f3f52ef6440c536dab7cc4637f47f2a#r427837341) 
(b) it added a bunch of unnecessary spaces

ayrton04:
There. Please leave this as-is until the build farm passes. Then I'll merge. Thanks!

florianspy:
> There. Please leave this as-is until the build farm passes. Then I'll merge. Thanks!

Sorry i had some difficulties with your parser checks. They always showed me wrong linesnumbers regarding errors in my format is, now i finally finished the build without errors.
I was trying to fix the "comma"-problem  and didn't read your comment that's why i experimented in wrong lines and caused all this confusion

PR #556, opened 2020-04-21
Fixing dual_ekf_navsat_example.yaml parameter file for dashing branch

pankhurivanjani:
While working with [dual_ekf_navsat_example.yaml](https://github.com/cra-ros-pkg/robot_localization/blob/dashing-devel/params/dual_ekf_navsat_example.yaml) parameter file I came across 3 major issues and errors because of which nodes kept dying :

1. 'expected [double] got [integer]' 
    I converted frequency data into double as suggested in [ros-answer](https://answers.ros.org/question/349995/issue-with-integrating-imu-encoder-odomtery-with-gps-for-localization-using-robot-localization-package-in-ros2/) and [Fix for 'expected [double] got [integer]' ](https://github.com/cra-ros-pkg/robot_localization/pull/547)

2.  Sequence should be of same type. Value type 'integer' do not belong at line_num
     I converted covariance matrix data into double type for this fix.

3.  Error due to missing ros__parameters: under nodes in yaml file. 

This updated file works well with [dual_ekf_navsat_example.launch.py](https://github.com/pankhurivanjani/ARL_Work_Dec19-june20/blob/master/scripts/dual_ekf_navsat_example.launch.py) that I created inspired from similar ros1 launch file after fixes.


SteveMacenski (approved):

SteveMacenski:
Awesome, thanks for the update.

PR #557, opened 2020-04-22
[melodic] Fix Windows build break.

seanyen:
* Conditionally guard the GCC\CLANG flags.
* Use `CMAKE_CXX_STANDARD` to set C++14 policy.
* Make the install location more friendly for Windows platform.
   * https://docs.ros.org/api/catkin/html/howto/format1/building_libraries.html#installing
   * https://docs.ros.org/api/catkin/html/howto/format1/building_executables.html#installing
* Avoid the `ERROR` macro conflicts due to `Windows.h` gets included in the header inclusion chain.

Verified with https://aka.ms/ros Windows build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,20 +173,16 @@ target_link_libraries(navsat_transform_nodelet navsat_transform ${catkin_LIBRARI
 ## Mark executables and/or libraries for installation
 install(TARGETS
   ekf
-  ekf_localization_node
   ekf_localization_nodelet
   filter_base
   filter_utilities
   navsat_transform
-  navsat_transform_node
   navsat_transform_nodelet
   ros_filter
   ros_filter_utilities
   robot_localization_estimator
   ros_robot_localization_listener
-  robot_localization_listener_node
   ukf
-  ukf_localization_node
   ukf_localization_nodelet
   ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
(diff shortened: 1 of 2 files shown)


seanyen:
@SteveMacenski  This is ready for review and merge. Thanks!

SteveMacenski:
LGTM but I mostly leave ROS1 things to Tom. Might be better targeted for a Noetic branch so that there's no breakage on a new release.

seanyen:
Sounds good. @ayrton04 Let me know what do you think. Perhaps we can have a `noetic-devel` branch first?

SteveMacenski:
We'll see what he says first. I'd like to avoid extra branches to try to keep synced if possible.

ayrton04:
These changes look pretty tame. I think I'd be fine with it going into Melodic.

@SteveMacenski, I've wondered if there's a better model for branch management w.r.t. ROS distros for a while. Maybe we just have a `devel` branch, and all distros use that until there's an ABI breakage, or we need to EOL a distribution or something?

ayrton04 (approved):
Thank you!

SteveMacenski:
ROS1’s on its last legs so I’m not super concerned with its management. I think as long as the changes aren’t massive we can keep 1 branch for Kinetic / melodic / noetic. Reducing the maintenance overhead of LTS distro can’t be overstated IMO. Keeping multiple LTS distro branches in sync is a pain so using the same when possible is a big stress reliever.

ayrton04:
Works for me.

seanyen:
Thanks for the merge!

PR #559, opened 2020-04-25
fix finding headers for yaml-cpp when building foxy

ManuelMeraz:
Hi there,

I was building on two machines with your current eloquent branch on foxy (seems like eloquent is your most recent ros2 branch) and on one machine it seemed to compile just fine. On my other computer with the same exact setup it seems to find the yaml-cpp header files that were included by the yaml cpp vendor with foxy in source build. 

I changed your cmakelists to find the yaml-cpp package directly and add the yaml-cpp include directories to your current project include directories. 

I made the minimal changes, but ideally cmake code should be done using targets and utilize ament cmake for ros 2, but for now this fixes my issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,9 +30,7 @@ find_package(tf2_eigen REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(Eigen3 REQUIRED)
-
-find_package(PkgConfig REQUIRED)
-pkg_check_modules(YAML_CPP yaml-cpp)
+find_package(yaml-cpp REQUIRED)
 
 set(library_name rl_lib)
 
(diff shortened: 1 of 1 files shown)


SteveMacenski:
> yaml cpp vendor

Wouldn't the best solution be to change to using the vendor?

ManuelMeraz:
Yeah that was the problem, it wasn't finding the vendor. When sourcing the the setup bash script from ros2 it overrides the LD_LIBRARY_PATH so all the libraries from the ros2 installation are preferred. It seemed like it was linking to the library, but it was not seeing the header files.

ayrton04:
@SteveMacenski, can this be merged?

SteveMacenski:
I don't think this is correct, I think it should be libyaml_vendor or yaml_cpp_vendor. If my memory recalls, those vendors are necessary to be cross platform. This implementation won't work on other systems (though `yaml-cpp` is in the package.xml so I suppose this is would work for what it is).

I'm on the fence, I think if we're going to make an update here, we should fix the problem completely, not just a lateral shift

ayrton04:
@ManuelMeraz, what is the status of this PR? Would it be possible to do what @SteveMacenski suggests?

ManuelMeraz:
I'm currently not working on this PR. It was mainly a fix for my own system and decided to create a PR because I found it useful. Feel free to close it if you want, or fork it if you want to continue with it.

Issue #564, opened 2020-05-02
[ROS2] Replace publishers with msg references with unique_ptrs


SteveMacenski:
Right now, all the publishers use the message reference API for publishing information. If using intra-process components, that will incur a copy. For inter-process it adds jitter to the system. While for messages of this size it is probably not going to make a large system-level impact, it is better practice and may be good samples for future folks and 0.1% improvement for an hour's work is a pretty reasonable tradeoff. 

Ref: https://github.com/SteveMacenski/ros2_ouster_drivers/pull/41 on discussion and also how to change to the other API with `std::move`.

Sarath18:
I would like to take up this issue

SteveMacenski:
Go for it!

PR #565, opened 2020-05-06
[ROS2] Change publishers to publish unique pointers

Sarath18:
### Basic info
Info | |
-- | --
Ticket(s) this addresses | #564 
Primary OS tested on | Ubuntu 18.04
Distro tested on | Eloquent

I haven't made modifications to the test cases.

### DIFF
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -139,13 +139,13 @@ private:
    * @brief Converts the odometry data back to GPS and broadcasts it
    * @param[out] filtered_gps The NavSatFix message to prepare
    */
-  bool prepareFilteredGps(sensor_msgs::msg::NavSatFix & filtered_gps);
+  bool prepareFilteredGps(std::unique_ptr<sensor_msgs::msg::NavSatFix> & filtered_gps);
 
   /**
    * @brief Prepares the GPS odometry message before sending
    * @param[out] gps_odom The odometry message to prepare
    */
-  bool prepareGpsOdometry(nav_msgs::msg::Odometry & gps_odom);
+  bool prepareGpsOdometry(std::unique_ptr<nav_msgs::msg::Odometry> & gps_odom);
 
   /**
    * @brief Used for setting the GPS data that will be used to compute the
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -139,13 +139,13 @@ private:
    * @brief Converts the odometry data back to GPS and broadcasts it
    * @param[out] filtered_gps The NavSatFix message to prepare
    */
-  bool prepareFilteredGps(std::unique_ptr<sensor_msgs::msg::NavSatFix> & filtered_gps);
+  bool prepareFilteredGps(sensor_msgs::msg::NavSatFix * filtered_gps);
 
   /**
    * @brief Prepares the GPS odometry message before sending
    * @param[out] gps_odom The odometry message to prepare
    */
-  bool prepareGpsOdometry(std::unique_ptr<nav_msgs::msg::Odometry> & gps_odom);
+  bool prepareGpsOdometry(nav_msgs::msg::Odometry * gps_odom);
 
   /**
    * @brief Used for setting the GPS data that will be used to compute the
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -48,6 +48,7 @@
 #include <string>
 #include <vector>
 #include <memory>
+#include <utility>
 
 using std::placeholders::_1;
 using std::placeholders::_2;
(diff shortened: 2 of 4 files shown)


include/robot_localization/navsat_transform.hpp
```diff
@@ -139,13 +139,13 @@ class NavSatTransform : public rclcpp::Node
    * @brief Converts the odometry data back to GPS and broadcasts it
    * @param[out] filtered_gps The NavSatFix message to prepare
    */
-  bool prepareFilteredGps(sensor_msgs::msg::NavSatFix & filtered_gps);
+  bool prepareFilteredGps(std::unique_ptr<sensor_msgs::msg::NavSatFix> & filtered_gps);
```
SteveMacenski: The signatures of these function should probably use raw pointers `sensor_msgs::msg::NavSatFix *` and then the calls for them would be `prepareFilteredGps(msg.get())`. This is just is smart pointer convention.

SteveMacenski (changes requested):
Just need to change function signatures. Also, please backport this to dashing.

Sarath18:
I'll backport this to dashing as well.

src/navsat_transform.cpp
```diff
@@ -179,14 +179,14 @@ void NavSatTransform::transformCallback()
       imu_sub_.reset();
     }
   } else {
-    std::unique_ptr<nav_msgs::msg::Odometry> gps_odom;
-    if (prepareGpsOdometry(gps_odom)) {
+    auto gps_odom = std::unique_ptr<nav_msgs::msg::Odometry>();
```
SteveMacenski: If you're going to make it here, it should be make_unique not unique_ptr.

src/navsat_transform.cpp
```diff
@@ -686,20 +687,20 @@ bool NavSatTransform::prepareFilteredGps(
...
 }
 
-bool NavSatTransform::prepareGpsOdometry(nav_msgs::msg::Odometry & gps_odom)
+bool NavSatTransform::prepareGpsOdometry(nav_msgs::msg::Odometry * gps_odom)
 {
   bool new_data = false;
 
   if (transform_good_ && gps_updated_ && odom_updated_) {
-    gps_odom = utmToMap(latest_utm_pose_);
+    *gps_odom = std::move(utmToMap(latest_utm_pose_));
```
SteveMacenski: Please make sure this has the intended consequences. Its very hard to tell if this will work without extensive coverage of unit testing on this front. Can you make sure that this message is being properly populated?

SteveMacenski (changes requested):
Make sure to use `make_unique` when making smart pointers. Looks much better.

src/navsat_transform.cpp
Sarath18: Ya my bad. I missed this one when converting unique_ptr to make_unique.

SteveMacenski: No worries, I read that a few times and kept thinking "what's wrong here... this looks right... oh yea".

src/navsat_transform.cpp
Sarath18: I passed all the unit tests but I do understand your concerns. Just in this one case is it better to pass unique_ptr to this function? Because i have learnt returning smart pointers is not a good practice.

SteveMacenski: I'm wondering why not `*gps_odom = utmToMap(latest_utm_pose_);`? The std::move I think removes the memory from latest_utm_pose_ and navsat transform has alot of moving parts happening asynchronously. I'm not 100% rock-solid confident that that won't have other effects. 

I guess can you talk about why you did that here and maybe that will help me understand

include/robot_localization/navsat_transform.hpp
```diff
@@ -139,13 +139,13 @@ class NavSatTransform : public rclcpp::Node
    * @brief Converts the odometry data back to GPS and broadcasts it
    * @param[out] filtered_gps The NavSatFix message to prepare
    */
-  bool prepareFilteredGps(sensor_msgs::msg::NavSatFix & filtered_gps);
+  bool prepareFilteredGps(sensor_msgs::msg::NavSatFix * filtered_gps);
```
ayrton04, line 142: Just out of curiosity, and because I'm being too lazy to go look, is this what ROS2 suggests for style? The data type is `sensor_msgs::msg::NavSatFix*`, so the extra space in there - whether for reference or pointer - feels odd.

src/navsat_transform.cpp
ayrton04: I think move semantics dictate that objects created in methods automatically get moved when returned. `*gps_odom = utmToMap(latest_utm_pose_);` should work fine, I think.

ayrton04 (approved):

ayrton04:
(Approval is subject to comments being addressed).

include/robot_localization/navsat_transform.hpp
Sarath18: Yes. Used ament uncrustify linter to adhere to ROS2 code style standards.

src/navsat_transform.cpp
Sarath18: > I think move semantics dictate that objects created in methods automatically get moved when returned.

Correct me if I am wrong. `utmToMap(latest_utm_pose_)` is an rvalue and I wanted to move the temporary data (scope limited to this function) created to the pointer.

>  Navsat transform has alot of moving parts happening asynchronously.

True. Not a 100% sure but this should work fine. I'll change this to 
`*gps_odom = utmToMap(latest_utm_pose_)` just to be one the safe side.

include/robot_localization/navsat_transform.hpp
SteveMacenski: Correct. I'm also not a fan, I like having the `*` on the type to clearly show that it is a pointer of this type.

SteveMacenski (approved):

Issue #550, opened 2020-03-30
navsat_transform documentation diagram


mabelzhang:
#### Feature description

We were trying to use `navsat_transformation` with `ekf_node`s for localization, and we had a lot of trouble figuring out how the nodes should be set up from the existing documentation. We have been following [this answer from Tom](https://answers.ros.org/question/202117/integrate-a-gps-sensor-with-robot_localization-and-use-move_base-node/) on ROS Answers, and the circular dependency between the second EKF node and the navsat_transform node didn't quite make sense to us.

So we thought a clear diagram in the documentation would help future users set it up easier.

#### Implementation considerations

@clalancette came up with the following diagram of the setup. We're thinking it would be good to add something like it to [the navsat_transform_node documentation page](http://docs.ros.org/indigo/api/robot_localization/html/navsat_transform_node.html)

![https://ibb.co/jZhYj8c](https://i.ibb.co/CsBySwk/navsat-transform.png)

Could we get some feedback as to whether this is the correct understanding of how these nodes are supposed to work together?

If the maintainers think this is useful to have in the documentation, let us know what requirements there are for diagrams / images, and I can submit a PR after we get this right.

SteveMacenski:
Admittedly, I haven't used navsat transform before nor have I used R_L for global fusion. I just spent the last 45 minutes reading the source code to understand and demo what is happening. Frankly, this is something I should have done a long time ago so your question nudged me to figure it out. Way to help me procrastinate figuring out my optimization problem issues :wink:. I've come up with the following, of which the most important parts are:
- The robot is happy, so it is smiling
- The GPS receiver is really good to work under an umbrella in the rain

![image](https://user-images.githubusercontent.com/14944147/77974455-0d38f780-72ac-11ea-96d8-cb86f7aa65f2.png)

I'm going to assume you understand the left hand side with the usual `odom->base` transform and inputs. 

[...]

PR #570, opened 2020-05-16
navsat_transform diagram to address #550

mabelzhang:
I finally read and understood everything in #550 and created a diagram with all the points taken.

Modifications from the original diagram in #550 (other than formatting):
- `local_odometry` edge should not go from local EKF to global EKF, as Tim Clephas mentioned in the issue ticket
- Instead, global EKF takes IMU and wheel odometry as input, just like the local EKF. This makes the two EKFs have the exact same workflow, the only difference being that the global EKF exchanges information with `navsat_transform`. This is consistent with Swift Nav's diagram too.
- Changed local EKF's output `local_odometry` to `/odometry/filtered/local`, and global EKF's output to `/odometry/filtered/global`, to be uniform and clear. Also to align with EKF node's default output name.

I added the source TeX file I used to generate the image, in case people need to modify it in the future. I don't know if anybody wants that in here. Let me know if I should throw it out.

I stuck the image in the Integrating with GPS tutorial. Seems the right place.

@clalancette


ayrton04:
I haven't forgotten about this. I have a major deadline this week, but will review it early next week. Thanks!

mabelzhang:
Thank you Tom. I had waited long enough myself to make the PR that waiting for the review is fair :)

ayrton04 (approved):
The new diagram looks great! I think this will save considerable time for people trying to understand the setup. Thanks again for this!

ayrton04:
@SteveMacenski, if you agree, then feel free to merge.

PR #575, opened 2020-06-04
Local Cartesian Option

AGummyBear:
This PR adds a feature to `navsat_transform` that allows you to optionally use a Local Cartesian (also known as a [Local Tangent Plane](https://en.wikipedia.org/wiki/Local_tangent_plane_coordinates)) coordinate system rather than the UTM coordinate system internally. This feature is useful for robots that operate within several kilometers of their origin location, but may need to operate close to or on top of a UTM boundary. Currently, `navsat_transform` does not handle crossing UTM boundaries.

The Local Cartesian transform is handled by GeographicLib: [Local Cartesian](https://geographiclib.sourceforge.io/html/classGeographicLib_1_1LocalCartesian.html)

Changes in this PR:
- Geographic lib is added as a build dependecy to the `robot_localization` package
- The parameter `use_local_cartesian` is added to `navsat_transform`
- Internal member variables of `navsat_transform` are re-named from "utm" to "cartesian" to more accurately reflect the data they hold

This feature has been tested on our robotic lawnmower across the UTM boundary between zones 16 and 17 at -84.0 degrees longitude.

### DIFF
--- a/include/robot_localization/navsat_transform.h
+++ b/include/robot_localization/navsat_transform.h
@@ -50,10 +50,10 @@
 
 #include <Eigen/Dense>
 
-#include <string>
-
 #include <GeographicLib/Geocentric.hpp>
 #include <GeographicLib/LocalCartesian.hpp>
+
+#include <string>
 
 namespace RobotLocalization
 {
@@ -94,8 +94,8 @@ class NavSatTransform
     //! centroid and returns the cartesian-frame pose of said centroid.
     //!
     void getRobotOriginCartesianPose(const tf2::Transform &gps_cartesian_pose,
-                               tf2::Transform &robot_cartesian_pose,
-                               const ros::Time &transform_time);
+                                     tf2::Transform &robot_cartesian_pose,
+                                     const ros::Time &transform_time);
 
     //! @brief Given the pose of the navsat sensor in the world frame, removes the offset from the vehicle's centroid
     //! and returns the world-frame pose of said centroid.
--- a/include/robot_localization/navsat_transform.h
+++ b/include/robot_localization/navsat_transform.h
@@ -52,6 +52,9 @@
 
 #include <string>
 
+#include <GeographicLib/Geocentric.hpp>
+#include <GeographicLib/LocalCartesian.hpp>
+
 namespace RobotLocalization
 {
 
@@ -87,11 +90,11 @@ class NavSatTransform
     //!
     bool fromLLCallback(robot_localization::FromLL::Request& request, robot_localization::FromLL::Response& response);
 
-    //! @brief Given the pose of the navsat sensor in the UTM frame, removes the offset from the vehicle's centroid
-    //! and returns the UTM-frame pose of said centroid.
+    //! @brief Given the pose of the navsat sensor in the cartesian frame, removes the offset from the vehicle's centroid
+    //! and returns the cartesian-frame pose of said centroid.
     //!
-    void getRobotOriginUtmPose(const tf2::Transform &gps_utm_pose,
-                               tf2::Transform &robot_utm_pose,
+    void getRobotOriginCartesianPose(const tf2::Transform &gps_cartesian_pose,
+                               tf2::Transform &robot_cartesian_pose,
                                const ros::Time &transform_time);
 
     //! @brief Given the pose of the navsat sensor in the world frame, removes the offset from the vehicle's centroid
@@ -137,22 +140,22 @@ class NavSatTransform
     void setTransformOdometry(const nav_msgs::OdometryConstPtr& msg);
 
     //! @brief Transforms the passed in pose from utm to map frame
-    //! @param[in] utm_pose the pose in utm frame to use to transform
+    //! @param[in] cartesian_pose the pose in cartesian frame to use to transform
     //!
-    nav_msgs::Odometry utmToMap(const tf2::Transform& utm_pose) const;
+    nav_msgs::Odometry cartesianToMap(const tf2::Transform& cartesian_pose) const;
 
     //! @brief Transforms the passed in point from map frame to lat/long
     //! @param[in] point the point in map frame to use to transform
     //!
     void mapToLL(const tf2::Vector3& point, double& latitude, double& longitude, double& altitude) const;
 
-    //! @brief Whether or not we broadcast the UTM transform
+    //! @brief Whether or not we broadcast the cartesian transform
     //!
-    bool broadcast_utm_transform_;
+    bool broadcast_cartesian_transform_;
 
-    //! @brief Whether to broadcast the UTM transform as parent frame, default as child
+    //! @brief Whether to broadcast the cartesian transform as parent frame, default as child
     //!
-    bool broadcast_utm_transform_as_parent_frame_;
+    bool broadcast_cartesian_transform_as_parent_frame_;
 
     //! @brief Whether or not we have new GPS data
     //!
@@ -194,6 +197,14 @@ class NavSatTransform
     //! @brief Whether we get the transform's yaw from the odometry or IMU source
     //!
     bool use_odometry_yaw_;
+
+    //! @brief Whether we use a Local Cartesian (tangent plane ENU) or the UTM coordinates as our cartesian
+    //!
+    bool use_local_cartesian_;
+
+    //! @brief Local Cartesian projection around gps origin
+    //!
+    GeographicLib::LocalCartesian gps_local_cartesian_;
 
     //! @brief Whether or not to report 0 altitude
     //!
@@ -244,9 +255,9 @@ class NavSatTransform
     //!
     Eigen::MatrixXd latest_odom_covariance_;
 
-    //! @brief Covariance for most recent GPS/UTM data
+    //! @brief Covariance for most recent GPS/UTM/LocalCartesian data
     //!
-    Eigen::MatrixXd latest_utm_covariance_;
+    Eigen::MatrixXd latest_cartesian_covariance_;
 
     //! @brief Timestamp of the latest good GPS message
     //!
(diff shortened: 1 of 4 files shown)


SteveMacenski:
Is there a ticket that this is in response to?

SteveMacenski:
I can't speak to the functionality, but from a code perspective, I have no issue with this.

AGummyBear:
> Is there a ticket that this is in response to?

This is related to, but does not directly address #222

SteveMacenski:
Can you clarify the differences? this is both for me and Tom, but also for posterity of people looking at this PR. A full description and important context is helpful.  

> Greenzie

Oh cool, I know of you guys. Hope all is well on that front.

AGummyBear:
Sure. There has been some discussion already on ticket #222, but here's the gist of it:

The `navsat_transform` node within this package has the purpose of converting global GPS coordinates (in Latitude/Longitude) to a relative coordinate within the robot's world frame. This relative coordinate, and the robot's world frame, are both Cartesian coordinate systems. The UTM grid is a widely-known Cartesian representation of the Earth's surface, so it is useful for handling the conversion from Lat/Lon to Cartesian, and that is what `navsat_transform` currently uses internally. The problem is that because the Earth's surface is not flat, the UTM grid actually approximates the ellipsoid of the Earth using multiple flat zones, and each zone has its own coordinates and a unique origin. What this means is that coordinates in Zone 16 do not have a Cartesian relationship with coordinates in Zone 17. So if your robot were to cross a boundary between two UTM zones, `navsat_transform` is not able to handle the new conversion, since it's not a trivial relationship.

So UTM has issues that make it difficult to use `navsat_transform` near a zone boundary, but UTM is not the only Cartesian coordinate system for the Earth (it's just a very widely known and well-defined one). Because the Earth is so large, you can actually create a flat tangent-plane approximation centered at any point on the surface and use Cartesian math between coordinates. Unless you move a substantial distance (think tens to hundreds of km) from the origin of this Cartesian frame, your calculations remain accurate: moving 1m in the positive-Y direction is equivalent to moving 1m North. The change I've made for this PR allows you to select Local Cartesian as an alternative to UTM for internal coordinate conversions.

My change does not make `navsat_transform` robust to handling UTM boundaries, but it does provide a workaround that many people might find useful. Specifically,

- If your robot is operated at varied locations and may be near a UTM boundary
- If your robot will not travel over long distances from its origin location (the location at which it acquires a good GPS lock)
- If you do not use the option `broadcast_utm_transform` and specifically need it to be the UTM transform.
 
Then I recommend setting `use_local_cartesian=true`

For people whose robots typically operate in a few locations, and they know that these locations are not on a UTM boundary, it shouldn't make a difference whether they use Local Cartesian or UTM internally.

SteveMacenski:
Awesome, makes sense, thanks for the overview. @ayrton04 should be around in the next few days (probably) to review but nothing sticks out to me as long as the transforms are correct.

ayrton04:
Thank you very much for this PR. I will review it next week. In the meantime, the build farm is unhappy because of some linting:

http://build.ros.org/job/Mpr__robot_localization__ubuntu_bionic_amd64/37/testReport/roslint/LintCheck/roslint_robot_localization/

AGummyBear:
Oops, I missed that one in my earlier fix. I ran the linter locally and it seems to be happy now.

include/robot_localization/navsat_transform.h
```diff
@@ -87,11 +90,11 @@ class NavSatTransform
...
 
-    //! @brief Given the pose of the navsat sensor in the UTM frame, removes the offset from the vehicle's centroid
-    //! and returns the UTM-frame pose of said centroid.
+    //! @brief Given the pose of the navsat sensor in the cartesian frame, removes the offset from the vehicle's
+    //! centroid and returns the cartesian-frame pose of said centroid.
     //!
-    void getRobotOriginUtmPose(const tf2::Transform &gps_utm_pose,
-                               tf2::Transform &robot_utm_pose,
+    void getRobotOriginCartesianPose(const tf2::Transform &gps_cartesian_pose,
+                               tf2::Transform &robot_cartesian_pose,
```
ayrton04: With the method name change, the second and third parameters are now misaligned. Would you mind updating that?

package.xml
```diff
@@ -37,6 +37,7 @@
   <build_depend>message_generation</build_depend>
   <build_depend>python-catkin-pkg</build_depend>
   <build_depend>roslint</build_depend>
+  <build_depend>geographiclib</build_depend>
```
ayrton04: I'd like to retain the alphabetical order of these, if possible.

src/navsat_transform.cpp
```diff
@@ -79,12 +80,13 @@ namespace RobotLocalization
     // Load the parameters we need
     nh_priv.getParam("magnetic_declination_radians", magnetic_declination_);
     nh_priv.param("yaw_offset", yaw_offset_, 0.0);
-    nh_priv.param("broadcast_utm_transform", broadcast_utm_transform_, false);
-    nh_priv.param("broadcast_utm_transform_as_parent_frame", broadcast_utm_transform_as_parent_frame_, false);
+    nh_priv.param("broadcast_utm_transform", broadcast_cartesian_transform_, false);
```
ayrton04: Let's add two new parameters for `broadcast_cartesian_transform` and `broadcast_cartesian_transform_as_parent_frame` as well. Maybe check if the user set `broadcast_utm_transform` and issue a deprecation warning?

include/robot_localization/navsat_transform.h
```diff
@@ -52,6 +52,9 @@
 
 #include <string>
 
+#include <GeographicLib/Geocentric.hpp>
```
ayrton04, line 53: Nit: can we drop these headers before `<string>`? I like to keep the STL headers last.

src/navsat_transform.cpp
```diff
@@ -273,34 +276,38 @@ namespace RobotLocalization
...
-        utm_transform_stamped.header.frame_id = (broadcast_utm_transform_as_parent_frame_ ? "utm" : world_frame_id_);
-        utm_transform_stamped.child_frame_id = (broadcast_utm_transform_as_parent_frame_ ? world_frame_id_ : "utm");
-        utm_transform_stamped.transform = (broadcast_utm_transform_as_parent_frame_ ?
-                                             tf2::toMsg(utm_world_trans_inverse_) : tf2::toMsg(utm_world_transform_));
-        utm_transform_stamped.transform.translation.z = (zero_altitude_ ?
-                                                           0.0 : utm_transform_stamped.transform.translation.z);
-        utm_broadcaster_.sendTransform(utm_transform_stamped);
+        geometry_msgs::TransformStamped cartesian_transform_stamped;
+        cartesian_transform_stamped.header.stamp = ros::Time::now();
+        std::string cartesian_frame_id = (use_local_cartesian_)? "local_enu" : "utm";
```
ayrton04: The parentheses here are a little off. To retain the original style, we should do this:

`std::string cartesian_frame_id = (use_local_cartesian_? "local_enu" : "utm");`

I realize the parentheses are unnecessary, but IMO it makes these things more readable.

src/navsat_transform.cpp
```diff
@@ -386,56 +402,69 @@ namespace RobotLocalization
...
+                                 latitude,
+                                 longitude);
+      altitude = odom_as_cartesian.getOrigin().getZ();
+    }
   }
 
-  void NavSatTransform::getRobotOriginUtmPose(const tf2::Transform &gps_utm_pose,
-                                              tf2::Transform &robot_utm_pose,
+  void NavSatTransform::getRobotOriginCartesianPose(const tf2::Transform &gps_cartesian_pose,
+                                              tf2::Transform &robot_cartesian_pose,
```
ayrton04: Same comment re: parameter alignment.

src/navsat_transform.cpp
```diff
@@ -448,26 +477,26 @@ namespace RobotLocalization
...
       mat.getRPY(roll, pitch, yaw);
       yaw += (magnetic_declination_ + yaw_offset_ + utm_meridian_convergence_);
-      utm_orientation.setRPY(roll, pitch, yaw);
+      cartesian_orientation.setRPY(roll, pitch, yaw);
 
       // Rotate the GPS linear offset by the orientation
       // Zero out the orientation, because the GPS orientation is meaningless, and if it's non-zero, it will make the
-      // the computation of robot_utm_pose erroneous.
-      offset.setOrigin(tf2::quatRotate(utm_orientation, offset.getOrigin()));
+      // the computation of robot_c_pose erroneous.
```
ayrton04: *robot_cartesian_pose

ayrton04 (approved):
Very cool! Approving pending the changes I've requested. Thanks for this, and sorry for the delay in reviewing.

include/robot_localization/navsat_transform.h
AGummyBear: Done.

include/robot_localization/navsat_transform.h
AGummyBear: Done.

package.xml
AGummyBear: Done.

src/navsat_transform.cpp
AGummyBear: Done.

src/navsat_transform.cpp
AGummyBear: Done.

src/navsat_transform.cpp
AGummyBear: I made this change and tested it locally, but I would like a second pair of eyes on it. Basically, if the old parameter is set, it issues a warning telling the user to use the new parameter. The value of the old parameters is not used, so they will revert to the defaults (`false`) for any user once they receive this update.

src/navsat_transform.cpp
```diff
@@ -92,6 +94,18 @@ namespace RobotLocalization
     nh_priv.param("transform_timeout", transform_timeout, 0.0);
     transform_timeout_.fromSec(transform_timeout);
 
+    // Check for deprecated parameters
+    if (nh_priv.param("broadcast_utm_transform", unused, false))
+    {
+      ROS_WARN("navsat_transform, Parameter 'broadcast_utm_transform' has been deprecated. Please use"
```
ayrton04, line 99: I think we should maybe warn them, but then set the equivalent `broadcast_cartesian_transform` setting anyway. It gets a little more complicated if they do something silly like set `broadcast_cartesian_transform: false` and then `broadcast_utm_transform: true` (or vice-versa), but there's no reason for this to get too complex.

AGummyBear: Fair point. I've made the change to set the parameters when using the deprecated names.

ayrton04: Sorry to be a pain, but I think this will overwrite the value of `broadcast_cartesian_transform_` with _false_ if they _don't_ have `broadcast_utm_transform` set. Maybe change this to `getParam`?

```
if (nh_priv.getParam("broadcast_utm_transform", broadcast_cartesian_transform_))
```

AGummyBear: Yes, you are absolutely right. I changed that to `getParam()` and verified that it doesn't overwrite the values.

ayrton04 (approved):

ayrton04:
Thanks again for the PR!

Issue #524, opened 2020-01-19
navsat_transform: Parameter 'wait_for_datum' causes change to UTM transform. 


Karl60:
## Bug report
Selecting 'true' or 'false' for the navsat_transform_node parameter "wait_for_datum" changes the published utm transform between odom->utm and map->utm .

**Required Info:**

- Operating System:
  - Ubuntu 18.04

- Installation type: 
  - binaries
- Version or commit hash:
  - 

#### Steps to reproduce issue

- navsat_transform node config: 
```
    # Odometry input with frame_id: map. (From the ekf_se_map node, world_frame: map)
    broadcast_utm_transform: true
    broadcast_utm_transform_as_parent_frame: false
    datum: [56.00000, -3.00000, 0.0]
```
 - Change parameter "wait_for_datum" between True and False.  

#### Expected behavior:
UTM transform created from world-frame of subscribed odometry topic, ie map->utm.

#### Actual behavior
```
wait_for_datum: true  # we get an odom->utm transform.
wait_for_datum: false # we get a map->utm transform.
```

RonaldEnsing:
I am running into this as well.

ayrton04:
Try this?

https://github.com/cra-ros-pkg/robot_localization/pull/576

RonaldEnsing:
Yes, #576 works for me.

RonaldEnsing:
Will #576 be merged?

PR #576, opened 2020-07-14
Parameter for world_frame_id for navsat_transform

cdimmig:
Added parameter for world_frame_id to navsat transform to allow for setting frame ID to be something other than the default when wait_for_datum is true. Issue noted in #524

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -96,6 +96,7 @@ namespace RobotLocalization
     nh_priv.param("publish_filtered_gps", publish_gps_, false);
     nh_priv.param("use_odometry_yaw", use_odometry_yaw_, false);
     nh_priv.param("wait_for_datum", use_manual_datum_, false);
+    nh_priv.param("world_frame_id", world_frame_id_, std::string("odom")); // Set automatically if wait_for_datum = false
     nh_priv.param("frequency", frequency, 10.0);
     nh_priv.param("delay", delay, 0.0);
     nh_priv.param("transform_timeout", transform_timeout, 0.0);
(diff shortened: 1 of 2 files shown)


ayrton04:
So the real root of this issue is that, if we specify the datum manually, we aren't able to get an odometry message before the frame_ids get used in the datum callback. I think that this PR needs three more things:

1. Please also load the `base_link_frame` as a parameter.
2. Both the `world_frame` and `base_link_frame` parameters should only get loaded if `wait_for_datum` is *true*. If those parameters are set and `wait_for_datum` is *false*, then we should warn the user that the parameters have no effect.
3. If `wait_for_datum` is true, and we load the frame parameters, but we later [get an odometry message](https://github.com/cra-ros-pkg/robot_localization/blob/2059153b39f534013490ebf2ae2788dd6a82d7e9/src/navsat_transform.cpp#L560) and the frame_ids in that message don't match the ones in the parameters we set, we should issue a warning (once).

cdimmig:
Updated with @ayrton04 's comments!

src/navsat_transform.cpp
```diff
@@ -557,6 +586,18 @@ namespace RobotLocalization
 
   void NavSatTransform::odomCallback(const nav_msgs::OdometryConstPtr& msg)
   {
+    if (world_frame_param_loaded_ && (world_frame_id_.compare(msg->header.frame_id) != 0))
```
ayrton04: `compare` will obviously work, but I think this would be more readable as `world_frame_id_ != msg->header.frame_id`. Same below with the `base_link_frame_id_`.

ayrton04 (approved):
Just a small change, but I'll merge after that gets in. Thanks! Sorry for the delay.

cdimmig:
@ayrton04 updated!

ayrton04:
Thanks again!

PR #579, opened 2020-07-20
adding boost dep

SteveMacenski:


SteveMacenski:
@ayrton04 FYI RL was really broken before these changes. I'm honestly surprised if the EKF/UKF nodes worked at all in Foxy / Eloquent before this... The install paths to the libraries and executives were horribly wrong and we lacked some critical dependencies to build 

(literally every single test failed from not being able to locate the executables or libraries)

Fixing one more issue with cppcheck..

ayrton04:
Wowza, well done. Maybe leave the cppcheck stuff. I'll take that as a "get my ROS2 environment dusted off and functional" task.

Issue #325, opened 2016-11-17
Rare issue caused by navsat_transform and bad IMU data


mattforestyoung:
I've identified a rare issue that I think was caused by navsat_transform and an especially bad IMU reading. It's only happened once to my knowledge so this issue can take very low priority. However I wanted to mention it here because it could lead to a simple enhancement for _navsat_transform_. 

I'm using a Jackal UGV in a sloped, outdoor area and localizing it using the _robot_localization_ package. Normally each data set is consistent, however one data set sticks out. The following chart shows the profile of the slope, both datasets are taken directly from the _/odometry/gps_ published by _navsat_transform_. 

![comparison of z2 profiles](https://cloud.githubusercontent.com/assets/17462459/20374903/647fba0a-ace0-11e6-97e2-2a29cbc01447.JPG)

The red trendline shows the correct cross-sectional profile of the slope. Every dataset except one agrees with this. The blue dataset is wrong, the raw GPS data matches the normal slope but the data in _/odometry/gps_ does not.  

To give another view of the same data, this is what _/odometry/gps_ looks like when its displayed in RViz for the correct slope profile:

![20161115_normal_z2_profile _cropped](https://cloud.githubusercontent.com/assets/17462459/20406264/80ed5512-ad71-11e6-8759-8f80d6300d51.jpg)

And this is the same view of the distorted dataset:

![20161027_distorted_z2_alt_cropped](https://cloud.githubusercontent.com/assets/17462459/20406278/92fff890-ad71-11e6-9f66-e43e10dbe2f4.jpg)

[...]

PR #582, opened 2020-07-23
fix issue caused by starting on uneven terrain

jrbaxter0:
This resolves an bug in `navsat_transform` generating the static transform between `map` and `utm` while on a slope. It does this by stripping the pitch and roll from the odometry message. This is noted in issue #325. Essentially, since the pitch and roll are stripped from the `utm` orientation, the same must be done for the `map` orientation.

`cartesian_world_transform_`, which is the transform from utm to map is calculated in the following way:
<a href="https://www.codecogs.com/eqnedit.php?latex=T_{utm}^{map}&space;=&space;T_{base\_link}^{map}&space;*&space;(T_{base\_link\_yaw\_only}^{utm})^{-1}" target="_blank"></a>
<a href="https://www.codecogs.com/eqnedit.php?latex=T_{utm}^{map}&space;=&space;T_{base\_link}^{map}&space;*&space;T_{utm}^{base\_link\_yaw\_only}" target="_blank"></a>

This only works if `base_link` and `base_link_yaw_only` are the same frame. On a slope, this won't be true, which is why the bug appears there.

These images show the output of navsat_transform under several configurations. In each, a sinusoidal altitude is given on the input gps/fix message, while latitude and longitude are held fixed.

This is the normal output of the system when the input imu and pose messages have an identity quaternion. Note that the X and Y positions are fixed, while Z tracks altitude.
normal

This is the output of the system before the patch, where the input imu and pose messages have a 45 degree upwards pitch. Note that X is no longer fixed. The static transform between utm and map has relative pitch, which is causing this issue.

Remember that [REP 105](https://www.ros.org/reps/rep-0105.html) specifies that map should have Z pointing straight upwards. Therefore, there can only be relative yaw between utm and map. 
broken

This is the output of the system after the patch, where the input imu and pose messages have a 45 degree upwards pitch. Note that X is once again fixed and Z tracks altitude.
fixed

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -296,7 +296,16 @@ namespace RobotLocalization
       cartesian_pose_with_orientation.setOrigin(transform_cartesian_pose_corrected.getOrigin());
       cartesian_pose_with_orientation.setRotation(imu_quat);
 
-      cartesian_world_transform_.mult(transform_world_pose_, cartesian_pose_with_orientation.inverse());
+      // Remove roll and pitch from odometry pose
+      // Must be done because roll and pitch is removed from cartesian_pose_with_orientation
+      double odom_roll, odom_pitch, odom_yaw;
+      tf2::Matrix3x3(transform_world_pose_.getRotation()).getRPY(odom_roll, odom_pitch, odom_yaw);
+      tf2::Quaternion odom_quat;
+      odom_quat.setRPY(0.0, 0.0, odom_yaw);
+      tf2::Transform transform_world_pose_yaw_only(transform_world_pose_);
+      transform_world_pose_yaw_only.setRotation(odom_quat);
+
+      cartesian_world_transform_.mult(transform_world_pose_yaw_only, cartesian_pose_with_orientation.inverse());
 
       cartesian_world_trans_inverse_ = cartesian_world_transform_.inverse();
 


ayrton04:
Just a note that I am not ignoring this. I am swamped at the moment, and will review soon. Thanks for submitting!

jrbaxter0:
@ayrton04 any update on this?

ayrton04 (approved):
Sorry, @jrbaxter0. Thanks for the thorough PR with visualizations. This has been on my hit list for a very long time, but I've just never found the time (as you can see!). Anyway, this LGTM. @SteveMacenski, if you have no objections, I'll merge this later today or tomorrow.

SteveMacenski:
No objections

ayrton04:
Thanks again!

Timple:
Would it be possible to trigger a new bloom release which includes this fix? :slightly_smiling_face:

ayrton04:
Yeah, let me get the branches in sync and I'll cut the releases.

Timple:
Great, thanks!

Do you expect to finish this in the next two days?
Reason for asking is the end-of-year sync: https://discourse.ros.org/t/request-for-melodic-sync/17972

ayrton04:
Yikes, sorry! I'll do it now.

ayrton04:
https://github.com/ros/rosdistro/pull/27783

PR #583, opened 2020-08-01
[ros2] trim boost dependencies

mikaelarguedas:
In an [ongoing effort](https://discourse.ros.org/t/generating-dev-and-runtime-artefacts-from-ros-packages/12448) to reduce the footprint on the target systems, it is preferred to declare specific dependencies instead of generic ones.

This PR gets rid of the runtime dependency on boost completely and trims down the build dependencies to only the boost headers and no the boost libraries, the part of boost being used in this package being a header only lib.

---

@SteveMacenski I see #579 added a build, build_export and exec dependency on *all* boost libraries (~800MB) but the CMakeLists doesnt make use of the Boost headers or libraries for any of its targets. What are the steps I should take to confirm this PR doesnt reintroduce the problems #579 was fixing ?
Currently I only confirmed that, in a fresh environment, I can use rosdep to install the deps of the package and then the package builds successfully and all the tests pass.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,7 @@ include_directories(SYSTEM ${Eigen_INCLUDE_DIRS})
 include_directories(
   include
   ${EIGEN3_INCLUDE_DIRS}
+  ${Boost_INCLUDE_DIRS}
 )
 
 add_library(
--- a/package.xml
+++ b/package.xml
@@ -17,6 +17,9 @@
   <buildtool_depend>builtin_interfaces</buildtool_depend>
   <buildtool_depend>rosidl_default_generators</buildtool_depend>
 
+  <build_depend>libboost-dev</build_depend>
+  <build_export_depend>libboost-dev</build_export_depend>
+
   <depend>eigen</depend>
   <depend>geographic_msgs</depend>
   <depend>geometry_msgs</depend>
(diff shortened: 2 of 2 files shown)


SteveMacenski:
Good question, that was done because the build farm was failing to release binaries because of the missing dependency (and various other warnings / errors that PR resolved). This should probably be OK. I tend to use `depend` general tag as I'm not generally concerned about image sizes (and I generally don't use boost when I have an option). Happy to help accommodate for users that do need to keep things small.

SteveMacenski (approved):

mikaelarguedas:
> I tend to use depend general tag as I'm not generally concerned about image sizes

Yeah this became pretty common in ROS 2 packages.
Even using those it's possible to avoid requiring to pull GB of unnecessary libraries by selecting the specific rosdep keys needed by a package (e.g. https://github.com/ros/class_loader/pull/136)

> and I generally don't use boost when I have an option

Mentioning it here because ROS 2 packages being mostly Boost-free, the impact of a package pulling every boost library is big (for target image size but also machine time in general) and out of the handful of packages that do use the `boost` key are `karto_sdk` and `slam_toolbox` :wink:

SteveMacenski:
>  Even using those it's possible to avoid requiring to pull GB of unnecessary libraries by selecting the specific rosdep keys needed by a package (e.g. ros/class_loader#136)

I wasn't aware of that off hand, good to know. My modified open karto library is extremely ingrained with boost that cannot be removed. We need the serialization module, there's no way around that. The thread and other system modules potentially could be removed.

mikaelarguedas:
> We need the serialization module, there's no way around that. The thread and other system modules potentially could be removed.

:+1:
I'm working my way through having all the rosdep keys available to be able to declare only the modules needed. It seems (didnt check very closely) that the only ones missing for karto and slam_toolbox are included in [this PR](https://github.com/ros/rosdistro/pull/25995)

PR #594, opened 2020-09-15
Fix bug in creating frequency diag.

### LINKED ISSUE #593: [ROS1] Diagnostic frequency monitor min/max values are 0.
https://github.com/cra-ros-pkg/robot_localization/blob/285a38614135f064ced8f2532530fca271eacd87/src/ros_filter.cpp#L1956

The min/max values input to the frequency monitor must be on heap (not stack). Pointers to those memory locations are used by the diagnostic monitor task and therefore can end up accessing stack memory (reads) when reading the min/max frequency thresholds.

tgreier:
Resolves #593 
Moved the min and max frequency values to the heap since pointers to those values are passed to diagnostic_updater::FrequencyStatusParam (to allow the values to be updated on the fly).

### DIFF
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -774,6 +774,16 @@ protected:
   //! @brief optional signaling diagnostic frequency
   //!
   std::unique_ptr<diagnostic_updater::HeaderlessTopicDiagnostic> freq_diag_;
+
+  //! @brief minimum frequency threshold for frequency diagnostic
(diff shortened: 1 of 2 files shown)


tgreier:
Does anyone know why these build tests are failing? I don't think it has to do with this pull request.

ayrton04:
That seems to happen with the ROS2 tests, and I'm not sure why. I don't actually have a test setup locally for ROS2 just now, so I'm not sure.

ayrton04:
Let me close and open to kick the job off again.

ayrton04 (approved):

Issue #574, opened 2020-06-04
TF_REPEATED_DATA warning in Noetic


ayrton04:
I'm running the package test suite in Noetic, and I am seeing this getting spammed:

```
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame base_link at time 1432235782.066014 according to authority unknown_publisher
         at line 278 in /tmp/binarydeb/ros-noetic-tf2-0.7.1/src/buffer_core.cpp
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame base_link at time 1432235782.433942 according to authority unknown_publisher
         at line 278 in /tmp/binarydeb/ros-noetic-tf2-0.7.1/src/buffer_core.cpp
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame base_link at time 1432235782.865701 according to authority unknown_publisher
         at line 278 in /tmp/binarydeb/ros-noetic-tf2-0.7.1/src/buffer_core.cpp
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame base_link at time 1432235782.965662 according to authority unknown_publisher
         at line 278 in /tmp/binarydeb/ros-noetic-tf2-0.7.1/src/buffer_core.cpp
```

This is clearly a new warning, and I'm assuming it's due to [this](https://github.com/cra-ros-pkg/robot_localization/issues/336#issuecomment-629725698) common situation:

- Cycle 1: receive measurement from sensor with time stamp `t1`. Fuse measurement. Publish state at time `t1`.
- Cycle 2: receive measurement from sensor with time stamp `t2` < `t1`. Rewind filter, insert measurement from `t2`, fuse measurement from `t2` and then measurement from `t1`. Publish state at time `t1`.

It looks like this helpful change was added, in which `tf2` would just overwrite the old transform with the new one:

https://github.com/ros/geometry2/commit/eefb50935bfd28223c87b3d708e916f7bcc4b4ca

However, that was later changed, such that attempting to broadcast the same transform at the same time would fail and spit out a warning:

https://github.com/ros/geometry2/commit/04625380bdff3f3e9e860fc0e85f71674ddd1587#diff-c25683741e1f7b4e50eb6c5cdcad9661R275

@SteveMacenski, what are your thoughts here? Should I open a ticket on the `geometry2` repo?

ayrton04:
Started a discussion here:

https://github.com/ros/geometry2/issues/467

[...]

PR #595, opened 2020-09-18
Making repeated state publication optional

ayrton04:
Addresses #574.

As far as how this will behave, if the filter publishes a state with time `t`, and we later get a measurement with stamp < `t`, we'll still integrate that measurement and correct our state internally, but we won't publish a state that includes that correction until the next publication cycle _after_ `t`.

This may get its base changed to a new `noetic-devel` branch.

### DIFF
--- a/doc/state_estimation_nodes.rst
+++ b/doc/state_estimation_nodes.rst
@@ -174,6 +174,10 @@ If *true*, the state estimation node will publish the transform from the frame s
 ^^^^^^^^^^^^^^^^^^^^^
 If *true*, the state estimation node will publish the linear acceleration state. Defaults to *false*.
 
+~permit_corrected_publication
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When the state estimation nodes publish the state at time `t`, but then receive a measurement with a timestamp < `t`, they re-publish the corrected state, with the same time stamp as the previous publication. Setting this parameter to *false* disables that behavior. Defaults to *false*.
+
 ~print_diagnostics
 ^^^^^^^^^^^^^^^^^^
 If true, the state estimation node will publish diagnostic messages to the ``/diagnostics`` topic. This is useful for debugging your configuration and sensor data.
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1922,7 +1922,9 @@ namespace RobotLocalization
       // If we're trying to publish with the same time stamp, it means that we had a measurement get inserted into the
       // filter history, and our state estimate was updated after it was already published. As of ROS Noetic, TF2 will
       // issue warnings whenever this occurs, so we make this behavior optional.
-      corrected_data = (!permitCorrectedPublication_ && lastPublishedStamp_ == filteredPosition.header.stamp);
+      // Just for safety, we also check for the condition where the last published stamp is *later* than this stamp.
+      // This should never happen, but we should handle the case anyway.
+      corrected_data = (!permitCorrectedPublication_ && lastPublishedStamp_ >= filteredPosition.header.stamp);
 
       // If the worldFrameId_ is the odomFrameId_ frame, then we can just send the transform. If the
       // worldFrameId_ is the mapFrameId_ frame, we'll have some work to do.
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -38,6 +38,7 @@
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
 #include <algorithm>
+#include <iostream>
 #include <map>
 #include <memory>
 #include <string>
@@ -823,6 +824,9 @@ namespace RobotLocalization
     // Whether we're publishing the acceleration state transform
     nhLocal_.param("publish_acceleration", publishAcceleration_, false);
 
+    // Whether we'll allow old measurements to cause a re-publication of the updated state
+    nhLocal_.param("permit_corrected_publication", permitCorrectedPublication_, false);
+
     // Transform future dating
     double offsetTmp;
     nhLocal_.param("transform_time_offset", offsetTmp, 0.0);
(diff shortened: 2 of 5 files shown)


src/ros_filter.cpp
```diff
@@ -991,20 +995,21 @@ namespace RobotLocalization
              "\ntransform_timeout is " << tfTimeout_.toSec() <<
              "\nfrequency is " << frequency_ <<
              "\nsensor_timeout is " << filter_.getSensorTimeout() <<
-             "\ntwo_d_mode is " << (twoDMode_ ? "true" : "false") <<
-             "\nsmooth_lagged_data is " << (smoothLaggedData_ ? "true" : "false") <<
+             "\ntwo_d_mode is " << std::boolalpha << twoDMode_ <<
```
ayrton04, line 998: Sorry, I printed out the new param here using `boolalpha`, and thought I might as well pick up the others.

src/ros_filter.cpp
```diff
@@ -1912,9 +1919,14 @@ namespace RobotLocalization
               " This was likely due to poorly coniditioned process, noise, or sensor covariances.");
       }
 
+      // If we're trying to publish with the same time stamp, it means that we had a measurement get inserted into the
+      // filter history, and our state estimate was updated after it was already published. As of ROS Noetic, TF2 will
+      // issue warnings whenever this occurs, so we make this behavior optional.
+      corrected_data = (!permitCorrectedPublication_ && lastPublishedStamp_ == filteredPosition.header.stamp);
```
SteveMacenski: `lastPublishedStamp_ == filteredPosition.header.stamp` I don't think that works, because what if its older than just the last update. If you have something with substantial latency (100ms) but your filter is spinning at 100hz, then it could be much older than last, right?

ayrton04: So I'm thinking aloud here. The filter history handles this situation:

- In the first cycle, the filter has four measurements to process. They have timestamps `1`, `2`, `4`, and `5`. They get processed, and we publish a state with timestamp `5`.
- In the second cycle, the filter receives only a measurement with timestamp `3`. It rewinds history to the posterior state after measurement `2` was fused, and then processes `3`, `4`, and `5` again. Then it publishes a state with timestamp `5`. This creates the issue in question, and is solved by this PR.

Now let's go over your scenario, or at least what I think you are saying.

- In the first cycle, the filter has four measurements to process. They have timestamps `1`, `2`, `4`, and `5`. They get processed, and we publish a state with timestamp `5`.
- In the second cycle, the filter has four measurements to process. They have timestamps `6`, `7`, `8`, and `9`. They get processed, and we publish a state with timestamp `9`.
- In the third cycle, the filter receives only a measurement with timestamp `3`. It rewinds history to the posterior state after measurement `2` was fused, and then processes `3`, `4`, `5`, `6`, `7`, `8`, and `9` again. Then it publishes a state with stamp `9` again. This also causes the issue, and is solved by this PR.

The filter was never publishing the updated state as it processed the history. It was rewinding, inserting, and processing all the way up to the latest measurement, then publishing. So I think checking the latest publication should work?

Note that in both of those cases, anyone listening to `tf` will get the exact same behavior before and after this change: the first transform published won't get updated when the second one arrives. It's just that in Noetic, we also get yelled at. :)

There is a third case.

- In the first cycle, the filter has four measurements to process. They have timestamps `1`, `2`, `4`, and `5`. They get processed, and we publish a state with timestamp `5`.
- In the second cycle, the filter has two measurements to process. They have timestamps `3` and `6`. The filter rewinds history to the posterior state after measurement `2` was fused, and then processes `3`, `4`, `5`, and `6`, and publishes the state with timestamp `6`. This is obviously not an issue.

Anyway, I will readily admit that I might be totally missing your point, and I apologize if so. Can you provide an example?

SteveMacenski:
Shouldn't this be targeted only onto noetic and foxy? I don't think this issue affects melodic

doc/state_estimation_nodes.rst
```diff
@@ -174,6 +174,10 @@ If *true*, the state estimation node will publish the transform from the frame s
 ^^^^^^^^^^^^^^^^^^^^^
 If *true*, the state estimation node will publish the linear acceleration state. Defaults to *false*.
 
+~permit_corrected_publication
```
SteveMacenski, line 177: Any "new in XYZ distro" marker we can do?

src/ros_filter.cpp
```diff
@@ -1974,7 +1986,13 @@ namespace RobotLocalization
...
 
       // Fire off the position and the transform
-      positionPub_.publish(filteredPosition);
+      if (!corrected_data)
+      {
+        positionPub_.publish(filteredPosition);
+      }
+
+      // Retain the last published stamp so we can detect repeated transforms in future cycles
+      lastPublishedStamp_ = filteredPosition.header.stamp;
```
SteveMacenski, line 1997: This should be in the if statement, or else you could update based on a `corrected_data` update.

src/ros_filter.cpp
SteveMacenski: OK, I wasn't sure where `filteredPosition`'s stamp was from, I didn't know that it would already be, at this point, rewinded back up to the latest message in all cases. That works. 

I might just make it a little more error proof with `lastPublishedStamp_ >= filteredPosition.header.stamp`. I agree it shouldn't matter, but I think it makes syntactically more clear why this is being checked (or why they might be equal). I feel like a native reader can pull more from that than `==`.

doc/state_estimation_nodes.rst
ayrton04: Should we just make a `noetic-devel` and then change this PR to target that? Then we don't have to worry about that.

src/ros_filter.cpp
ayrton04: Yeah, that situation should never happen, but it also won't hurt to change it, and I agree with the increased clarity.

src/ros_filter.cpp
ayrton04: So the earlier suggestion about >= notwithstanding, as of now, if `corrected_data == true`, then `lastPublishedStamp_ == filteredPosition.header.stamp` anyway. However, I'm perfectly happy to move this up to line 1992.

ayrton04:
> Shouldn't this be targeted only onto noetic and foxy? I don't think this issue affects melodic

Yes, but we don't have a `noetic-devel`. Are you cool if we make one now? I'll take care of it.

Will roll your suggestion into the PR tomorrow. Thanks!

doc/state_estimation_nodes.rst
SteveMacenski: Yeah

SteveMacenski:
Yeah sure

SteveMacenski (approved):

ayrton04:
@reobaird, were you able to try this out?

PR #602, opened 2020-10-11
Fix warning "Failed to meet update rate!"

### LINKED ISSUE #601: Failed to meet update rate!
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source from git repo
- Version or commit hash:
  - HEAD of melodic-devel branch

#### Description

We are running the ekf_localization_node on a Clearpath Jackal.

Commit edc5bea3 causes the performance of the filter to be very poor, and the node continuously logs the warning "Failed to meet update rate!" This message happens many times per second.

This commit causes an access to the ros parameter server for every IMU packet that arrives. On the Jackal, this message comes into the filter at 75 Hz, so the node is trying to access the parameter server every 13 milliseconds. In our environment, the ROS master and parameter server do not run locally on the robot. The robot communciates over a wireless link to a server, and this param() call sometimes takes > 50 milliseconds to execute. That causes input queue overflow and very erratic timing of the filter update callback.

One way to fix this is to cache the silent_tf_failure flag in the RosFilter class. I will submit a pull request shortly.

#### Steps to reproduce issue

See descrption.

#### Expected behavior

See description.

#### Actual behavior

See description.

mjs973:
This patch reads the parameter server key silent_tf_failure only once at startup. It resolves #601.

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -476,6 +476,10 @@ template<class T> class RosFilter
     //!
     bool useControl_;
 
+    //! @brief Whether or not to print warning for tf lookup failure
+    //!
+    bool silent_tf_failure_;
+
     //! @brief The max (worst) dynamic diagnostic level.
     //!
     int dynamicDiagErrorLevel_;
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -478,7 +478,7 @@ template<class T> class RosFilter
 
     //! @brief Whether or not to print warning for tf lookup failure
     //!
-    bool silent_tf_failure_;
+    bool silentTfFailure_;
 
     //! @brief The max (worst) dynamic diagnostic level.
     //!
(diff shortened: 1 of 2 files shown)


include/robot_localization/ros_filter.h
```diff
@@ -476,6 +476,10 @@ template<class T> class RosFilter
     //!
     bool useControl_;
 
+    //! @brief Whether or not to print warning for tf lookup failure
+    //!
+    bool silent_tf_failure_;
```
ayrton04: Thank you very much for both catching and addressing this issue.

Given that this is against `melodic-devel`, and we're still using the old (and technically incorrect) style for variable names in this branch, would you mind making this consistent with the others, and changing this to `silentTfFailure_`? I'll make sure it matches the style when I cherry pick it into the other branches. Thanks!

ayrton04 (approved):

include/robot_localization/ros_filter.h
mjs973: My branch has been updated with this change.

ayrton04 (approved):

ayrton04:
Thanks again!

PR #605, opened 2020-10-21
Fix compilation with latest ROS 2 source

jacobperron:
This PR adds two changes to fix building with the latest version of ROS 2 from source:

- Include tf2_ros/buffer.h where needed (4e4f54c)
After some cleanup in ros2/geometry2#325, the header is no longer being transitively included.
- Fix deprecated usage of FutureReturnCode (2891cbc)
The namespace rclcpp::executors:: was deprecated in Foxy and removed in ros2/rclcpp#1327.

I believe both changes are compatible with ROS Foxy.


jacobperron:
I'm not sure if the test failure is related. Seems like it could be a flake?

SteveMacenski:
toggling CI

jacobperron:
CI looks good now :slightly_smiling_face:

SteveMacenski (approved):

PR #611, opened 2020-11-16
Add example launch file for dual_ekf_navsat

jediofgever:
Not sure if it is intentional that there is no `launch.py` example for `dual_ekf_navsat` included in ros2 branches. 
In any case, I believe the commit in this PR will be helpful for the folks that want to use GPS, since remappings are not clarified in none of ros2 branches.

### DIFF
--- a/launch/dual_ekf_navsat_example.launch.py
+++ b/launch/dual_ekf_navsat_example.launch.py
@@ -1,0 +1,67 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+# Copyright 2020 Fetullah Atas
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from launch import LaunchDescription
+import launch_ros.actions
+import os
+import yaml
+from launch.substitutions import EnvironmentVariable
+import pathlib
+import launch.actions
+from launch.actions import DeclareLaunchArgument
+from ament_index_python.packages import get_package_share_directory
+
+def generate_launch_description():
+    robot_localization_dir = get_package_share_directory('robot_localization')
+    parameters_file_dir = os.path.join(robot_localization_dir, 'params')
+    parameters_file_path = os.path.join(parameters_file_dir, 'dual_ekf_navsat_example.yaml')
+    os.environ['FILE_PATH'] = str(parameters_file_dir)
+    return LaunchDescription([
+        launch.actions.DeclareLaunchArgument(
+            'output_final_position',
+            default_value='false'),
+        launch.actions.DeclareLaunchArgument(
+            'output_location',
+	    default_value='~/dual_ekf_navsat_example_debug.txt'),
+	
+    launch_ros.actions.Node(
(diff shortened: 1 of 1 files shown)


launch/dual_ekf_navsat_example.launch.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+# Copyright 2020 Fetullah Atas
```
SteveMacenski: I don't think this is a substantial enough contribution to list your copyright here

SteveMacenski:
Otherwise, looks good

launch/dual_ekf_navsat_example.launch.py
jediofgever: Adding the second copyright here isn’t necessarily about how substantial the contribution is, it is there to help next user have an sense for tracking the history of the file.

SteveMacenski (approved):

launch/dual_ekf_navsat_example.launch.py
```diff
@@ -0,0 +1,53 @@
+from launch import LaunchDescription
```
SteveMacenski, line 14: Wait, you can't just remove someone elses license when this was based on it. Please readd the OSRF license header.

SteveMacenski (changes requested):

launch/dual_ekf_navsat_example.launch.py
jediofgever: would OSRF really agree that I do bunch of unrelated changes to their code and not implying that on the license ? 
I will do what you are asking here but
with all due respect,  I hate to post things as below but I have no choice here; 

Question : What to do a File with Apache 2.0 and with modifications 
```
Section 4 of the Apache License 2.0 is quite clear on what you must do when you distribute the changed file:

    You must not remove the existing copyright claim (the one by 'the Best Company in the World')

    You must make it clear the the file has been changed. The easiest way is to simply add your copyright after the original ones:

    Modifications copyright (C) 2013 <your company/name>

    If you did the modifications on behalf of your company, then that is in most cases also the name that you must put on the copyright notice
```
link to source of above information; https://softwareengineering.stackexchange.com/questions/220068/file-with-apache-2-0-and-my-modifications

SteveMacenski: My aim isn't to get into any argument about it, but my rule of thumb on this is to be generous with copyrights that exist currently when copying files around but "significant" changes have to be added to add a new copyright marker on the header. "Significant" is ill-defined, but I usually say its something like a strong refactor or adding substantial new code to the file.  

I would qualify launch file work as not really that significant unless you were to add in a bunch of declared parameters or significantly refactor it. You still own your own copyright for this work (or your employer), listing it here doesn't gain or deny you anything. 

Either way, I don't care that much. The "most-wrong" answer here is to simply delete it along with the OSRF statement. The statement your linked to there is a clear sense that the file changed (e.g. Apache doesn't specify anything about version control, we have git and there's blaming on git so we know exactly who's modified things from that history and generated change logs on release).

SteveMacenski: If you feel strongly about having a copyright on this launch file, go for it and I'll approve it this time. But we might revisit this conversation at a later date in other PRs if you're only lightly modifying other people's work. I do have a principled issue with adding copyright headers to files without substantial changes representing authorship / such large change that you are now the primary go-to person for that file after the changes. Else, many of our files would have 20+ headers and that's just not really a great situation.

jediofgever: I do agree with you that this is a simple launch file and does not add any novelty I added the second license because of my concern here 
 > would OSRF really agree that I do bunch of unrelated changes to their code and not implying that on the license ?

since I am not affiliated with OSRF.
if that's not an issue then in the current form this PR should be ready to go. Apologies for any inconvenience.

PR #615, opened 2020-12-15
Fix update rate with gazebo clock

blacksoul000:
Default gazebo /clock publish frequency is 10Hz. With timer old initialization and "use_sim_time:=true" robot_localization frequency is limited by 10Hz

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1981,10 +1981,7 @@ void RosFilter<T>::initialize()
   }
 
   const std::chrono::duration<double> timespan{1.0 / frequency_};
-  timer_ = rclcpp::GenericTimer<rclcpp::VoidCallbackType>::make_shared(
-    this->get_clock(), std::chrono::duration_cast<std::chrono::nanoseconds>(timespan),
-    std::bind(&RosFilter<T>::periodicUpdate, this), this->get_node_base_interface()->get_context());
-  this->get_node_timers_interface()->add_timer(timer_, nullptr);
+  timer_ = this->create_wall_timer(timespan, std::bind(&RosFilter<T>::periodicUpdate, this));
 }
 
 template<typename T>


src/ros_filter.cpp
```diff
@@ -1981,10 +1981,7 @@ void RosFilter<T>::initialize()
   }
 
   const std::chrono::duration<double> timespan{1.0 / frequency_};
-  timer_ = rclcpp::GenericTimer<rclcpp::VoidCallbackType>::make_shared(
-    this->get_clock(), std::chrono::duration_cast<std::chrono::nanoseconds>(timespan),
-    std::bind(&RosFilter<T>::periodicUpdate, this), this->get_node_base_interface()->get_context());
-  this->get_node_timers_interface()->add_timer(timer_, nullptr);
+  timer_ = this->create_wall_timer(timespan, std::bind(&RosFilter<T>::periodicUpdate, this));
```
ayrton04, line 1984: I haven't used this package in earnest in ROS2, but if you are trying to run the EKF live against a bag file, and you pause it, won't this cause the EKF to continue to update?

ayrton04:
I'm not sure that switching to wall time to get past a Gazebo update rate issue is the answer. In ROS1, I thought Gazebo defaulted to 100 Hz. Is that no longer the case? Can you not change the Gazebo update rate?

blacksoul000:
@ayrton04 
[gazebo clock](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-ROS-Clocks-and-sim-time#clock-rate)
Yes, I can change gazebo rate but it's not so simple. And I don's see something about sim_time frequency limitations in robot_localization in documentation.

src/ros_filter.cpp
blacksoul000: Yes, when I pause simulation EKF update method is still calling. Is this wrong?

ayrton04:
> And I don's see something about sim_time frequency limitations in robot_localization in documentation.

I've not used ROS2 and Gazebo before, so I never thought to document the default 10 Hz limit as something to watch out for. It's not really a limitation of r_l; the package is using ROS time, which is by design, so that we can play against bag files, sim, or real robots, and not have to change anything. It's the point of ROS time.

It looks to me like the answer here is to just increase the clock publication frequency in Gazebo. I'm not sure why they separated the world update rate and the clock publication time, much less made the latter default to 10 Hz, but they've made it easy to correct.

If you want to add a note to the documentation, feel free to submit a PR. Thanks!

src/ros_filter.cpp
ayrton04: I am not really sure how this will behave offhand. What happens to the output topic when this happens?

blacksoul000: Still publishing at requested frequency with constant values

ayrton04: OK, so I think the problem here will be this: even if you put this change in and left the Gazebo clock at 10 Hz, the other parts of the EKF that use the clock will (and must) continue to use ROS time for updating the filter. I'm assuming the internal ROS callback timing is also based on ROS time, but I could be wrong. If I'm not, though, your state estimate itself is still only going to actually update at 10 Hz, even if you tell the EKF to publish at 50 Hz. You'll get ~5 messages from the EKF with the same data in it, and then it will change and repeat for 5 more, etc.

In the end, I think the solution here is to make sure your Gazebo clock update rate is >= your EKF frequency. Feel free to add a note to the documentation somewhere with a note about this.

PR #618, opened 2021-01-12
yaml-cpp using find_package as backup

Tobias-Fischer:
On some systems the yaml-cpp.pc does not exist, but instead yaml-cpp can be found using cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,9 @@ find_package(catkin REQUIRED COMPONENTS
 
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(YAML_CPP yaml-cpp)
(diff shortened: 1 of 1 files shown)


SteveMacenski:
I think ROS uses YAML vendors to make sure we support the right version for an operating system. We shouldn't be directly find_package-ing it if we have a vandorization

Tobias-Fischer:
Fair enough, that makes sense. Feel free to close if it does not make sense here directly - we need this patch for our efforts on ROS+Conda (see https://github.com/RoboStack/ros-noetic), but can patch it there directly.

SteveMacenski:
Lets see what Tom says

ayrton04:
This appears to only execute if the prescribed package finding method fails, which seems harmless enough.

Issue #619, opened 2021-01-29
Robot_localization services /toLL and /fromLL respond with different result on the same data


kosmastsk:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Binaries
- Version or commit hash:
  - 2.6.9

#### Steps to reproduce issue
I am using the robot_localization package and in the terminal, I run the services `/fromLL` and `/toL`L for the same position. But the results do not agree. Specifically, running:  
`rosservice call /robot/fromLL "ll_point:
  latitude: 40.5665059791
  longitude: 22.9989873398
  altitude: 122.0"`
returns:
`map_point: 
  x: 140892.72962
  y: 5405179.36945
  z: 0.0`
And when I run:
`rosservice call /robot/toLL "map_point:
  x: 0.0
  y: 0.0
  z: 0.0"`
I get:
`ll_point: 
  latitude: 40.5665059791
  longitude: 22.9989873398
  altitude: 122.375866815`
which is the same position as the first one. 

#### Expected behavior
The lat/long and the map points should be the same in both cases, no matter the direction that the conversion is asked.

#### Additional information
The solution is described in ros.answers here: [https://answers.ros.org/question/369570/robot_localization-services-toll-and-fromll-respond-with-different-result/](https://answers.ros.org/question/369570/robot_localization-services-toll-and-fromll-respond-with-different-result/)

ayrton04:
Thanks for the report. Will fix this ASAP.

ayrton04:
Fixed with https://github.com/cra-ros-pkg/robot_localization/pull/620. This doesn't appear to affect the foxy ROS2 branch, but the fix is now in melodic and noetic.

PR #620, opened 2021-01-29
Fixing lat-long to UTM conversion

ayrton04:
Addresses https://github.com/cra-ros-pkg/robot_localization/issues/619.

FYI @kosmastsk

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -412,7 +412,7 @@ namespace RobotLocalization
     else
     {
       std::string utm_zone_tmp;
-      NavsatConversions::LLtoUTM(latitude, longitude, cartesian_x, cartesian_y, utm_zone_tmp);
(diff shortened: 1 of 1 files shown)


ayrton04:
@kosmastsk, would you be able/willing to test this?

kosmastsk:
@ayrton04 I will test it soon and let you know if the issue is fixed.

SteveMacenski:
@kosmastsk please submit for ROS2 as well

ayrton04:
I'll take care of porting it once we've confirmed it works.

kosmastsk:
@ayrton04 I just checked it with the same data that I used when I made the issue, and now it works fine! Thank you!
Do you want me to file an issue for nav2 as well?

SteveMacenski (approved):
No issues if you say it works! I'm not working with GPS right now to be able to independently verify but the code looks good (obviously...)

PR #623, opened 2021-02-19
Update state_estimation_nodes.rst

JJBUT:
Make the intention/use/function of the publish_tf parameter clear. In my experience, what the current description describes is not actually what happens. I add a description that indicates what happens with the two most common (and suggested) use cases.


ayrton04:
That's a better description, agreed. Thanks.

PR #624, opened 2021-02-21
SHARED linking for Geographiclib

AchmadFathoni:
Arch Linux doesn't install static library by default. So I made this PR for AUR package patch. It is OK if this PR is not eventually accepted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ endif()
 
 # Geographiclib installs FindGeographicLib.cmake to this non-standard location
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "/usr/share/cmake/geographiclib/")
-find_package(GeographicLib REQUIRED COMPONENTS SHARED)
+find_package(GeographicLib REQUIRED)
 
 # Attempt to find Eigen using its own CMake module.
 # If that fails, fall back to cmake_modules package.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ endif()
 
 # Geographiclib installs FindGeographicLib.cmake to this non-standard location
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "/usr/share/cmake/geographiclib/")
-find_package(GeographicLib REQUIRED COMPONENTS STATIC)
+find_package(GeographicLib REQUIRED COMPONENTS SHARED)
```
acxz: Is there a way to just remove the explicit `STATIC`/`SHARED` dependency on the line?
Something like just
```suggestion
find_package(GeographicLib REQUIRED)
```

CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ endif()
 
 # Geographiclib installs FindGeographicLib.cmake to this non-standard location
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "/usr/share/cmake/geographiclib/")
-find_package(GeographicLib REQUIRED COMPONENTS STATIC)
+find_package(GeographicLib REQUIRED)
```
ayrton04, line 37: @AGummyBear Was there any reason for the static linkage previously?

AGummyBear: My thinking was just that it would result in not needing to install GeographicLib as a dependency unless building, since I think most users are probably running this package as installed. I'm fine with the change to shared though. There was no technical reason to use static.

PR #626, opened 2021-02-23
UTM conversions using geographiclib

Timple:
For #222 it made sense to use the GeographicLib library that #575 already introduced.

In preparation I rewrote the current UTM functions.

### DIFF
--- a/include/robot_localization/navsat_conversions.h
+++ b/include/robot_localization/navsat_conversions.h
@@ -175,86 +175,32 @@ static inline void LLtoUTM(const double Lat, const double Long,
 }
 
 /**
- * Converts UTM coords to lat/long.  Equations from USGS Bulletin 1532
+ * Converts UTM coords to lat/long.
  *
  * East Longitudes are positive, West longitudes are negative.
  * North latitudes are positive, South latitudes are negative
  * Lat and Long are in fractional degrees.
- * Meridian convergence is computed as for Spherical Transverse Mercator,
- * which gives quite good approximation.
- * 
- * @param[out] gamma meridian convergence at point (degrees).
  *
- * Written by Chuck Gantz- chuck.gantz@globalstar.com
+ * @param[out] gamma meridian convergence at point (degrees).
  */
 static inline void UTMtoLL(const double UTMNorthing, const double UTMEasting,
                            const std::string &UTMZone, double& Lat, double& Long, double &gamma)
 {
-  double k0 = UTM_K0;
-  double a = WGS84_A;
-  double eccSquared = UTM_E2;
-  double eccPrimeSquared;
-  double e1 = (1-sqrt(1-eccSquared))/(1+sqrt(1-eccSquared));
-  double N1, T1, C1, R1, D, M;
-  double LongOrigin;
-  double mu, phi1Rad;
-  double x, y;
-  int ZoneNumber;
-  char* ZoneLetter;
-
-  x = UTMEasting - 500000.0;  // remove 500,000 meter offset for longitude
-  y = UTMNorthing;
-
-  ZoneNumber = strtoul(UTMZone.c_str(), &ZoneLetter, 10);
-  if ((*ZoneLetter - 'N') < 0)
-          {
-            // remove 10,000,000 meter offset used for southern hemisphere
-            y -= 10000000.0;
-          }
-
-  // +3 puts origin in middle of zone
-  LongOrigin = (ZoneNumber - 1)*6 - 180 + 3;
-  eccPrimeSquared = (eccSquared)/(1-eccSquared);
-
-  M = y / k0;
-  mu = M/(a*(1-eccSquared/4-3*eccSquared*eccSquared/64
-                   -5*eccSquared*eccSquared*eccSquared/256));
-
-  phi1Rad = mu + ((3*e1/2-27*e1*e1*e1/32)*sin(2*mu)
-                        + (21*e1*e1/16-55*e1*e1*e1*e1/32)*sin(4*mu)
-                        + (151*e1*e1*e1/96)*sin(6*mu));
-
-  N1 = a/sqrt(1-eccSquared*sin(phi1Rad)*sin(phi1Rad));
-  T1 = tan(phi1Rad)*tan(phi1Rad);
-  C1 = eccPrimeSquared*cos(phi1Rad)*cos(phi1Rad);
-  R1 = a*(1-eccSquared)/pow(1-eccSquared*sin(phi1Rad)*sin(phi1Rad), 1.5);
-  D = x/(N1*k0);
-
-  Lat = phi1Rad - ((N1*tan(phi1Rad)/R1)
-                         *(D*D/2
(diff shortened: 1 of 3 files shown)


ayrton04:
Thanks for this, and for the tests. Are you happy to have this reviewed now?

Timple:
Certainly.

include/robot_localization/navsat_conversions.h
```diff
@@ -135,232 +138,69 @@ static inline void UTM(double lat, double lon, double *x, double *y)
...
-  else if ((-64 > Lat) && (Lat >= -72)) LetterDesignator = 'D';
-  else if ((-72 > Lat) && (Lat >= -80)) LetterDesignator = 'C';
-        // 'Z' is an error flag, the Latitude is outside the UTM limits
-  else LetterDesignator = 'Z';
-  return LetterDesignator;
-}
-
-/**
- * Convert lat/long to UTM coords.  Equations from USGS Bulletin 1532
+ * Convert lat/long to UTM coords.
```
ayrton04, line 142: Can we also get rid of all of this? Or do we still need it?

https://github.com/cra-ros-pkg/robot_localization/blob/ab6e85139d42989e6b988f61fa19aa9583b72afd/include/robot_localization/navsat_conversions.h#L65-L139

And if we do get rid of that, can we also get rid of the other comments and update the license?

https://github.com/cra-ros-pkg/robot_localization/blob/ab6e85139d42989e6b988f61fa19aa9583b72afd/include/robot_localization/navsat_conversions.h#L65-L139

https://github.com/cra-ros-pkg/robot_localization/blob/ab6e85139d42989e6b988f61fa19aa9583b72afd/include/robot_localization/navsat_conversions.h#L65-L139

test/test_navsat_conversions.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "robot_localization/navsat_conversions.h"
+
+#include <gtest/gtest.h>
+
+#include <string>
+
+void NavsatConversionsTest(const double lat, const double lon,
+                           const double UTMNorthing, const double UTMEasting, const std::string UTMZone)
+{
+    double UTMNorthing_new;
```
ayrton04: Mind if we stick with two-space indentations?

include/robot_localization/navsat_conversions.h
Timple: We can.
I left them in because they are exposed to other packages. So I'm not sure who might be using these variables...

ayrton04: That's a good point. I'll bring this forward to the ROS2 branches, and for Galactic, I'll remove them. Thanks.

ayrton04 (approved):

include/robot_localization/navsat_conversions.h
Timple: I think the whole conversions file can be removed in that case. Since there is now a library doing all that.

ayrton04: Agreed.

ayrton04:
@SteveMacenski, anything glaring to you on this or #627?

SteveMacenski:
I haven't used geographic lib / GPS before in a product so I can't say I have an opinion. Though I'd want a PR analog to this for ROS2 to keep them in sync

Timple:
I could do that. Which branch(es) would I need to target? And which of those would allow breaking changes?

SteveMacenski:
The ros2 branch would allow breaking changes

Timple:
> Though I'd want a PR analog to this for ROS2 to keep them in sync

#643 

I could not target the ROS2 branch because it's too much behind on Foxy.

SteveMacenski:
@Timple what do you mean, I'm looking at the [ROS2->Foxy diff right here](https://github.com/cra-ros-pkg/robot_localization/compare/foxy-devel...ros2) its only off by 4 commits that are just rclcpp API changes (only covering like ~80 lines). We need any changes to Foxy to be in `ros2` as well or else this change will never be reflected into future distributions and we'll be back here every 12 months

Timple:
Well that is strange. Than how come there is no mention of `geographiclib` in the[ ros2](https://github.com/cra-ros-pkg/robot_localization/blob/ros2/package.xml) package.xml but there is in the [foxy-devel](https://github.com/cra-ros-pkg/robot_localization/blob/foxy-devel/package.xml) branch?

Did a merge go wrong somewhere?

SteveMacenski:
It looks like those were added 2 weeks ago only in the ROS1 feature sync PR https://github.com/cra-ros-pkg/robot_localization/commit/3dbd48a56c927c035b22c3495926c3f38c2bdac2 

@ayrton04 is it your intention to merge this into `ros2`? Any conflicts you'd expect from cherry picking between foxy and ros2?

ayrton04:
I can't think of anything offhand.

For the record, the model that I originally tried to follow was;

1. Create a new `<release_name>-devel` branch prior to the release being finalized
2. Let all the new featured (and ABI breakages) go into `<release_name>-devel`
3. When `<release_name>-devel` is released, that branch should only accept bug fixes or minor features that don't break ABI
4. When `<release_name>-devel` is released, we also create `<next_release_name>-devel` and start putting both the bug fixes from `<release_name>-devel` and new features into `<next_release_name>-devel`.

Obviously, there have been some violations of that paradigm, but that's generally the way I wanted to do things.

The `ros2` branch suggests a slightly different model, I think, though I'm not sure what we do for new ROS2 releases.

1. All new features and bug fixes to any `<release_name>-devel` branch should find their way to `ros2`
2. If a new ROS2 release is cut, do we then branch `ros2` into `<previous_release>-devel` and let all future development happen on the `ros2` branch? Or do we just keep using `ros2` for all ROS2 releases until ABI breakage or incompatibility with the rest of the stack for a given ROS2 release forces us to branch?

Timple:
Both suggestions result in the `ros2` branch to be some kind of a 'rolling' branch allowing breaking changes.

> All new features and bug fixes to any <release_name>-devel branch should find their way to ros2

Certainly this. 

---

So I'll target the `foxy-devel` for the backwards compatibility I build in. And I'll target the `ros2` (once it gets updated) to include the breaking changes proposed.

SteveMacenski:
The new ROS2-y workflow is 
- Have a master `ros2` or similarly named branch that all development is done on, targeting ROS2 rolling / master builds
- When you release a new distro, you branch off `ros2` at the time of a new distro and that's the new distro release to use
- Over time, sync the distro and ros2 branches by cherry picking commits from `ros2` to the distro that are ABI compatible for release to those distros

Depending on my laziness, I also sometimes ask people to target their PR to multiple branches (ros2, and foxy, and noetic or something) to avoid the cherry picking later / forgetting.

include/robot_localization/navsat_conversions.h
Timple: @ayrton04 do you still plan for removing these (quite empty) functions for ROS2 and galactic?

Seems like they are re-used: https://github.com/cra-ros-pkg/robot_localization/pull/676
Which is a bit of a waste.

Issue #309, opened 2016-10-08
Update navsat_transform_node to reflect changes to REP-105


ayrton04:
[REP-105](http://www.ros.org/reps/rep-0105.html) now includes an `earth` frame. `navsat_transform_node` should be updated to match.

PR #627, opened 2021-02-23
Stick to the global utm_zone_ when transforming gps to UTM

### LINKED ISSUE #222: Handle crossing of UTM zone boundaries

Timple:
Fixes #222 partly.

It doesn't handle multiple UTM zones but at least sticks to the initial one. Preventing the jump in position.

For the actual multiple UTM solution #309 would need to be solved.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -110,6 +110,7 @@ namespace RobotLocalization
 
     to_ll_srv_ = nh.advertiseService("toLL", &NavSatTransform::toLLCallback, this);
     from_ll_srv_ = nh.advertiseService("fromLL", &NavSatTransform::fromLLCallback, this);
+    set_utm_zone_srv_ = nh.advertiseService("setUTMZone", &NavSatTransform::setUTMZoneCallback, this);
 
     if (use_manual_datum_ && nh_priv.hasParam("datum"))
     {
@@ -427,6 +428,17 @@ namespace RobotLocalization
 
     response.map_point = cartesianToMap(cartesian_pose).pose.pose.position;
 
+    return true;
+  }
+
+  bool NavSatTransform::setUTMZoneCallback(robot_localization::SetUTMZone::Request& request,
+                                           robot_localization::SetUTMZone::Response& response)
+  {
+    double x_unused;
+    double y_unused;
+    int prec_unused;
+    GeographicLib::MGRS::Reverse(request.utm_zone, utm_zone_, northp_, x_unused, y_unused, prec_unused, true);
+    ROS_INFO("UTM zone set to %d %s", utm_zone_, northp_ ? "north" : "south");
     return true;
   }
 
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -61,7 +61,7 @@ namespace RobotLocalization
     yaw_offset_(0.0),
     base_link_frame_id_("base_link"),
     gps_frame_id_(""),
-    utm_zone_(""),
+    utm_zone_(0),
     world_frame_id_("odom"),
     transform_timeout_(ros::Duration(0)),
     tf_listener_(tf_buffer_)
@@ -411,8 +411,9 @@ namespace RobotLocalization
     }
     else
     {
-      std::string utm_zone_tmp;
-      NavsatConversions::LLtoUTM(latitude, longitude, cartesian_y, cartesian_x, utm_zone_tmp);
+      int zone_tmp;
+      bool nortp_tmp;
+      GeographicLib::UTMUPS::Forward(latitude, longitude, zone_tmp, nortp_tmp, cartesian_x, cartesian_y, utm_zone_);
     }
 
     cartesian_pose.setOrigin(tf2::Vector3(cartesian_x, cartesian_y, altitude));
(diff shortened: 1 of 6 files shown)


src/navsat_transform.cpp
```diff
@@ -411,8 +413,9 @@ namespace RobotLocalization
     }
     else
     {
-      std::string utm_zone_tmp;
-      NavsatConversions::LLtoUTM(latitude, longitude, cartesian_y, cartesian_x, utm_zone_tmp);
+      int zone_tmp;
+      bool nortp_tmp;
+      GeographicLib::UTMUPS::Forward(latitude, longitude, zone_tmp, nortp_tmp, cartesian_x, cartesian_y, utm_zone_);
```
ayrton04, line 418: Why can't we keep using `LLtoUTM` here? Doesn't your other PR just wrap to the GeographicLib call anyway?

src/navsat_transform.cpp
```diff
@@ -473,11 +488,12 @@ namespace RobotLocalization
     }
     else
     {
-      NavsatConversions::UTMtoLL(odom_as_cartesian.getOrigin().getY(),
-                                 odom_as_cartesian.getOrigin().getX(),
-                                 utm_zone_,
-                                 latitude,
-                                 longitude);
+      GeographicLib::UTMUPS::Reverse(utm_zone_,
```
ayrton04, line 491: Same here. If we updated the methods in the other PR to call the GeographicLib equivalents, do we need to directly use GeographicLib here? Same for all the direct references to GeographicLib.

src/navsat_transform.cpp
Timple: Sorry for the confusion.

I started on the other PR. But since that makes the whole conversion functions almost empty, one might as well use the functions directly, removing all custom functions. Actually I would want to remove the whole conversions file in this PR. But I don't know who might be using it.

The reason that this is also needed is because the library uses a boolean `northp` instead of string parsing the `utm_zone_` variable. It's much easier to track the boolean than to keep the information in a string and encode/decode that.

ayrton04: OK, sounds good. As I said in the other PR, we'll remove that file for ROS2 Galactic onward.

ayrton04 (approved):

Timple:
Clarifying note:

As the diff shows, instead of `std::string utm_zone_` there are now two variables `int utm_zone_` and `bool northp_`. This is inline with the definition in the geographiclib.

These two variables where first 'combined' in a string and had to be encoded/decoded

ayrton04:
@Timple, would you mind rebasing and fixing the conflict? Thanks.

Timple:
Done

ayrton04:
Thank you for the tests. It makes me feel better about not having a live robot with a GPS for testing. :-/

Timple:
I actually created the test first, because that was initially less work than dragging the live robot outside :)

Timple:
Would it be possible to get a melodic and noetic release with this fix/feature?

ayrton04:
Yes, will bloom this today.

ayrton04:
@Timple, do you need this to go into melodic as well?

Timple:
Yes please, we're migrating to Noetic, but the actual machine is still on melodic.

ayrton04:
OK, will cherry-pick this change.

ayrton04:
Does that mean you also want this in melodic?

https://github.com/cra-ros-pkg/robot_localization/commit/5eec32bfbbb1681a59ff295a52e5f4034427480d

Timple:
Yes to both. Simulator is already on Noetic while actual machine is Melodic.

edit: I now realize melodic has it's own branch without this patch. If this is too much hassle we can simply build from source until the migration is finished. One more reason to push for Noetic :slightly_smiling_face:

ayrton04:
OK. I'll just go back through the recent noetic commits and find anything that should be backported. Thanks.

ayrton04:
@Timple, if you have some time, would you mind trying out the `melodic-devel` branch? I cherry-picked the commits I believe we needed. If it works as you'd expect, I'll release it.

Timple:
Yup, seems to work!

Timple:
Should I also port this to the ROS2 branch?

And a subtle ping for the bloom release :slightly_smiling_face:

ayrton04:
Bloom release was done [a few weeks ago](https://github.com/ros/rosdistro/commit/4852893f1ff73c2607f0ad12a919047a45473831#diff-8bc0a880ad8fcb113c8937d81165e14ce415f99c2765b381170af9f282e1be4c). Did they do a sync recently?

EDIT: [Melodic, too](https://github.com/ros/rosdistro/commit/053ab39bc591cb371ac551db63f6dcbfda998168#diff-7bb80f0172ce917506e9060a609badb4304bb3dec88cd14acc4f5c6fc5e3774b).

Timple:
Ow, whups. I only looked at the latest entry in https://github.com/cra-ros-pkg/robot_localization/blob/noetic-devel/CHANGELOG.rst

But the last version is not on top it seems...

ayrton04:
Odd! I'll correct that in the next one.

Issue #621, opened 2021-02-14
UKF transfer function does not capture nonlinearity correctly


jola6897:
## Bug report

**Required Info:**
- Version or commit hash:
melodic-devel branch as of hash a7da2f6ea29bb2606358438376fdce8b3461b5c2

#### Steps to reproduce issue
Check the `predict` function inside [ukf.cpp](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ukf.cpp).

#### Expected behavior
The non-linear state transfer function is evaluated on each sigma point  individually.

#### Actual behavior
In [ukf.cpp](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ukf.cpp) inside the `predict` function the sigma points are generated and the `transferFunction_` is applied to them. So basically we try to estimate the a priori state. Using sigma points allows us to capture also non-linear functions quite easy as in theory `transferFunction_` could also be non-linear. And indeed the function that is used here is non-linear.

BUT: The problem is that in the implementation `transferFunction_`  is a matrix which is the same for all sigma points. It is obtained by evaluating the non-linear function at the a posteriori state (`state_`). This means that the `transferFunction_` is **only** correct for the a posteriori state (`state_`) which happens to be also the first sigma point. All other sigma points have by construction a slightly different orientation therefore `roll`, `pitch`, `yaw` are different for each sigma point. This means that the actual behavior is not captured correctly and that the a priori state and covariance are not correct either. If the variance of the orientation is small then the error will be also really small because the sampled sigma points in terms of orientation are quite close together, but this is not always the case. 

#### Additional information
My feedback is based on "A New Extension of the Kalman Filter to Nonlinear Systems" that is also referenced in "A Generalized Extended Kalman Filter Implementation for the Robot Operating System".

I do not know if this behavior was intended to speed up computation or maybe there is also a proof that this is a sufficient approximation that I am unaware of. If so could you please share the reasons for your decision to implement it like that, as I would like to understand it.

Thanks a lot for your effort. Looking forward to your feedback.

ayrton04:
Nope, definite oversight on that front. The transfer function matrix formulation should really go away. Care to submit a PR that constructs the transfer function matrix for every sigma point? The computation expense will go up, but the behaviour will at least be correct.

jola6897:
Hi,

I added a pull request that should resolve this bug (#628). If you accept this pull request or implement  something similar you can close this issue :).

[...]

PR #628, opened 2021-02-26
Fixed state transition of sigma points in UKF

jola6897:
Hi,
I put a fix for #621 (Incorrect state transition of sigma points). The state transition function for each sigma point now uses its sigma point's posteriori state instead of always the posteriori state.

Hope the changes are in line with your code style and satisfy your requirement for documentation. I kept the actual state transition function as is, to limit the number of changes needed. From a concept standpoint I think transferFunction_ might not be necessary in the UKF - a local matrix should be sufficient also.

### DIFF
--- a/include/robot_localization/ukf.h
+++ b/include/robot_localization/ukf.h
@@ -89,6 +89,17 @@ class Ukf: public FilterBase
     void predict(const double referenceTime, const double delta);
 
   protected:
+    //! @brief Carries out the predict step for the posteriori state of a sigma
+    //! point.
+    //!
+    //! Projects the state and error matrices forward using a model of
+    //! the vehicle's motion.
+    //!
+    //! @param[in] posterioriState - State of teh sigma point.
+    //! @param[in] delta - The time step over which to predict.
+    //!
+    Eigen::VectorXd predict(Eigen::VectorXd const& posterioriState, double delta);
+
     //! @brief The UKF sigma points
     //!
     //! Used to sample possible next states during prediction.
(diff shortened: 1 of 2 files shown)


ayrton04:
Thanks for this! Will try to review soon.

jola6897:
Hi,
actually I noticed that I only fixed the predict function, but the correct function still does not properly compute the sigma points. I do not yet know if/when I will have time to continue this.

jola6897:
As mentioned in  #621, there are more issues with the UKF that are not fixed here. Therefore this pull request only fixes part of the problem.

ayrton04:
This LGTM. I need to close/reopen this to force the build farm to run again. Not sure why those tests are failing.

ayrton04:
I haven't forgotten about this. It looks like some of the failing tests are UKF tests that are failing by wide margins. Even with these changes, I wouldn't expect the state estimate to change that much in the tests. Will have to look into what's going on with those.

ayrton04:
Ok, a couple things:

1. I know why the tests are failing, and have fixed it in another branch. I'll merge this, rebase that branch, and PR it.
2. Re: this comment:

> actually I noticed that I only fixed the predict function, but the correct function still does not properly compute the sigma points

The correct function doesn't need to do projection/prediction, though, right? It's just computing the actual sigma points, as opposed to creating the sigma points _and_ passing them through the state transition function. That code block is only there to handle the specific case where we have > 1 measurement with the exact same time stamp. In that event, we'd call `predict` once, then call `correct` with the first measurement, and then `correct` again with the second.

I may be missing something else, though.

I won't bloom this branch again until I fix the sampling of the Euler angles.

PR #631, opened 2021-03-17
Fix deprecation warnings for upcoming ROS Galactic release

jacobperron:
Building against the latest version of rclcpp produces some deprecation warnings. This PR fixes warnings related to `rclcpp::Duration` construction (6b36efd) and `rclcpp::Node::declare_parameter` (7055b82).

Related tickets upstream, deprecating usage:

- https://github.com/ros2/rclcpp/pull/1432
- https://github.com/ros2/rclcpp/pull/1522

The second commit (7055b82) related to declare_parameter is not compatible with Foxy, so I recommend creating a new `galactic-devel` branch and we can retarget this PR.

### DIFF
--- a/src/robot_localization_estimator.cpp
+++ b/src/robot_localization_estimator.cpp
@@ -182,7 +182,7 @@ void RobotLocalizationEstimator::extrapolate(
   rclcpp::Time time_stamp = rclcpp::Time(
     boundary_state.time_stamp *
     1000000000);
-  rclcpp::Duration delta_duration = rclcpp::Duration(delta * 1000000000);
+  rclcpp::Duration delta_duration = rclcpp::Duration::from_nanoseconds(delta * 1000000000);
   filter_->predict(time_stamp, delta_duration);
 
   state_at_req_time.time_stamp = requested_time;
--- a/src/robot_localization_estimator.cpp
+++ b/src/robot_localization_estimator.cpp
@@ -182,7 +182,7 @@ void RobotLocalizationEstimator::extrapolate(
   rclcpp::Time time_stamp = rclcpp::Time(
     boundary_state.time_stamp *
     1000000000);
-  rclcpp::Duration delta_duration = rclcpp::Duration::from_nanoseconds(delta * 1000000000);
+  rclcpp::Duration delta_duration = rclcpp::Duration::from_seconds(delta);
   filter_->predict(time_stamp, delta_duration);
 
   state_at_req_time.time_stamp = requested_time;
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -833,19 +833,17 @@ void RosFilter<T>::loadParams()
 
   // Transform future dating
   double offset_tmp = this->declare_parameter("transform_time_offset", 0.0);
-  tf_time_offset_ =
-    rclcpp::Duration::from_nanoseconds(filter_utilities::secToNanosec(offset_tmp));
+  tf_time_offset_ = rclcpp::Duration::from_seconds(offset_tmp);
 
   // Transform timeout
   double timeout_tmp = this->declare_parameter("transform_timeout", 0.0);
-  tf_timeout_ = rclcpp::Duration::from_nanoseconds(filter_utilities::secToNanosec(timeout_tmp));
+  tf_timeout_ = rclcpp::Duration::from_seconds(timeout_tmp);
 
   // Update frequency and sensor timeout
   frequency_ = this->declare_parameter("frequency", 30.0);
 
   double sensor_timeout = this->declare_parameter("sensor_timeout", 1.0 / frequency_);
-  filter_.setSensorTimeout(
-    rclcpp::Duration::from_nanoseconds(filter_utilities::secToNanosec(sensor_timeout)));
+  filter_.setSensorTimeout(rclcpp::Duration::from_seconds(sensor_timeout));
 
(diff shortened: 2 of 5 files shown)


src/ros_filter.cpp
```diff
@@ -834,18 +834,18 @@ void RosFilter<T>::loadParams()
   // Transform future dating
   double offset_tmp = this->declare_parameter("transform_time_offset", 0.0);
   tf_time_offset_ =
-    rclcpp::Duration(filter_utilities::secToNanosec(offset_tmp));
+    rclcpp::Duration::from_nanoseconds(filter_utilities::secToNanosec(offset_tmp));
```
SteveMacenski: this and the 2 under are converting from s -> ns -> duration. You could remove the `filter_utilities::secToNanosec` call and replace it with just `from_seconds`

src/ros_filter.cpp
```diff
@@ -604,7 +604,7 @@ void RosFilter<T>::integrateMeasurements(const rclcpp::Time & current_time)
       const std::string first_measurement_topic =
         first_measurement->topic_name_;
       // revertTo may invalidate first_measurement
-      if (!revertTo(first_measurement_time - rclcpp::Duration(1))) {
+      if (!revertTo(first_measurement_time - rclcpp::Duration(1ns))) {
```
SteveMacenski, line 607: @ayrton04 should this be 1ns, 1s, ...?

SteveMacenski (changes requested):

src/ros_filter.cpp
jacobperron: Thanks, I meant to ask the same question.

SteveMacenski:
@jacobperron I opened a `ros2` branch to target it to since those `from_XYZ`'s aren't in foxy

src/ros_filter.cpp
jacobperron: Of course :upside_down_face: 
See ca9e807

jacobperron:
Hold up, there's more warnings in other files I've overlooked.

jacobperron:
> Hold up, there's more warnings in other files I've overlooked.

Okay, I think I got them all.

src/robot_localization_estimator.cpp
```diff
@@ -182,7 +182,7 @@ void RobotLocalizationEstimator::extrapolate(
   rclcpp::Time time_stamp = rclcpp::Time(
     boundary_state.time_stamp *
     1000000000);
-  rclcpp::Duration delta_duration = rclcpp::Duration(delta * 1000000000);
+  rclcpp::Duration delta_duration = rclcpp::Duration::from_nanoseconds(delta * 1000000000);
```
SteveMacenski: `from_seconds()` and remove the 1e9 multiplier

SteveMacenski:
Otherwise LGTM, waiting for Tom's response on that one comment

src/ros_filter.cpp
jacobperron: If I understand the equivalent ROS 1 code, it is also subtracting 1 nanosecond:

https://github.com/cra-ros-pkg/robot_localization/blob/a2da33bb3aaa4b10358de391221fe0ed97b16f49/src/ros_filter.cpp#L611

src/robot_localization_estimator.cpp
jacobperron: Done 774841b

src/ros_filter.cpp
```diff
@@ -815,7 +815,7 @@ void RosFilter<T>::loadParams()
   // Try to resolve tf_prefix
   std::string tf_prefix = "";
   std::string tf_prefix_path = "";
-  this->declare_parameter("tf_prefix");
+  this->declare_parameter<std::string>("tf_prefix");
```
jacobperron: I misunderstood how the new static type enforcement works. This code will throw an exception at this line if "tf_prefix" is not provided, but we want it to be optional.

jacobperron: I've addressed this by handling NoParameterOverrideProvided exceptions when no parameters are provided: 3326e80

src/ros_filter.cpp
ayrton04: So the PRs involved here are

https://github.com/cra-ros-pkg/robot_localization/pull/266/files

and

https://github.com/cra-ros-pkg/robot_localization/pull/279/files

The only reason we subtract any time at all is to ensure that we go to a measurement (presumably the first) that is before `first_measurement_time`. But we could also just change this to use `>=`:

https://github.com/cra-ros-pkg/robot_localization/blob/3326e80a4169365b4570b5adcd1a63f3e37ab54f/src/ros_filter.cpp#L3322-L3323

If I had to guess (I didn't dig too far), the measurement times were likely stored as doubles originally, and the `1e-9` thing was just me being careful with floating point numbers. But ROS times have exact representations, so I think the 1e-9 could go, and we can change the `revertTo` reverse iteration to check for `>=` instead of just `>`.

ayrton04: However, that change could be outside of scope for this. It was meant to be 1 nanosecond, yes.

ayrton04 (approved):

ayrton04:
@SteveMacenski, feel free to merge. I know you said LGTM, but your review still had comments requested, so I re-requested your review.

SteveMacenski (approved):

PR #632, opened 2021-03-17
[foxy] Expect bool return type for get_parameter calls

jacobperron:
In some places, the code was expecting a PARAMETER_NOT_SET value from get_parameter
when actually a bool is being returned.

The code just happened to work before since PARAMETER_NOT_SET has value 0, which is
equivalent to false.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1069,7 +1069,7 @@ void RosFilter<T>::loadParams()
     this->declare_parameter(odom_topic_name);
 
     rclcpp::Parameter parameter;
-    if (rclcpp::PARAMETER_NOT_SET != this->get_parameter(odom_topic_name, parameter)) {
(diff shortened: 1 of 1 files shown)


jacobperron:
We should either target the new `ros2` branch or cherry-pick this forward after it lands.

SteveMacenski:
Is this needed for foxy at all or just `ros2`?

jacobperron:
> Is this needed for foxy at all or just ros2?

I think this is a bug in all ROS 2 branches.

SteveMacenski:
Got it - can you open the PRs? I'll merge in bulk

SteveMacenski (approved):

jacobperron:
PRs for Dashing and Rolling:

- https://github.com/cra-ros-pkg/robot_localization/pull/637
- https://github.com/cra-ros-pkg/robot_localization/pull/638

PR #633, opened 2021-03-18
[foxy] Fix format string warnings

jacobperron:
Recent changes to the rclcpp logging macros exposed a couple bugs.


SteveMacenski (approved):
@jacobperron can you open an analog for ros2 since that's now branched off?

jacobperron:
> can you open an analog for ros2 since that's now branched off?

Sure. I guess technically this change applies to all branches, not just ROS 2 branches.

jacobperron:
I've opened equivalent PRs for `dashing-devel` and `ros2` branches. I skipped Eloquent since it is EOL and this bug does not exist in ROS 1, which uses string streams.

PR #639, opened 2021-03-23
Transform gravitational vector to IMU frame before removing acceleration

PaulVerhoeckx:
Hi,

When the IMU message does not contain an orientation, the gravitational vector should be transformed to the IMU frame before correcting for the gravitational accelerations.

Currently it is only corrected for the current attitude ([here](https://github.com/cra-ros-pkg/robot_localization/blob/a5eda3344f1c802a50f50164df32373999e053de/src/ros_filter.cpp#L2455-L2466)), but not for the transformation between `base_link` and the `imu_frame`. Thereto this is only an issue when these two frames have different orientations. 

This pull requests proposes a solution to this.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2444,7 +2444,6 @@ namespace RobotLocalization
       if (removeGravitationalAcc_[topicName])
       {
         tf2::Vector3 normAcc(0, 0, gravitationalAcc_);
-        tf2::Quaternion curAttitude;
         tf2::Transform trans;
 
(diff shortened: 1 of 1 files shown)


ayrton04 (approved):
LGTM. Thanks for this.

ayrton04:
Would you mind making sure this gets PR-ed into the `ros2` branch as well?

PaulVerhoeckx:
> Would you mind making sure this gets PR-ed into the `ros2` branch as well?

I made a new PR (#649) for ros2

PR #640, opened 2021-03-23
Switch Foxy to use yaml_cpp_vendor

smnogar:
I was getting yaml_cpp related linking errors when trying to build in ros2 foxy. Tested this in ubuntu 20.04 and it seemed to work.


SteveMacenski (approved):
Perfect, can you submit this PR against `ros2` too so that it fixes later distributions (galactic, etc)?

smnogar:
Done. See https://github.com/cra-ros-pkg/robot_localization/pull/641

SteveMacenski:
Thanks!

PR #643, opened 2021-03-31
Utm using geographiclib foxy

Timple:
Parallel to #626

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,7 +258,7 @@ if(BUILD_TESTING)
 
   #### NAVSAT CONVERSION TESTS ####
   ament_add_gtest(navsat_conversions-test test/test_navsat_conversions.cpp)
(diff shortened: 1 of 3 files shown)


SteveMacenski:
I leave this to @ayrton04 too since I really don't know much about this. If it compiles / dependencies are properly marked in package.xml/cmakelists, I see no issue with it

SteveMacenski:
https://build.ros2.org/job/Fpr__robot_localization__ubuntu_focal_amd64/31/gcc/

CI failed due to an unused variable warning

Timple:
The unused variable is there for backwards/abi compatibility. Shall I mark it as ignored?

SteveMacenski:
Which?

Timple:
The compiler warning was correct. `gamma` is actually a function-output, so it should have been set. 
This was correct in the noetic implementation, got lost in migration I guess. Fixed now.

ayrton04 (approved):

PR #652, opened 2021-04-15
Fix/navsat transform

Timple:
If the test is overdone (I'm adding to it later in a follow-up PR). You can also simply cherry-pick the second commit with the fix.


ayrton04 (approved):
Thanks for this!

ayrton04:
@Timple, would you mind fixing this conflict?

Timple:
Done

PR #662, opened 2021-05-19
Loosen IMU acceleration IO test

audrow:
This test was failing occasionally for me (Ubuntu 20.04 with Rolling), but not by much. In one place, the test is expecting a value less than 0.4, but is getting values, in my experience, that may be a little greater than 0.5. This PR increases the value to 0.6 so that the test doesn't fail as frequently.


ayrton04:
I'm a little concerned that these tests seem to do this more often in ROS 2 than in ROS 1. That doesn't prevent me from wanting to merge this. Just stating.

I plan to resume a more active role in maintenance of this package soon, so I'll be looking into that. In the meantime, thanks for the PR.

PR #663, opened 2021-05-20
Remove try-catch blocks around declare_parameter

jacobperron:
Try-catches were added in https://github.com/cra-ros-pkg/robot_localization/pull/631 due to a new rclcpp feature enforcing static parameters.
The behavior was later patched for this particular use-case in https://github.com/ros2/rclcpp/pull/1673, so now we can avoid having to try-catch.

---

Note, you'll need this change to compile against Galactic and newer.


wjwwood (approved):

SteveMacenski (approved):

SteveMacenski:
thanks!

PR #669, opened 2021-05-25
Made world_frame a ros param in navsat_transform_node

chriswsuarez:


ayrton04:
What's the use case here? If you specify `world_frame_id_` as a parameter, it's just going to get overwritten when you receive an odometry message:

https://github.com/cra-ros-pkg/robot_localization/blob/4f1665c48b7de6a5814a3e56efc3b4d0a6bd72f6/src/navsat_transform.cpp#L713

chriswsuarez:
When setting a manual datum and wait for datum is true it looks like it'll set that frame_id to "odom" always before it is overwritten by the odomcallback.  After digging through the code I'm not yet completely sure if that makes a huge difference but if the datum frame_id is used elsewhere then it would matter.

My specific use case is one where I'm using a nav stack that does a good job of localization and loop closure but I need to extract gps information from a robot without having a useful gps sensor.  I'm using navsat_transform node to generate simulated lat-lon based on an accurate datum and extrapolating from there.   I'm also launching another instance of navsat transform node to then convert information into UTM.  In my case the datum is vitally important.

ayrton04:
So this will only apply when using the manual datum? If so, we should be checking for that situation in the code. If the user specifies the `world_frame` and `use_manual_datum` is _not_ true, then we should warn the user and _not_ use their specified `world_frame` parameter.

chriswsuarez:
Currently, `world_frame_id_` is hard coded to "odom" until an Odom message is received. All this does is allow the user to specify what the `world_frame` is on initialization so that in case it is used anywhere before an Odom message is received, it will be correct with respect to the users setup.  In my brief debugging of the code the `world_frame_id_` is at least used in the manual datum as the frame_id before it has been reset to the correct value by the Odom callback.  If it is preferred to not have any user input for this variable I think it would be best to instead initialize `world_frame_id_` to an empty string and wait for an Odom message to be received before anything else is done with `world_frame_id_`.  Also upon more digging, this probably should also apply to `base_fame_id_`.  If it is preferred I can hold off on these pull requests and it can be fully fleshed out how the frame_ids are used before an Odom message is received but it has affected my setup because my robot's estimated location in lat-lon changes depending on whether `world_frame_id_` is hard coded to "odom" or whether I'm able to set a param to change it to "map".

chriswsuarez:
Just clearing up my last comment.  I don't know if there's any other cases that are affected by this change. It is possible it's only just the manual datum case but I'm not sure.  I just think the cleanest thing the do is either ask the user what the frame_id is before it is used (even though it will be rewritten by the Odom callback) or have it be empty and don't do anything with it until it is filled out by an Odom message.

ayrton04:
> in case it is used anywhere before an Odom message is received

But that should never be the case, unless you are using a manual datum.

Let's ignore the manual datum for a moment. Here are the places `world_frame_id_` is used.

https://github.com/cra-ros-pkg/robot_localization/blob/a61ee815f96d682d111b0c5ebff9f18199167f1f/src/navsat_transform.cpp#L324-L327
In this case, the check is [gated](https://github.com/cra-ros-pkg/robot_localization/blob/a61ee815f96d682d111b0c5ebff9f18199167f1f/src/navsat_transform.cpp#L237) by whether we've got the odom message.

https://github.com/cra-ros-pkg/robot_localization/blob/a61ee815f96d682d111b0c5ebff9f18199167f1f/src/navsat_transform.cpp#L368
This is the manual datum callback, so the value is being assigned.

https://github.com/cra-ros-pkg/robot_localization/blob/a61ee815f96d682d111b0c5ebff9f18199167f1f/src/navsat_transform.cpp#L457
This case is [gated](https://github.com/cra-ros-pkg/robot_localization/blob/a61ee815f96d682d111b0c5ebff9f18199167f1f/src/navsat_transform.cpp#L787) by `transform_good` _and_ `odom_updated_`. Again, this requires that we've received an odom message or set a manual datum.

https://github.com/cra-ros-pkg/robot_localization/blob/a61ee815f96d682d111b0c5ebff9f18199167f1f/src/navsat_transform.cpp#L569
Exactly the same [gating](https://github.com/cra-ros-pkg/robot_localization/blob/a61ee815f96d682d111b0c5ebff9f18199167f1f/src/navsat_transform.cpp#L787).

The bottom line is that the `world_frame_id_` is only ever accessed if it has been set by the manual datum or the odom callback. By exposing this parameter, you are creating an ambiguity for users: how do they know if they need to set the `world_frame_id`? If they are using the node 'normally,' I think the parameter creates confusion.

> If it is preferred to not have any user input for this variable I think it would be best to instead initialize world_frame_id_ to an empty string and wait for an Odom message to be received before anything else is done with world_frame_id_.

I agree that the `world_frame_id_` should default to the empty string. Can you help me find a code path _outside_ manual datum setting where we access `world_frame_id_` before the odom message has been received? From what I can see, this only solves a case where we set a manual datum _before_ we get an odom message. If that's the case, then I would like the node to print an error and exit if `use_manual_datum` is false and and `world_frame_id` has been set.

ayrton04:
Just checking if there's any intention of addressing the PR feedback. It's fine if not, but I'll close this out in that case.

chriswsuarez:
I began to take a look at fixing this the proper way but I don't think I have the bandwidth currently to finish it.  If anything I can fix it sometime in the future and submit a new PR.

RonaldEnsing:
I applied a similar modification so that I can choose to set the `world_frame_id` to `map` if `wait_for_datum` is `true`. I don't remember exactly why this was needed, I changed this a while ago: https://github.com/RonaldEnsing/robot_localization/commit/3d9fb8a5a16773c712215a685806c16990151b6c

PR #670, opened 2021-05-25
Revert 657 audrow/add qos overrides

audrow:
This change wasn't meant to make it into Galactic, as https://github.com/ros2/message_filters/pull/56 targets rolling.


SteveMacenski:
Thanks for helping sort that out - in the future it might be good to hold off API breaking changes for a new distribution until the one underway is released :wink:

PR #671, opened 2021-06-03
UKF cleanup

ayrton04:
FYI @jola6897. Nothing of note changed here. Just reorganized and renamed the method that carries out prediction/projection for each sigma point. Also fixed the failing tests.

EDIT: I rolled in changes to (hopefully) address the other items you brought up in https://github.com/cra-ros-pkg/robot_localization/issues/621.


test/test_ukf.cpp
```diff
@@ -57,9 +57,9 @@ class RosUkfPassThrough : public RosUkf
 TEST(UkfTest, Measurements)
 {
   std::vector<double> args;
-  args.push_back(0.001);
-  args.push_back(0);
-  args.push_back(2);
+  args.push_back(::sqrt(3));
```
ayrton04: These will return to normal in my next PR, which will also address the Euler angle issues.

ayrton04:
Closing and re-opening to get the tests to fire once more. The UKF tests seem to have some instability in them in other PRs, so I want to make sure that won't happen here.

PR #673, opened 2021-06-10
Fix bug in navsat_transform

kribe48:
When using the parameter "broadcast_cartesian_transform" as intended, the "broadcast_utm_transform" is declared twice.


ayrton04 (approved):

ayrton04:
Thanks!

Issue #674, opened 2021-06-11
Exception thrown from ros_filter.cpp during launch in ROS2 galactic


kribe48:
## Bug report

Exceptions are thrown when running code in ROS2 galactic. I believe the reason is due to the change in the declaration of parameters, which in galactic all are given a specific type. Parameters that are not set with default values can therefore not be used to check if parameter values have been provided.

One example is on line 822-823 in ros_filter.cpp:
this->declare_parameter<std::string>("tf_prefix"); 
if(this->get_parameter("tf_prefix", tf_prefix_path)) { ....

- Operating System:
Ubuntu 20.04
- Installation type:
Source
- Version or commit hash:
galactic

#### Steps to reproduce issue
Build the code and launch e.g. dual_ekf_navsat_example.launch.py

#### Expected behavior
No exceptions thrown

#### Actual behavior
Exception:  what():  expected [string] got [not set]

jacobperron:
Sorry, I introduced this bug in #663. See https://github.com/cra-ros-pkg/robot_localization/pull/675 for a fix.

PR #675, opened 2021-06-17
Fix runtime exception thrown when parameter override not provided

jacobperron:
Bug introduced in https://github.com/cra-ros-pkg/robot_localization/pull/663
Fixes https://github.com/cra-ros-pkg/robot_localization/issues/674

The templated declare_parameter method needs an override (or default value) since
it returns the actual parameter value, otherwise we get an exception.
To correctly declare a parameter without a default value, we should use a different
signature.

Related upstream issue: https://github.com/ros2/rclcpp/issues/1691


ayrton04 (approved):

ayrton04:
Thanks!

PR #676, opened 2021-06-23
Added FromLLArray service

MarcinSlomiany:
Service allows to quickly transform multiple ll points to ros coordinates system.


src/navsat_transform.cpp
```diff
@@ -448,6 +449,58 @@ bool NavSatTransform::fromLLCallback(
...
+      gps_local_cartesian_.Forward(
+        latitude,
+        longitude,
+        altitude,
+        cartesian_x,
+        cartesian_y,
+        cartesian_z);
+    } else {
+
+      navsat_conversions::LLtoUTM(
```
Timple: I thought I deprecated these functions in favor of geographilib. But I cannot confirm until next week.

srv/FromLLArray.srv
```diff
@@ -0,0 +1,3 @@
+geographic_msgs/GeoPoint[] ll_points
+---
+geometry_msgs/Point[] map_points
```
ayrton04: Missing a newline here.

src/navsat_transform.cpp
ayrton04: Yes, we should be consistent with whatever method we're using.

src/navsat_transform.cpp
```diff
@@ -448,6 +449,58 @@ bool NavSatTransform::fromLLCallback(
...
+        latitude,
+        longitude,
+        cartesian_y,
+        cartesian_x,
+        utm_zone_tmp);
+    }
+
+    cartesian_pose.setOrigin(tf2::Vector3(cartesian_x, cartesian_y, altitude));
+
+    // nav_msgs::msg::Odometry gps_odom;  // it's unused?
```
ayrton04: Please remove this line.

ayrton04 (changes requested):
Thanks for the PR.

I feel like there's a lot of repeat code between `fromLLArrayCallback` and `fromLLCallback`. Can we either make the former call the latter, or pull out the common logic into another method and have both callbacks call that method?

MarcinSlomiany:
> Thanks for the PR.
> 
> I feel like there's a lot of repeat code between `fromLLArrayCallback` and `fromLLCallback`. Can we either make the former call the latter, or pull out the common logic into another method and have both callbacks call that method?

Yes it's repeated `fromLLCallback`. I think that pulling out logic into another method would be better.

src/navsat_transform.cpp
Timple: Found it!

Looks like parts didn't make it into ROS2. The reasoning of using geographiclib directly instead of these navsat_transform functions is here: https://github.com/cra-ros-pkg/robot_localization/pull/627/files#r598612280

ayrton04:
Apologies for the ping, but is there any chance we'll address the PR feedback on this?

ayrton04:
Any update here? No worries if there's no interest in addressing the feedback, but I'll close the PR in that case.

pepisg:
Hi! I was not aware of this PR, I had opened #724 requesting exactly the same feature. I may pick this one up and address your comments soon, however I could not give you an specific time frame

MarcinSlomiany:
Hi, sorry for the long delay. I think next week I will start making changes based on your comments @ayrton04. Hope to finish it soon :)

ayrton04:
Just checking in on this again. As before, no rush.

ayrton04:
Ping! :)

MarcinSlomiany:
Hi, I think that before this week ends I'll start pushing some commits..

MarcinSlomiany:
Finally I've made update to this PR: https://github.com/piappl/robot_localization/commit/dc74221e0596b7a91b27c42b764d1037d8152d20

If this approach is fine I'll do the same for the **toLL** method. Also I'll get rid of an additional [commit](https://github.com/piappl/robot_localization/commit/cbcd769bbf82169ed3aea2dc3cb914c736950352) of my colleague.

src/navsat_transform.cpp
```diff
@@ -407,9 +408,38 @@ bool NavSatTransform::fromLLCallback(
...
+  }
+
+  return true;
+}
+
+bool NavSatTransform::fromLLArrayCallback(
+  const std::shared_ptr<robot_localization::srv::FromLLArray::Request> request,
+  std::shared_ptr<robot_localization::srv::FromLLArray::Response> response)
+{
+  for (auto it = request->ll_points.begin(); it != request->ll_points.end(); ++it) {
```
ayrton04, line 425: So the behavior here is not ideal, in that if one of the points throws an exception, the `response` object might be partially filled out. What we'd really want there is an "all or nothing" operation in which we return all points if they all succeed, otherwise we return an empty vector. 

In this case, we know that `fromLL` will only throw an exception if `!transform_good_`, so if any one point throws an exception, they all will. However, we don't want to assume we know how `fromLL` works internally here (and someone might change it later on). I might suggest something like this (note that I have not tested this):

```
#include <algorithm>
...

decltype(response->map_points) converted_points;
converted_points.reserve(request->ll_points.size());

try
{
  std::transform(
    request->ll_points.begin(),
    request->ll_points.end(),
    std::back_inserter(response->map_points),
    [] (const auto& point) { return fromLL(point); });
}
catch(...)
{
  return false;
}

response->map_points = std::move(converted_points);
```

ayrton04 (approved):
I'm going to approve this, but I'd prefer that we make the change I suggested (though you are free to fix it however you like; you don't have to use the code block I suggested).

YossiO-BWR:
Is this expected to be merged at some point? I almost came to implement this all over :)
If you want to improve something before merging - I can lend a hand.

ayrton04:
See my comment above. If the change I suggested is implemented, then I will happily merge.

YossiO-BWR:
@MarcinSlomiany are you up to it? Want to give me write permissions to your fork or should I copy your changes as a base to my fork?

MarcinSlomiany:
Hi @YossiO-BWR! I can't give you write permissions becaus it's my company repo but ~~this weekend~~ _next week_ 😅 I'll implement changes suggested by @ayrton04

YossiO-BWR:
@MarcinSlomiany Understandable, lookg forward seeing this merged!

iandresolares:
Hi!
In what branch is this available? Thanks!

ayrton04:
Not available anywhere in this repo yet. The PR has yet to be merged.

JayHerpin:
Not sure the proper way to handle, but I have updated this for jazzy with the requested changes here: https://github.com/Metal-Shark-Sharktech/robot_localization/tree/from-ll-array-service

I can open a new pull request if desired since the original author seems to have moved on.

ayrton04:
Yes, @JayHerpin, if you could make a PR for Jazzy and Rolling, I'd appreciate it!

JayHerpin:
This can be closed now. However while https://github.com/cra-ros-pkg/robot_localization/pull/912 was landed on jazzy, I dont have a rolling workspace setup to test there.. so that still needs to get done

PR #682, opened 2021-07-16
Prevent node from crashing on invalid UTM zone

MCFurry:
GeographicLib throws an error when a gps message is received and the current set UTM frame is invalid (i.e. too far away).
This is a valid error of course, but it also makes the node crash.
Now the error is catched and the user is warned about this.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -415,7 +415,15 @@ namespace RobotLocalization
     {
       int zone_tmp;
       bool nortp_tmp;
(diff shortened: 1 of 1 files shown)


ayrton04:
LGTM, thank you!

MCFurry:
Thanks for merging. I might create a similar try-catch-fix for ROS2, but I'm not sure which branch to use there?
It seems that the conversion to geographiclib is done here: https://github.com/cra-ros-pkg/robot_localization/pull/643
but didn't end up in the [ROS2](https://github.com/cra-ros-pkg/robot_localization/tree/ros2) branch at the moment?

PR #683, opened 2021-07-16
Prevent node from crashing on invalid UTM zone

MCFurry:
https://github.com/cra-ros-pkg/robot_localization/pull/682 but also for melodic user.


Timple:
Nice catch! Might be nice to get bloomed before the next melodic sync :grin: 
:angel:

ayrton04:
https://github.com/ros/rosdistro/pull/30323
https://github.com/ros/rosdistro/pull/30324

PR #684, opened 2021-07-17
changed geographiclib to <depend> tag

VAM7686:
Currently, robot_localization only pulls in the GeographicLib package as a build dependency and only installs `libgeographic19` which gives us the shared library files, but not the headers. 

Since #626 added headers from GeographicLib the tags for GeographicLib in package.xml need to be changed from ```<build_depend>``` to ```<depend>```

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -19,6 +19,7 @@
   <depend>diagnostic_updater</depend>
   <depend>eigen</depend>
   <depend>eigen_conversions</depend>
+  <depend>geographiclib</depend>
   <depend>geographic_msgs</depend>
   <depend>geometry_msgs</depend>
   <depend>message_filters</depend>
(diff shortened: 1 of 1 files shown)


ayrton04:
So changing it to a `depend` installs the headers? We don't need to change the dependency to `geographiclib-dev` or anything?

ayrton04:
Merging, since this won't hurt anything regardless of the answer to my question.

VAM7686:
> So changing it to a `depend` installs the headers? We don't need to change the dependency to `geographiclib-dev` or anything?

The `<build_depend>` tag only installs the dependency to build the package and not for execution time. Changing it to `<depend>` tags is the same as specifying `<build_depend>`, `<build_export_depend>`, and `<exec_depend>` for the dependency. Doing so installs the dependency to be used at build time and execution time. More information about the dependency tags can be found at [REP 149](https://ros.org/reps/rep-0149.html#dependency-tags).

PR #689, opened 2021-08-09
Moved GeographicLib from build_depend to depend

### LINKED ISSUE #687: GeographicLib dependency is not exported by CMakeLists.txt
There are inline utility functions in the navsat_conversions.h file, for example,  that can be called from a package that depends on robot_localization.  However, this results in a linker error.  The workaround is for the dependent package to find_package() GeographicLib and add it as a target link library, but this is really an implementation detail of robot_localization. Instead, I believe robot_localization should export GeographicLib.

### LINKED ISSUE #688: missing geographiclib dependency in package.xml for melodic
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  -  Binaries from Apt
- Version or commit hash:
  - From apt: 2.6.11-1bionic.20210723.234509

#### Steps to reproduce issue
Install using sudo apt install ros-melodic-robot-localization

#### Expected behavior
libgeographic-dev is installed as a dependency 
#### Actual behavior
libgeographic-dev is not installed as a dependency 
#### Additional information
geographiclib is listed in the package.xml for the noetic release which solves the problem there but was not included in the melodic release

joshjowen:
This ensures that GeographicLib is installed when the apt package is used which is required when GeographicLib is used as a shared library which fixes #688. It was also added to the depends line in CMakelists.txt which should fix #687 (It does on my systems i.e melodic on Ubuntu 18.04)

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -20,6 +20,7 @@
   <depend>eigen</depend>
   <depend>eigen_conversions</depend>
   <depend>geographic_msgs</depend>
+  <depend>geographiclib</depend>
   <depend>geometry_msgs</depend>
   <depend>message_filters</depend>
   <depend>nav_msgs</depend>
@@ -32,7 +33,6 @@
   <depend>tf2_geometry_msgs</depend>
   <depend>tf2_ros</depend>
   <depend>yaml-cpp</depend>
-  <depend>geographiclib</depend>
 
   <build_depend>message_generation</build_depend>
   <build_depend condition="$ROS_PYTHON_VERSION == 2">python-catkin-pkg</build_depend>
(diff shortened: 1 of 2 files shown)


package.xml
```diff
@@ -32,8 +32,8 @@
   <depend>tf2_geometry_msgs</depend>
   <depend>tf2_ros</depend>
   <depend>yaml-cpp</depend>
+  <depend>geographiclib</depend>
```
Timple: The rest of this list seems sorted :slightly_smiling_face:

joshjowen: Fixed

Timple:
Seems like a proper fix!

_Not a maintainer here, but I did the UTM implementation of Geographiclib (was already a dependency then)._

ayrton04 (approved):

PR #705, opened 2021-11-16
Propagate the suppression of tf warnings

caguero:
This patch propagates the suppression of `tf` warnings to all places where there's a `lookupTransformSafe()`.
Signed-off-by: Carlos Agüero <caguero@openrobotics.org>

### DIFF
--- a/include/robot_localization/navsat_transform.h
+++ b/include/robot_localization/navsat_transform.h
@@ -287,6 +287,10 @@ class NavSatTransform
     //!
     ros::Duration transform_timeout_;
 
+    //! @brief When true, do not print warnings for tf lookup failures.
+    //!
+    bool tf_silent_failure_;
+
     //! @brief timer calling periodicUpdate
     //!
     ros::Timer periodicUpdateTimer_;
--- a/include/robot_localization/navsat_transform.h
+++ b/include/robot_localization/navsat_transform.h
@@ -210,6 +210,10 @@ class NavSatTransform
     //!
     bool use_local_cartesian_;
 
+    //! @brief When true, do not print warnings for tf lookup failures.
+    //!
+    bool tf_silent_failure_;
+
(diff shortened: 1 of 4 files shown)


ayrton04:
Thanks for this. Test failures are unrelated to this change, so merging.

include/robot_localization/navsat_transform.h
```diff
@@ -287,6 +287,10 @@ class NavSatTransform
     //!
     ros::Duration transform_timeout_;
 
+    //! @brief When true, do not print warnings for tf lookup failures.
+    //!
+    bool tf_silent_failure_;
```
ayrton04: Actually, sorry, small request: can we put this up with the other booleans? My original aim here was to group variables by data type and _then_ alphabetically, though I realise there are some variables that don't adhere to that standard here.

include/robot_localization/ros_filter.h
```diff
@@ -638,6 +638,10 @@ template<class T> class RosFilter
     //!
     ros::Duration tfTimeout_;
 
+    //! @brief When true, do not print warnings for tf lookup failures.
+    //!
+    bool tfSilentFailure_;
```
ayrton04: Same thing here, if you don't mind.

include/robot_localization/navsat_transform.h
caguero: Totally, done in 8c435d5.

include/robot_localization/ros_filter.h
caguero: Done in 8c435d5.

caguero:
@ayrton04, I think this should be good to go. Do you mind to review it again? Thanks!

ayrton04 (approved):
Sorry, @caguero, lots going on just now. LGTM, thanks!

Issue #711, opened 2022-01-05
UKF, missing addition with matrix R in the covariance correction update


lkdo:
In function Ukf::correct,
  - the Kalman gain is computed correctly using the (P_zz + R)^-1
  - and in the covariance update, the P_zz must similarly be added with R i.e. 
        P = P - (K * (P_zz + R) * K')  instead of P = P - (K * P_zz * K')

ayrton04:
I believe you are correct, yes. The formulation within the [original paper](https://groups.seas.harvard.edu/courses/cs281/papers/unscented.pdf) seems to ignore the measurement covariance at all, whereas the formulation on [Wikipedia](https://en.wikipedia.org/wiki/Kalman_filter#Unscented_Kalman_filter) includes it.

![image](https://user-images.githubusercontent.com/5042191/148285526-89c4f4f0-48b4-4276-976a-42c68ce7c935.png)

![image](https://user-images.githubusercontent.com/5042191/148285589-9ee6315e-9ae3-4e25-a2ac-a0a175e8ddbb.png)

In any case, including `R` as you stated is clearly correct. I need to get the UKF cleanup PR merged, and then if you have the cycles, I'd welcome a PR to fix this. Thanks for pointing it out!

lkdo:
Thanks. Sure, I'll do a PR after the merge.

ayrton04:
Sorry for the delay. The PR is merged into `noetic-devel`. If you wouldn't mind submitting your PR to that branch, I can make sure it finds its way to the ROS 2 branches. I'm reluctant to push all this back to melodic, but could be convinced otherwise.

ayrton04:
And thank you!

roncapat:
Any updates on this?

lkdo:
Hi.

The last news on this was that one of the unit tests is breaking as a result of the change. 

I don't really understand why, so I tried a small change in the unit test UKF ImuDifferentialIO  ( which is the one that fails as far as I can see):

Changing: 
   ros::Publisher imuPub = nh.advertise<sensor_msgs::Imu>("/imu_input3", 5);
 to 
   ros::Publisher imuPub = nh.advertise<sensor_msgs::Imu>("/imu_input0", 5);
 
Makes the test suceed. 
 
Both topics have the following config: [false, false, false, true, true, true, false, false, false, false, false, false, false, false, false]
 
[...]

Issue #690, opened 2021-08-09
Control term is not accounting for time delta


improve100:
TM: editing issue for clarity.

`prepareControl` doesn't appear to be using the `predictionDelta` term to compute an acceleration from the error between the controlled velocity and the current velocity.

https://github.com/cra-ros-pkg/robot_localization/blob/da0f9676c27db71b9e7cc76fbb227c92202ffd49/src/filter_base.cpp#L351-L375

Should there be a `controlAcceleration_(controlInd) /= predictionDelta;` line after line 371?

ayrton04:
Yes, I think you're right. The [original code](https://github.com/cra-ros-pkg/robot_localization/commit/6e83ef2f3635fc84efc2413db0fc05e294c4d07e#diff-b96ffbfcfc303ae826323830be2c943272c096bf8cb93fdb56fef4c949a19ce2R283) was multiplying the error term by a frequency, and that frequency was computed as `1.0 / predictionDelta`:

https://github.com/cra-ros-pkg/robot_localization/commit/6e83ef2f3635fc84efc2413db0fc05e294c4d07e#diff-8a83a6a31c3794ee91112354ac3dcceaa70fb760089c851e5998dcc186dbaf65R329
https://github.com/cra-ros-pkg/robot_localization/commit/6e83ef2f3635fc84efc2413db0fc05e294c4d07e#diff-b96ffbfcfc303ae826323830be2c943272c096bf8cb93fdb56fef4c949a19ce2R283

But then we capped the acceleration with a gain term instead, and lost the time delta term. However, I don't think your proposed fix will work either. The acceleration limits get applied in `computeControlAcceleration`, so we need to pass the `predictionDelta` into that method so we can compute the final acceleration value within it, and then cap it. In other words, I think [this line](https://github.com/cra-ros-pkg/robot_localization/blob/da0f9676c27db71b9e7cc76fbb227c92202ffd49/include/robot_localization/filter_base.h#L360) needs to be:

`const double finalAccel = std::min(std::max(gain * error / predictionDelta, -limit), limit);`

But in order to do that, you need to pass `predictionDelta` into `computeControlAcceleration`, preferably as the third argument.

ayrton04:
Care to submit a PR?

Issue #612, opened 2020-12-01
UKF control of state vector missing?


wienans:
Hey Guys,

i was wondering if the ukf doesn't support control of the state vector or if it isn't implemented completely. 
In ukf.ccp the function 
`prepareControl(referenceTime, delta);`
is called which writes to the variable `controlAcceleration_`
but `controlAcceleration_` isn't used later in the code, or i can't find the position where the control is added.
in the ekf.cpp there is a block directly before the state transfer 
 ```
// (1) Apply control terms, which are actually accelerations
    state_(StateMemberVroll) += controlAcceleration_(ControlMemberVroll) * delta;
    state_(StateMemberVpitch) += controlAcceleration_(ControlMemberVpitch) * delta;
    state_(StateMemberVyaw) += controlAcceleration_(ControlMemberVyaw) * delta;
 state_(StateMemberAx) = (controlUpdateVector_[ControlMemberVx] ?
      controlAcceleration_(ControlMemberVx) : state_(StateMemberAx));
    state_(StateMemberAy) = (controlUpdateVector_[ControlMemberVy] ?
      controlAcceleration_(ControlMemberVy) : state_(StateMemberAy));
    state_(StateMemberAz) = (controlUpdateVector_[ControlMemberVz] ?
      controlAcceleration_(ControlMemberVz) : state_(StateMemberAz));
```

   

Shouldn't the ukf.cpp include something similar to this too?
Thanks for help

ayrton04:
Yeah, it should do something like that, though not exactly the same, because prediction is done very differently in the UKF. I'll mark this as a bug. Thanks!

lkdo:
Hi, 

What does it mean that prediction is done very differently in the UKF ?

What am I missing ? To account for the control term, isn't it as simple as below ?

For each sigma point prediction (taking into account the fixes of  https://github.com/cra-ros-pkg/robot_localization/issues/621 ),

void Ukf::projectSigmaPoint(Eigen::VectorXd& sigmaPoint, double delta)
{
 ...

    sigmaPoint(StateMemberVroll) += controlAcceleration_(ControlMemberVroll) * delta;
    sigmaPoint(StateMemberVpitch) += controlAcceleration_(ControlMemberVpitch) * delta;
    sigmaPoint(StateMemberVyaw) += controlAcceleration_(ControlMemberVyaw) * delta;
  
    sigmaPoint(StateMemberAx) = (controlUpdateVector_[ControlMemberVx] ?
[...]

PR #717, opened 2022-01-17
Updates to UKF 

lkdo:


lkdo:
Some test cases fail, but I'm not sure they are relevant !
#612, #690, #711

PR #718, opened 2022-01-17
Adding matrix R in UKF covariance update in correct (711)

lkdo:


lkdo:
@ayrton04 
Hi,

These changes trigger a regression test failure, 
but since this is a modification of the actual numerics of the filter,
I belive this is ok, and that the regression test must be modified.

Please let me know how to proceed.

best regards,

ayrton04:
Hmmm, I'd want to track down why it's different. If you look at the failing tests, you can see that the error between what we're expecting from the integration and what we're seeing has grown considerably:

https://build.ros.org/job/Npr__robot_localization__ubuntu_focal_amd64/68/testReport/junit/(root)/InterfacesTest/ImuDifferentialIO/

https://github.com/lkdo/robot_localization/blob/e6d56e024b9d5b02d562dcb6ddf8606f53726da1/test/test_ukf_localization_node_interfaces.cpp#L1061-L1063

It may be the case that the changes I made to get these tests to work after I updated the UKF need to be put back.

https://github.com/lkdo/robot_localization/commit/0fe3097f58b4c7f7645aba20e0e675468182e707#diff-7f0519215d4ab3adad9e1164ae9d97e87a8d700152dbfe4bd85108301db448b7R49-R54

https://github.com/lkdo/robot_localization/commit/0fe3097f58b4c7f7645aba20e0e675468182e707#diff-c484b9dfe684344cfebf71c5b5619f237c50e872c2f520d5300698333efc5426R986-R997

Note that the same limits are used for the EKF.

https://github.com/lkdo/robot_localization/blob/e6d56e024b9d5b02d562dcb6ddf8606f53726da1/test/test_ekf_localization_node_interfaces.cpp#L1057-L1059

So we definitely need to make sure we're not doing something erroneous.

ayrton04:
I'm having a hard time finding our conversation about `NaN` values creeping into the UKF estimate. Did you ever manage to re-create the same issue with the upstream code?

ayrton04:
Just checking in on this ticket.

PR #721, opened 2022-01-24
Fix unused-parameter warning

ardabbour:
When building a package that uses this header, the compiler throws this unused parameter warning (and rightly so). This solution is probably incompatiable with Windows, so if intercompatiability is an issue, perhaps [[maybe_unused]] should be used instead, but that requires C++17: [More here](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)

### DIFF
--- a/include/robot_localization/navsat_conversions.h
+++ b/include/robot_localization/navsat_conversions.h
@@ -185,7 +185,7 @@ static inline void LLtoUTM(const double Lat, const double Long,
  */
 static inline void UTMtoLL(const double UTMNorthing, const double UTMEasting,
                            const std::string &UTMZone, double& Lat, double& Long,
-                           double& /*unused*/)
+                           double& /*gamma*/)
 {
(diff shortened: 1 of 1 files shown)


Timple:
Otherwise 
```c++
double& Long, double& /*gamma*/)
```
 might be an option as well.

SteveMacenski:
@Timple's suggestion is more in line with standard practice, if you wouldn't mind making that update? You'd do that in the cpp file, not the header file.

ardabbour:
There is no `navsat_conversions.cpp` seperate from the header (or I'm missing something), but I've changed it to the standard practice; should do the trick

Timple:
Nitpick: the comment already suggests that the code is unused. Leaving the word `gamma` also indicates _what_ is unused.

ayrton04 (approved):

ayrton04:
Thanks!

PR #723, opened 2022-01-26
Update documentation

scheunemann:
Updating the readme link to point to the noetic documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,4 +3,4 @@ robot_localization
 
 robot_localization is a package of nonlinear state estimation nodes. The package was developed by Charles River Analytics, Inc.
 
-Please see documentation here: http://docs.ros.org/melodic/api/robot_localization/html/index.html
+Please see documentation here: http://docs.ros.org/noetic/api/robot_localization/html/index.html


Timple:
You might as well leave the distro out?
http://docs.ros.org/en/api/robot_localization/html/index.html

scheunemann:
@Timple yes good idea. However, I think it is quite nice that the documentation points to the latest documentation of the development branch. For example, if melodic were to point to your link, users may think, for example, they can use [the newly introduced](https://github.com/cra-ros-pkg/robot_localization/commit/660793f4) parameter `permit_corrected_publication`, while this was only introduced for `noetic` ([`melodic`](https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ros_filter.cpp) vs. [`noetic`](https://github.com/cra-ros-pkg/robot_localization/blob/noetic-devel/src/ros_filter.cpp)). 

Let me know what you think and/or if that is an unintended feature in the first place, I'll be happy to update the link.

ayrton04 (approved):
Works for me. Thanks!

PR #725, opened 2022-02-15
Moving to C++17 support

ayrton04:
Trying to address https://build.ros2.org/job/Rbin_uJ64__robot_localization__ubuntu_jammy_amd64__binary/10/consoleFull

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(robot_localization)
 
 if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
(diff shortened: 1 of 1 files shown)


SteveMacenski:
If you don't specify anything, C++17 is now default, so you wouldn't need to specifically call it out. But not bad to specify anyway

SteveMacenski (approved):

ayrton04:
Ah, maybe that's why our build broke when (I assume) most people's didn't (that is, because we were specifying C++14).

SteveMacenski:
Likely the case. I'm not sure when rolling will move to C++17, but from the TSC meetings it seems to be happening soon, so that's why I think that. I'm not 100% sure but the evidence seems to point to it. If you make this change and it doesn't fix the build, ping me and I'll take a look

ayrton04:
I have a PR in to the rosdistro to enable PR testing on this branch. Once that gets merged, I'll close/open this to kick it off again.

Issue #726, opened 2022-02-15
Child Frame of Odometry Messages is ignored


frygge:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binary and source
- Version or commit hash:
  - git rev-parse HEAD b5345749e1449a5ec3536d3c8cc80f1de65ae9f6

#### Steps to reproduce issue
- measure an odometry in a sensor frame that is rotated by (for example) 90 degree
- process this odometry as odom0 (and as single measurement source) in robot_localization as odom0

#### Expected behavior
- the robot should run forward

#### Actual behavior
- the robot is running sideways

#### Additional information
- I could fix it already and would like to open a PR

ayrton04:
Please ask this as a question in answers.ros.org or robotics.stackexchange.com. When you do, please include your full configuration and a sample input message from every sensor input. I'm sceptical that there is a bug here, but regardless, this process should start with a question on one of those forums. Thanks.

frygge:
Hmm, I could do that. Still, please give it a thought since I read the code and can see the error there:
- odometryCallback(..) calls poseCallback and misses to send the childFrameID (~line 1744)
- poseCallback calls preparePose but the child Frame is gone already (~line 1810)
- in differential mode, preparePose calls prepareTwist but the message gets the baseLinkFrame (since it doesn't know the right one) (~line 2868 and 2894)
- Hence, the correction of the child frame is just wrong (the identity) (~line 3041)

The line numbers might be off by a little since I fixed it already (my fix is basically just passing the childFrame)

Do you still think, I should open a question somewhere? Then I would do it...

ayrton04:
I will need to read the code, but aren't we transforming the pose data into the target frame (e.g., `odom`) _before_ we differentiate it? If so, the correct child frame would be `base_link`.

ayrton04:
Oh, you're measuring pose data from a sensor that is on the robot and has a non-zero transform from the robot origin. This is not a use case that is well supported by the package. Pose data needs to have a direct transform from the world frame of that pose data to your world frame (so `odom`->`your_world_frame`). If you are measuring pose data from a sensor with a transform relative to `base_link`, it doesn't behave well. You'd need to have something that is _transforming_ your `base_link`->`sensor` transform into the world frame, then applying that transform to the pose measurement.

You can hold off on asking the question. I'll reopen this and look into what you're saying, but I am concerned that we'll break other use cases if we change that.

[...]

PR #728, opened 2022-02-17
Fix odometry msgs with child frame other than baseLink

frygge:
This PR addresses #726

It fixes the bug that ignores the child frame of odometry messages. This is relevant for processing odometry messages in differential mode (since there, it computes the differential in the child frame, hence it needs to know its name) and odometry poses in relative mode.

### DIFF
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -245,11 +245,13 @@ template<class T> class RosFilter
     //! @param[in] msg - The ROS stamped pose with covariance message to take in
     //! @param[in] callbackData - Relevant static callback data
     //! @param[in] targetFrame - The target frame_id into which to transform the data
+    //! @param[in] childFrame - The source frame_id from which to transform the data
     //! @param[in] imuData - Whether this data comes from an IMU
     //!
     void poseCallback(const geometry_msgs::PoseWithCovarianceStamped::ConstPtr &msg,
                       const CallbackData &callbackData,
                       const std::string &targetFrame,
+                      const std::string &childFrame,
                       const bool imuData);
 
     //! @brief Callback method for manually setting/resetting the internal pose estimate
@@ -384,6 +386,7 @@ template<class T> class RosFilter
     //! @param[in] msg - The pose message to prepare
     //! @param[in] topicName - The name of the topic over which this message was received
     //! @param[in] targetFrame - The target tf frame
+    //! @param[in] childFrame - The source tf frame
     //! @param[in] differential - Whether we're carrying out differential integration
     //! @param[in] relative - Whether this measurement is processed relative to the first
     //! @param[in] imuData - Whether this measurement is from an IMU
@@ -395,6 +398,7 @@ template<class T> class RosFilter
     bool preparePose(const geometry_msgs::PoseWithCovarianceStamped::ConstPtr &msg,
                      const std::string &topicName,
                      const std::string &targetFrame,
+                     const std::string &childFrame,
                      const bool differential,
                      const bool relative,
                      const bool imuData,
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -82,14 +82,14 @@ struct CallbackData
                const int updateSum,
                const bool differential,
                const bool relative,
-               const bool diff_rel_use_child_frame,
+               const bool pose_use_child_frame,
                const double rejectionThreshold) :
     topicName_(topicName),
     updateVector_(updateVector),
     updateSum_(updateSum),
     differential_(differential),
     relative_(relative),
-    diff_rel_use_child_frame_(diff_rel_use_child_frame),
+    pose_use_child_frame_( pose_use_child_frame),
     rejectionThreshold_(rejectionThreshold)
   {
   }
@@ -99,7 +99,7 @@ struct CallbackData
   int updateSum_;
   bool differential_;
   bool relative_;
-  bool diff_rel_use_child_frame_;
+  bool pose_use_child_frame_;
   double rejectionThreshold_;
 };
 
@@ -248,13 +248,13 @@ template<class T> class RosFilter
     //! @param[in] msg - The ROS stamped pose with covariance message to take in
     //! @param[in] callbackData - Relevant static callback data
     //! @param[in] targetFrame - The target frame_id into which to transform the data
-    //! @param[in] childFrame - The source frame_id from which to transform the data
+    //! @param[in] poseSourceFrame - The source frame_id from which to transform the data
     //! @param[in] imuData - Whether this data comes from an IMU
     //!
     void poseCallback(const geometry_msgs::PoseWithCovarianceStamped::ConstPtr &msg,
                       const CallbackData &callbackData,
                       const std::string &targetFrame,
-                      const std::string &childFrame,
+                      const std::string &poseSourceFrame,
                       const bool imuData);
 
     //! @brief Callback method for manually setting/resetting the internal pose estimate
@@ -389,7 +389,7 @@ template<class T> class RosFilter
     //! @param[in] msg - The pose message to prepare
     //! @param[in] topicName - The name of the topic over which this message was received
     //! @param[in] targetFrame - The target tf frame
-    //! @param[in] childFrame - The source tf frame
+    //! @param[in] sourceFrame - The source tf frame
     //! @param[in] differential - Whether we're carrying out differential integration
     //! @param[in] relative - Whether this measurement is processed relative to the first
     //! @param[in] imuData - Whether this measurement is from an IMU
@@ -401,7 +401,7 @@ template<class T> class RosFilter
     bool preparePose(const geometry_msgs::PoseWithCovarianceStamped::ConstPtr &msg,
                      const std::string &topicName,
                      const std::string &targetFrame,
-                     const std::string &childFrame,
+                     const std::string &sourceFrame,
                      const bool differential,
                      const bool relative,
                      const bool imuData,
--- a/include/robot_localization/ros_filter.h
+++ b/include/robot_localization/ros_filter.h
@@ -89,7 +89,7 @@ struct CallbackData
     updateSum_(updateSum),
     differential_(differential),
     relative_(relative),
-    pose_use_child_frame_( pose_use_child_frame),
+    pose_use_child_frame_(pose_use_child_frame),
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
LGTM

ayrton04:
Moving the conversation from #726 over here.

First, thank you for the PR and the conversation in the original ticket.

I get what you're attempting to do here (and it's clever), but I can't help feeling that we're abusing the frame IDs in messages a bit. It also changes the filter behaviour: if you enable differential mode on a `nav_msgs/Odometry` input but your `child_frame_id` is filled out, it will always use that. I'm trying to determine if that will be a problem.

Moreover, anyone who has their transform tree set up such that the `frame_id` of the `nav_msgs/Odometry` input is set to the sensor's frame could see a change in behaviour.

Having said that, I like having this as a solution to pose data being measured from a sensor that is onboard the vehicle and reported in a frame that does not align with the robot's world frame.

I wonder if it makes sense to gate the behaviour behind a parameter that is only relevant for `nav_msgs/Odometry` inputs? Something like `differential_use_child_frame` or something?

Just for my own curiosity, what happens if you don't use your code change, but change the `frame_id` of your message to be `alb_sensor_frame`? Technically, the data from your sensor is _not_ reported in the `odom` frame, but if we're not using it directly, it doesn't matter.

SteveMacenski:
I think adding a parameter would be a reasonable move given what you describe above on differential mode.

frygge:
Sorry for the delay. I agree with you both, There might be use cases that break with this patch; thus, a gating parameter is a good approach.

Since the PR solves a similar problem with `relative` enabled, I'd rather propose a parameter name such as `pose_from_child_frame`, `pose_use_child_frame`, `diff_rel_use_child_frame`, or something alike. What do you think?

@ayrton04 I changed the `header.frame_id` to `alb_sensor_frame` in the bag file with a short python script. The differential use case did not show any changes (it's still going sideways). The relative use case now does not work at all since it can't find the transformation between `alb_sensor_frame` and `odom`.

ayrton04:
Never apologise to me for taking a long time to respond! I am notorious for it here.

I am really happy about this PR, as it enables the package to support a use case that it didn't support previously. I like `diff_rel_use_child_frame` the most. Naming parameters with very specific purposes is hard.

In any case, thank you for the PR!

ayrton04:
Zero rush, just checking in on this. Will you be willing/able to add the parameter in question?

frygge:
Yes, sure. I'm happy to finish it as we agreed. I just have an urgent paper deadline ahead for next week monday :)

frygge:
Aiming at writing easy-to-read code for better maintainability, I could set the child_frame parameter in the _pose-callback_ functions, i. e.
* on `geometry_msgs::PoseWithCovarianceStamped`, the callback passes the child frame `baseLinkFrameId_` (as it was before)
* on `sensor_msgs::Imu`, the callback passes the child frame `baseLinkFrameId_` (as it was before; only relevant if `differential` is enabled since there is a `imuData` flag also passed through, which changes the logic)
* on `nav_msgs::Odometry`, the callback either passes the child frame `baseLinkFrameId_` or `header.child_frame_id` depending on a parameter.

The last point yields that the newly introduced parameter also changes the child frame without enabling `differential` or `relative`. It seems to me that this is a) the easier implementation and b) a more consistent implementation. In that case, I would suggest a parameter `odomN_pose_uses_child_frame`, which also is easier to read while precisely describing what it does.

Implementing the new behavior only for `differential` and `relative` odometry messages includes a lot of `if-else` conditions, which makes the code much harder to read and maintain.

What do you think about this?

frygge:
I implemented the proposal and compared the behavior with the parameter enabled and disabled.
* with `odom0_diff_rel_use_child_frame=False`, everything looks as without the patch
* with `odom0_diff_rel_use_child_frame=True`, the new feature looks just fine

I found a bug in my previous implementation: it always interpreted the pose from `child_frame_id` to `header.frame` when differential and relative was disabled, which could break certain use cases. With the way of gating the behavior as described in the previous post, this bug is fixed.

Now, the parameter `odomN_pose_uses_child_frame` (or something alike) makes even more sense than `diff_rel_use_child_frame`. I'd change that if you'd agree.

frygge:
FYI: I posted a [related question on stackoverflow](https://stackoverflow.com/q/71172940/5296179) regarding the interpretation of the child_frame. No one seems to know the answer. So I took the freedom to link this pull request there since it basically creates the answer "You're free to choose." :)

src/ros_filter.cpp
```diff
@@ -1065,7 +1066,11 @@ namespace RobotLocalization
           relative = false;
         }
 
-        std::string odomTopic;
+        // Consider odometry transformation from the child_frame_id instead of the base_link_frame_id
```
ayrton04, line 1069: Might be worth checking if `differential` or `relative` is false and `diff_rel_use_child_frame` is true and warning the user if so.

include/robot_localization/ros_filter.h
```diff
@@ -245,11 +248,13 @@ template<class T> class RosFilter
     //! @param[in] msg - The ROS stamped pose with covariance message to take in
     //! @param[in] callbackData - Relevant static callback data
     //! @param[in] targetFrame - The target frame_id into which to transform the data
+    //! @param[in] childFrame - The source frame_id from which to transform the data
```
ayrton04: I'm not crazy about how the pose preparation method is now concerning itself with variables that are used for velocity computation, but the way I designed this doesn't really give us a better choice.

However, I'd like for this to be called `twistSourceFrame` everywhere, if possible, and for the doxygen to read "If using differential or relative mode, the twist data's source frame will be set to this value."

src/ros_filter.cpp
```diff
@@ -2903,6 +2921,24 @@ namespace RobotLocalization
       }
       else
       {
+        // make pose refer to the baseLinkFrame as source
+        if (childFrame != baseLinkFrameId_)
```
ayrton04: Are we sure this will work (as in, was it tested?). I feel like we should apply this transform _after_ line 2951, but the math may work out the same.

Regardless, I really only want this logic to apply to cases where `differential` or `relative` are true (as the parameter name implies), so this should go inside the `if (relative)` block below. As it's written now, this logic will apply even if `differential` and `relative` are both false.

ayrton04 (changes requested):
Thanks for the PR! Just a couple things I want to iron out.

src/ros_filter.cpp
frygge: Hey Tom, thanks for your feedback.

What about the bug I found in my previous implementation (see [this](https://github.com/cra-ros-pkg/robot_localization/pull/728#issuecomment-1062673239) comment)? So, the same problem with the "wrong" child frame also exists without `relative` or `differential` mode. That's why, I proposed to extend the new (parameterized) behavior to all three cases (normal, `relative`, and `differential`). That makes more sense to me and it makes the implementation (as it is now) much easier. What do you think about that?

If we agree to implement the "use other child frame for pose" behavior (behind a parameter), then the test above actually does the right thing (i also tested it): whenever the parameter **is not set**, the `childFrame` is going to be `baseLinkFrameId_`, and hence the program ignores the new code; whenever the parameter **is set**, we need to correct the transformation in both cases (normal and `relative`).

include/robot_localization/ros_filter.h
frygge: OK, I see your point. Please first read the comment in your feedback below.

....

So, what do you think about `poseSourceFrame` (in case you agree to applying the new behavior to the pose without `relative` and `differential` as well)?

src/ros_filter.cpp
frygge: If we only apply the new behavior to `relative` and `differential` mode, you're right. In this case, I will change it as suggested. It would help the user find inconsistencies in his configuration.

If we apply the new behavior to the *normal* mode as well, I rather propose to rename `diff_rel_use_child_frame` to `pose_use_child_frame`. Then, setting the new parameter to `true` even makes sense without setting `relative` or `differential`.

src/ros_filter.cpp
ayrton04: OK, then I guess we just need to make sure the parameters and variable names are sensible, as discussed above. The most critical thing is that the behavior is unchanged when the parameter is false.

frygge: Alright, so I'll rename the parameter to `pose_use_child_frame` and double-check that the old behavior is not touched when the parameter is unset.

include/robot_localization/ros_filter.h
ayrton04: Also, renaming this would be great, thanks. Your name sounds fine.

src/ros_filter.cpp
frygge: Done.

src/ros_filter.cpp
frygge: Done.
I double checked all settings: normal, diff, rel in combination with an odometry topic having base_link as the pose's source and another frame as the pose's source. Each combination worked as expected. In particular, the old behavior is untouched.

include/robot_localization/ros_filter.h
frygge: Done.

frygge:
The debian based pipeline failed with the following error message:

`08:27:52 time="2022-04-12T06:30:47Z" level=error msg="error waiting for container: unexpected EOF"`

This seems to be no bug related to this PR. Can you look into it or maybe just restart this particular pipeline?

ayrton04:
We don't have the ability to restart jobs on the ROS build farm. It would likely just get kicked off again in a bit, but we can force its hand by just closing and re-opening your PR.

ayrton04:
OK, merging this. I don't suppose you have a ROS 2 workspace? It would be good to get this change in there. I can also manually pull the change over.

ayrton04:
Thank you for the PR!

frygge:
Unfortunately, I don't yet.

Thanks a lot to you as well. We're happy to use the software in our cars :)

HaoguangYang:
> OK, merging this. I don't suppose you have a ROS 2 workspace? It would be good to get this change in there. I can also manually pull the change over.

I would be happy to port this over to ros2 branches as well.

roncapat:
Any news on a ros2 port?

HaoguangYang:
> Any news on a ros2 port?

Actually, I do have a compiling version but I got side-tracked by other projects recently. I will test my modifications when I have time and keep you posted.

PR #737, opened 2022-03-08
Read predict_to_current_time from ROS parameters

zygfrydw:
I have added a parameter `predict_to_current_time` to read the value from ROS parameters, otherwise the member `predict_to_current_time_` was not initialized.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -78,6 +78,7 @@ RosFilter<T>::RosFilter(const rclcpp::NodeOptions & options)
   latest_control_(),
   last_diag_time_(0, 0, RCL_ROS_TIME),
   last_published_stamp_(0, 0, RCL_ROS_TIME),
+  predict_to_current_time_(false),
   last_set_pose_time_(0, 0, RCL_ROS_TIME),
   latest_control_time_(0, 0, RCL_ROS_TIME),
   tf_timeout_(0ns),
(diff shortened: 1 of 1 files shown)


SteveMacenski (changes requested):
Where is this used? It doesn’t seem to be used anywhere in the codebase.

zygfrydw:
In the file **ros_filter.cpp** line 584 and 697 (function `RosFilter<T>::integrateMeasurements`) on branch **galactic**

```cpp
bool predict_to_current_time = predict_to_current_time_;
[...]
if (filter_.getInitializedStatus() && predict_to_current_time) {
rclcpp::Duration last_update_delta =
    current_time - filter_.getLastMeasurementTime();

filter_.validateDelta(last_update_delta);
filter_.predict(current_time, last_update_delta);

// Update the last measurement time and last update time
filter_.setLastMeasurementTime(
    filter_.getLastMeasurementTime() +
    last_update_delta);
}
```

On **noetic** branch this variable is called `predictToCurrentTime_`

ayrton04:
Yeah, looks like we just failed to load it. It gets loaded in noetic here:

https://github.com/cra-ros-pkg/robot_localization/blob/87b8d6bf6c94933da9f2f830460fa1c38c651db5/src/ros_filter.cpp#L871

ayrton04 (approved):

PR #738, opened 2022-03-11
Allow robot localization to export include directory

firemark:


firemark:
Probably I should describe why I need this export :)

For my cases I have written script to compare many localization estimators outside the ROS.
In [CMakeLists.txt](https://github.com/cra-ros-pkg/robot_localization/pull/738/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a) I see, library is already exported but without headers files. I think, exporting headers is nothing vulnerable.

CMakeLists.txt
```diff
@@ -331,7 +334,10 @@ install(DIRECTORY
   USE_SOURCE_PERMISSIONS
 )
 
+install(DIRECTORY include/ DESTINATION include)
+
 ament_export_include_directories(include)
 ament_export_libraries(${library_name})
```
SteveMacenski, line 337: Why not add to export libraries? It should already be being done from this line

SteveMacenski (changes requested):

CMakeLists.txt
SteveMacenski: I would probably reject this PR @ayrton04 specific user modifications for non-standard applications is exactly the reason to fork and I'm not sure introducing this has any benefit to us as maintainers / users moving forward

ayrton04: So I agree with that sentiment for the extra `export` and enabling PIC, but it seems like we install the `include` directory in ROS 1:

https://github.com/cra-ros-pkg/robot_localization/blob/87b8d6bf6c94933da9f2f830460fa1c38c651db5/CMakeLists.txt#L208-L212

Thoughts?

SteveMacenski: The installing headers is fine, its the exports

ayrton04: @firemark I am inclined to agree with @SteveMacenski on this. We can install the header files, but I'd prefer not to have the export. Not sure how I feel about enabling PIC.

ayrton04: Any update here? If we can remove the export and PIC for now, I'd appreciate it.

firemark: I'm sorry for a late response - I don't have time to finish the PR. I leave only line with installing includes.

ayrton04 (approved):

PR #739, opened 2022-03-11
Using angles library to normalize angles

firemark:
During experimenting with library I stuck on infinity loop with `NaN` values because:

```python
>>> from math import pi
>>> float("nan") + pi
nan
>>> float("nan") - pi
nan
```

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -28,6 +28,7 @@
   <depend>geographiclib</depend>
   <depend>message_filters</depend>
   <depend>nav_msgs</depend>
+  <depend>angles</depend>
   <build_depend>rclcpp</build_depend>
   <build_depend>rmw_implementation</build_depend>
   <depend>sensor_msgs</depend>
--- a/src/ekf.cpp
+++ b/src/ekf.cpp
@@ -33,10 +33,10 @@
 
 #include <robot_localization/ekf.hpp>
 #include <robot_localization/filter_common.hpp>
+#include <angles/angles.h>
 #include <Eigen/Dense>
 #include <rclcpp/duration.hpp>
 #include <vector>
-#include <cmath>
 
 namespace robot_localization
 {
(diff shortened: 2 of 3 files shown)


src/ekf.cpp
```diff
@@ -173,9 +174,10 @@ void Ekf::correct(const Measurement & measurement)
 
   // Wrap angles in the innovation
   for (size_t i = 0; i < update_size; ++i) {
-    if (update_indices[i] == StateMemberRoll ||
+    if ((update_indices[i] == StateMemberRoll ||
       update_indices[i] == StateMemberPitch ||
-      update_indices[i] == StateMemberYaw)
+      update_indices[i] == StateMemberYaw) &&
+      std::isfinite(innovation_subset(i)))
```
ayrton04: So if the innovation has values that are not finite, we should look into why. Otherwise, this feels like we're trying to just cover up a bigger issue.

Also, oof. I need to update this code to use [this](https://github.com/ros/angles/blob/2392b18b4101cf2e81b77e89e3535959ad03d1ca/angles/include/angles/angles.h#L83).

firemark: Probably is not your fault - I modified little a bit this code and BANG I got infinity loop. I think, this function should be protected against wrong values.

I will try to add "angles" library

ayrton04 (approved):

package.xml
```diff
@@ -28,6 +28,7 @@
   <depend>geographiclib</depend>
   <depend>message_filters</depend>
   <depend>nav_msgs</depend>
+  <depend>angles</depend>
```
tonynajjar, line 31: Don't you need to also added in the CMakelist?

tonynajjar: https://github.com/cra-ros-pkg/robot_localization/pull/747

firemark: That's weird because project has compiled properly on bare `ros:galactic` image with rosdep. Probably another dependency has this library.

PR #742, opened 2022-04-14
Fixing code style divergence for ament_uncrustify

AnishGDev:
Was failing colcon test before because of code style divergence when using ament_uncrustify. This PR should fix this.


SteveMacenski (approved):
Can you submit to ROS2 branch as well? That way it gets updated in future distros too!

AnishGDev:
yep will do

PR #746, opened 2022-04-25
move updating measurements into new method

firemark:
For my cases I have written script to compare many localization estimators outside the ROS.

For my usage I need only method to update filter with specified time, but this logic is inside `periodicUpdate` method, so I extracted into new method.

### DIFF
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -353,6 +353,11 @@ protected:
   //!
   void clearMeasurementQueue();
 
+  //! @brief Update filter with data from measurements queue
+  //! @param[in] time - The time at which to carry out integration
+  //!
+  void updateFilterWithMeasurements(const rclcpp::Time & time);
+
   //! @brief Adds a diagnostic message to the accumulating map and updates the
   //! error level
   //! @param[in] error_level - The error level of the diagnostic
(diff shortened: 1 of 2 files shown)


ayrton04 (approved):
While I am hesitant to merge something that seems like it is intended to solve one user's needs, this shouldn't affect the behavior or performance of the package. I'll ask @SteveMacenski to weigh in as well.

SteveMacenski:
Galactic goes EOL in 6 months, so really I don't care very much, but if we wanted to make this change across the board, we should do so across the board so that we don't revisit this every new distribution. Looking at the function `periodicUpdate`, I think it should be broken out into some smaller functional units so this would be in the right direction, but I wouldn't merge as is with just doing 1/10th of the work needed to accomplish it.

So I suppose in its current form, my thought would be "no", but if a well done separation of functional code in `periodicUpdate` was done to break out individual operations to make the code more self documenting, I could support merging that total change.

Although, this is a long term stable package that isn't really getting many updates, so I am susceptible to the "don't fix what ain't broke" argument to keep maintainer overhead as low as possible. 

So my answer is a tentative 'no' unless the user wants to patch up the entire function, and even then subject to @ayrton04's appetite for a large change.

firemark:
I decided to close this pull request and created new one: https://github.com/cra-ros-pkg/robot_localization/pull/761

PR #747, opened 2022-04-26
Fix angles dependency

tonynajjar:
Continuation of https://github.com/cra-ros-pkg/robot_localization/pull/739


ayrton04:
Thanks!

firemark:
I think, this list of `find_package` invokes is not necessary - you probably should use `ament_auto_find_build_dependencies()` - this macro finds dependencies from `package.xml` and try to add in `CmakeLists.txt`  

More info here: https://github.com/ament/ament_cmake/blob/master/ament_cmake_auto/cmake/ament_auto_find_build_dependencies.cmake

PR #748, opened 2022-04-30
Uploaded launch files of robot_localization for turtlebot3

AhmadHazem:


SteveMacenski:
I’m glad you’ve found some params that work well for you! But I don’t think this is the right place to store particular robot’s configurations.

ayrton04:
Thank you for the PR, but I have to agree with Steve here. As this is a config file for a specific platform, it seems that it should live in another repo, perhaps even the repo for the [Turtlebot 3](https://github.com/ROBOTIS-GIT/turtlebot3).

PR #752, opened 2022-05-09
Linting and header cleanup, part 1

ayrton04:
Draft PR for now. Just want the build farm to test this. I have more commits a-comin'.


SteveMacenski (approved):

ayrton04:
The random UKF test failures are odd. It passes every time locally. When it fails on the build farm now, it's always as if it's not getting _any_ of the sensor measurements:

```
[test_ukf_localization_node_interfaces-2] /tmp/ws/src/robot_localization/test/test_ukf_localization_node_interfaces.cpp:309: Failure
[test_ukf_localization_node_interfaces-2] Expected: (::fabs(filtered_.pose.pose.position.x - pose.pose.pose.position.x)) < (0.1), actual: 20 vs 0.1
[test_ukf_localization_node_interfaces-2] /tmp/ws/src/robot_localization/test/test_ukf_localization_node_interfaces.cpp:315: Failure
[test_ukf_localization_node_interfaces-2] Expected: (::fabs(filtered_.pose.pose.position.z - pose.pose.pose.position.z)) < (0.1), actual: 40 vs 0.1
[ukf_node-1] [INFO] [1652184610.335751570] [test_ukf_localization_node_interfaces_ukf]: Received a request to reset filter.
```

https://github.com/cra-ros-pkg/robot_localization/blob/215636a8119b4f5fdbdc3cddabe8f4f4c6ad3306/test/test_ukf_localization_node_interfaces.cpp#L309-L317

The only way we have that error at the end is if:

- The node wasn't receiving the sensor data
- The node wasn't publishing the filtered output
- The test sub wasn't receiving the filtered output

Does anyone know if ROS 2 has ever shown instability in pub and subs being correctly established? It seems unlikely, but maybe it's come up before.

ayrton04:
I think I have a handle on why that's happening. Will address in a forthcoming PR.

PR #753, opened 2022-05-10
Fix odometry and acceleration processing pipeline

HaoguangYang:
Fix odometry pose transform when odometry origin is different from base_link.
Fix covariance transformation when Odometry msg child_frame is not base_link.
Fix acceleration transformation, considering angular acceleration and centripetal acceleration.
Fix gravity removal for IMU with relative orientation (initial pose as [0,0,0,1]).

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -445,7 +445,9 @@ namespace RobotLocalization
       message.accel.accel.linear.x = state(StateMemberAx);
       message.accel.accel.linear.y = state(StateMemberAy);
       message.accel.accel.linear.z = state(StateMemberAz);
-
+      message.accel.accel.angular.x = angular_acceleration_.x();
+      message.accel.accel.angular.y = angular_acceleration_.y();
+      message.accel.accel.angular.z = angular_acceleration_.z();
       // Fill the covariance (only the left-upper matrix since we are not estimating
       // the rotational accelerations arround the axes
       for (size_t i = 0; i < ACCELERATION_SIZE; i++)
@@ -455,6 +457,15 @@ namespace RobotLocalization
           // We use the POSE_SIZE since the accel cov matrix of ROS is 6x6
           message.accel.covariance[POSE_SIZE * i + j] =
               estimateErrorCovariance(i + POSITION_A_OFFSET, j + POSITION_A_OFFSET);
+        }
+      }
+      for (size_t i = ACCELERATION_SIZE; i < POSE_SIZE; i++)
+      {
+        for (size_t j = ACCELERATION_SIZE; j < POSE_SIZE; j++)
+        {
+          // fill out the angular portion. We assume the linear and angular portions are independent.
+          message.accel.covariance[POSE_SIZE * i + j] =
+              angular_acceleration_cov_(i - ACCELERATION_SIZE, j - ACCELERATION_SIZE);
         }
       }
 
@@ -698,6 +709,26 @@ namespace RobotLocalization
     }
 
     RF_DEBUG("\n----- /RosFilter::integrateMeasurements ------\n");
+  }
+
+  template<typename T>
+  void RosFilter<T>::differentiateMeasurements(const ros::Time &currentTime){
+    if (filter_.getInitializedStatus()){
+      const double dt = (currentTime - lastAngAccTime_).toSec();
+      const Eigen::VectorXd &state = filter_.getState();
+      tf2::Vector3 newStateTwistRot(state(StateMemberVroll),
+                                    state(StateMemberVpitch),
+                                    state(StateMemberVyaw));
+      angular_acceleration_ = (newStateTwistRot - lastStateTwistRot_)/dt;
+      const Eigen::MatrixXd &cov = filter_.getEstimateErrorCovariance();
+      for (size_t i=0; i<3; i++){
+        for (size_t j=0; j<3; j++){
+          angular_acceleration_cov_(i,j) = cov(i+ORIENTATION_V_OFFSET, j+ORIENTATION_V_OFFSET)*2./(dt*dt);
+        }
+      }
+      lastStateTwistRot_ = newStateTwistRot;
+      lastAngAccTime_ = currentTime;
+    }
   }
 
   template<typename T>
@@ -1517,6 +1548,8 @@ namespace RobotLocalization
       }
     }
     while (moreParams);
+    angular_acceleration_cov_.resize(ORIENTATION_SIZE, ORIENTATION_SIZE);
+    angular_acceleration_cov_.setZero();
 
     // Now that we've checked if IMU linear acceleration is being used, we can determine our final control parameters
     if (useControl_ && std::accumulate(controlUpdateVector.begin(), controlUpdateVector.end(), 0) == 0)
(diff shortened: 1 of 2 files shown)


ayrton04:
Thank you for the PR.

Can you clarify some of these? Everything is marked as a fix, but these appear to be feature enhancements. Are you fixing any specific bugs that you can detail? Maybe just generally provide use cases that are solved by these changes.

> Fix odometry pose transform when odometry origin is different from base_link

I need to go over the PR, but what will happen if the odometry data is not provided in a frame relative to base_link, e.g., if it comes from a motion capture system that has a fixed transform to the robot's world frame (direct transform from mocap_frame to odom)?

HaoguangYang:
Items 1 and 2 were made in parallel with #728 , but with the additional enhancements of transforming the covariance correspondingly. To enable this feature, set parameter odomX_pose_use_child_frame to true.
Background of the fix:
A mobile robot has both a wheel odometer and an Intel RealSense T265 camera. The wheel odometer registers odom message at the geometry center of the robot (header.frame_id is odom and child_frame is base_link), while T265 reports odometry of itself -- header.frame_id  is T265_odom and child_frame is T265_link. I want to fuse these two odom sources with EKF.
With the T265 mounted on the edge of the robot, which is 30cm from base_link, a simple static transform from odom to T265_odom is not enough, as the size of the robot will couple rotation into the position components. Therefore, the differences in the two odom sources needs a proper transformation of the odom headers and child frames.

Q: "what will happen if the odometry data is not provided in a frame relative to base_link, e.g., if it comes from a motion capture system that has a fixed transform to the robot's world frame (direct transform from mocap_frame to odom)?"
A: This PR, as well as #728, fixes exactly that issue. E.g. in a motion capturing system, the origin of the robot pose is not necessarily the same as the origin of the wheel odometer on the robot, and the velocity reference frames (child_link) are not necessarily the same. Assume the mocap system provides odom message describing tf: mocap_odom->mocap_frame, the pose and velocity measured by the mocap on the child link (mocap_frame) should undergo the following transformation chain:
mocap_frame->mocap_odom->odom->base_link
to obtain the pose of base_link in odom.
According to the definition of Odometry message, mocap_odom->odom is addressed by a static transform, and odom->base_link is provided by the robot odometer.

Item 3 resolves the TODO in the prepareAcceleration function. The acceleration, after transformed to base_link, needs to remove the centripetal acceleration Omega x (Omega x r), where r is the translation vector. It also needs to remove the linear acceleration due to angular acceleration, beta x r, where beta is the time differential of Omega.

Item 4 adds a flag to support removal of gravity on IMU sensors that are not mounted upright and yet reports a pose relative to its intialization pose. Examples include the IMU sensor on the Intel RealSense cameras, which reports the readings in the optical frame and with tf to base_link specified. The feature is enabled by setting the imuX_relative parameter to true.

✅TODO: I will populate comments on the changed sections and fix the linting issue.
✅TODO: A bug in the PR is found when removing gravity. I will push a follow-up commit.

________________________________

ayrton04:
I haven't forgotten about this. Just need to make time for review. Thanks again for the PR and details.

ayrton04:
> Q: "what will happen if the odometry data is not provided in a frame relative to base_link, e.g., if it comes from a motion capture system that has a fixed transform to the robot's world frame (direct transform from mocap_frame to odom)?"
> A: This PR, as well as [#728](https://github.com/cra-ros-pkg/robot_localization/pull/728), fixes exactly that issue.

Hang on, I want to back up on this point. I'm sure you are aware of all of this, but for completeness, I want to go over it here.

`r_l` has always made the assumption that pose data will arrive in a frame that can be transformed to its world frame (e.g., `odom`) directly, without passing through the `world_frame` -> `base_link_frame` transform. So, for example, if I have an external motion capture  system that generates my pose in the `mocap` frame, and I know the transform from `mocap` to `odom`, then I can publish the `mocap`->`odom` transform, and the EKF can use messages from the motion capture system directly. It will transform the messages from the `mocap` frame to the `odom` frame and fuse it correctly.

This has been supported from day 1; the reason I use this specific example is because I have done this exact thing. It also works if you want to fuse, e.g., Gazebo's "perfect" state messages into the robot's pose. But again, for externally-referenced world frames, this already worked.

Likewise, the package makes the assumption that any velocity data will be reported in a frame that can be transformed to `base_link` directly (i.e., without having to "pass through" the `odom`->`base_link` transform in the tree).

If a `nav_msgs/Odometry` message gets fed to the filter, the pose data gets transformed from the messages's `frame_id` into the filter's `world_frame` (e.g., `odom`), and the twist data gets transformed from the message's `child_frame_id` into the filter's `base_link_frame` (e.g., `base_link`). They are then fused directly.

One of the main shortcomings of `r_l` has always been the case where someone mounts a sensor with its own world frame _to the robot itself_. This gets complicated, because the sensor's offset from the robot's origin will be defined relative to `base_link`. So now we have pose data that is coming from a frame relative to `base_link`, and that pose data also has to get transformed using the transform that the EKF itself is producing (i.e., `odom`->`base_link`).

#728 solved this, I believe, by making it so that `differential` mode, when applied, would use the `child_frame_id` of the odometry message. This would allow us to define a `base_link`->`<child_frame_id>` transform. Then the `differential` mode would differentiate two consecutive pose measurements from that sensor (which would yield velocity in the sensor frame), and then that sensor-frame velocity would get transformed by the `base_link`->`<child_frame_id>` transform, and we'd get robot body-frame velocity.

src/ros_filter.cpp
```diff
@@ -701,6 +712,32 @@ namespace RobotLocalization
...
+    {
+      const double dt = (currentTime - lastDiffTime_).toSec();
+      const Eigen::VectorXd &state = filter_.getState();
+      // Specific to angular acceleration for now...
+      tf2::Vector3 newStateTwistRot(state(StateMemberVroll),
+                                    state(StateMemberVpitch),
+                                    state(StateMemberVyaw));
+      angular_acceleration_ = (newStateTwistRot - lastStateTwistRot_)/dt;
+      const Eigen::MatrixXd &cov = filter_.getEstimateErrorCovariance();
+      for (size_t i = 0; i < 3; i ++)
```
ayrton04, line 728: Just for consistency, it might make sense to change 3 to `ORIENTATION_SIZE`.

ayrton04 (approved):
OK, took a deeper read through this, and it looks sound. Thanks for the PR!

ayrton04:
I see #728 was actually broader in scope than I realised, and isn't just targeting differential mode.

src/ros_filter.cpp
HaoguangYang:
```suggestion
      for (size_t i = 0; i < ORIENTATION_SIZE; i ++)
```

src/ros_filter.cpp
```diff
@@ -701,6 +712,32 @@ namespace RobotLocalization
...
+      {
+        for (size_t j = 0; j < 3; j ++)
```
HaoguangYang, line 730:
```suggestion
        for (size_t j = 0; j < ORIENTATION_SIZE; j ++)
```

rokusottervanger:
I came here because I suddenly see drift in my odometry state because my IMU is at a small but non-negligable angle (maybe the robot is at that angle, or the imu is not mounted perfectly straight.

Looking at this PR, I find it odd that:
- There's suddenly math being performed on the state in the `ros_filter` layer (calculating the angular acceleration by means of differentiation)?
- There's suddenly state being held outside of the `filter_`'s state (specifically the angular acceleration)?
- The filter assumes a constant linear acceleration and constant angular velocity model, but now the angular acceleration is also used?

PR #754, opened 2022-05-11
Fixing and cleaning up interface tests

ayrton04:
This PR does two things:

1. Makes the interface test (hopefully) more deterministic, with the aim of eliminating the spurious failures that happen now and again on the buildfarm.
2. The interface tests for the EKF and UKF were identical; we were just launching different nodes. So I made a single test and just built it twice, and then pass in the filter node type to be used to the test from CMake.


ayrton04:
Huh, the test in question is passing, but it looks like it's unhappy, probably because the node is getting destroyed before all the pub/subs. I'll try to fix.

```
[test_ekf_node_interfaces-2] cannot publish data, at ./src/rmw_publish.cpp:62 during '__function__'
[test_ekf_node_interfaces-2] destroy_client() failed to delete datareader, at ./src/rmw_client.cpp:91 during '__function__'
[ERROR] [test_ekf_node_interfaces-2]: process has died [pid 1780, exit code -11, cmd '/tmp/ws/install_isolated/robot_localization/lib/robot_localization/test_ekf_node_interfaces'].
[INFO] [ekf_node-1]: sending signal 'SIGINT' to process[ekf_node-1]
[ekf_node-1] [INFO] [1652267873.238070354] [rclcpp]: signal_handler(signum=2)
[INFO] [ekf_node-1]: process has finished cleanly [pid 1778]
```

ayrton04:
Hurray, all tests pass.

SteveMacenski:
I saw -1000 lines and freaked out about the size of PR I'd have to review :laughing:

ayrton04:
I love seeing red in PRs.

PR #755, opened 2022-05-12
Adding support for setting diagonal values for covariance matrices

ayrton04:
I've been meaning to do this for years. Should make config files less ugly.


SteveMacenski:
i don't disagree that this would be easier to read, but this seems late in the game for such a big change to folk's configuration files migrating between distributions. If this was something that we were super interested in re-architecting / fixing up for the next 10+ years of use, I agree refreshes would be nice. But I thought R_L was in a 'don't fix what ain't broke' phase and looking forward to something like Fuse?

ayrton04:
Both are supported. I haven't removed support for loading the full matrix.

And yes, we'll get to porting fuse next. I just want to clean a few things up here before I tie it off.

SteveMacenski:
OK! I just saw the defaults change in the example files, but LGTM. If its both, then no problemo

djaniel:
Hi, 
I have been doing some tests with robot localization.  I installed robot localization using apt and I was getting an error related to this Issue. I checked and noticed that the commit that solves this issue is on branch ros2, but not in branch humble-devel (the one provided by apt).

ayrton04:
This isn't an issue; it was a new feature. What error did you get?

PR #756, opened 2022-05-12
Removal of dead code and better debug output

A-I-Ivanov:
This PR does a few things:
1) Removes a variety of dead code within ros_filter.cpp
2) Ensures that measurements are checked uniformly for time instead of ad-hoc
3) Takes care of a corner case that generates a memory leak from measurement_queue_ when clock times are wrong. 
4) Produces warnings if incoming data is badly timestamped into the future. 

This was primarily borne out from difficulties I had with the package where the filter was failing silently

### DIFF
--- a/doc/state_estimation_nodes.rst
+++ b/doc/state_estimation_nodes.rst
@@ -242,6 +242,10 @@ If any of your sensors produce data with timestamps that are older than the most
 ^^^^^^^^^^^^^^^
 If ``smooth_lagged_data`` is set to *true*, this parameter specifies the number of seconds for which the filter will retain its state and measurement history. This value should be at least as large as the time delta between your lagged measurements and the current time.
 
+~max_future_queue_size
+^^^^^^^^^^^^^^^
+This parameter specifies the number of total data samples, arriving between filter runs, the filter will retain. This is to ensure excessive memory is not used as would happen if incoming data is badly time stamped or the filter ``frequency`` is badly set. 
+
 ~[sensor]_nodelay
 ^^^^^^^^^^^^^^^^^
 
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -473,6 +473,25 @@ protected:
     std::vector<bool> & updateVector, Eigen::VectorXd & measurement,
     Eigen::MatrixXd & measurementCovariance);
 
+  //! @brief Validates incoming measurement ordering and produces errors if
+  //! validation does not pass.
+  //! @param[in] topicName - The name of the topic over which this message was
+  //! received
+  //! @param[in] stamp - The timestamp associated with this message.
+  //! @return true if the measrement is newer than all known measurements.
+  //!
+  bool validateMeasurementOrdering(const std::string& topicName,
+                                   const builtin_interfaces::msg::Time& stamp);
+
+  //! @brief Validates incoming measurements are within a reasonable timeframe.
+  //! @param[in] topicName - The name of the topic over which this message was
+  //! received
+  //! @param[in] stamp - The timestamp associated with this message.
+  //! @return true if the measrement is newer than all known measurements.
+  //!
+  bool validateMeasurementTime(const std::string& topicName,
+                               const builtin_interfaces::msg::Time& stamp);
+
   //! @brief Whether or not we print diagnostic messages to the /diagnostics
   //! topic
   //!
@@ -545,13 +564,16 @@ protected:
   //!
   double gravitational_acceleration_;
 
-  //! @brief The depth of the history we track for smoothing/delayed measurement
-  //! processing
+  //! @brief The depth of the history we track for smoothing/delayed
+  //! measurement processing
   //!
   //! This is the guaranteed minimum buffer size for which previous states and
   //! measurements are kept.
   //!
   rclcpp::Duration history_length_;
+
+  //! @brief This is the maximum measurement queque length.
+  uint64_t max_future_queue_size_;
 
   //! @brief tf frame name for the robot's body frame
   //!
@@ -622,8 +644,8 @@ protected:
   //! @brief This object accumulates static diagnostics, e.g., diagnostics
   //! relating to the configuration parameters.
   //!
-  //! The values are treated as static and always reported (i.e., this object is
-  //! never cleared)
+  //! The values are treated as static and always reported (i.e., this object
+  //! is never cleared)
   //!
   std::map<std::string, std::string> static_diagnostics_;
 
@@ -646,8 +668,8 @@ protected:
   //! stores the initial measurements. Note that this is different from using
   //! differential mode, as in differential mode, pose data is converted to
   //! twist data, resulting in boundless error growth for the variables being
-  //! fused. With relative measurements, the vehicle will start with a 0 heading
-  //! and position, but the measurements are still fused absolutely.
+  //! fused. With relative measurements, the vehicle will start with a 0
+  //! heading and position, but the measurements are still fused absolutely.
   std::map<std::string, tf2::Transform> initial_measurements_;
 
   //! @brief If including acceleration for each IMU input, whether or not we
@@ -688,9 +710,9 @@ protected:
   //!
   std::map<std::string, Eigen::MatrixXd> previous_measurement_covariances_;
 
-  //! @brief By default, the filter predicts and corrects up to the time of the
-  //! latest measurement. If this is set to true, the filter does the same, but
-  //! then also predicts up to the current time step.
+  //! @brief By default, the filter predicts and corrects up to the time of
+  //! the latest measurement. If this is set to true, the filter does the
+  //! same, but then also predicts up to the current time step.
   //!
   bool predict_to_current_time_;
 
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -83,6 +83,7 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
   yaw_offset_(0.0),
   zero_altitude_(false)
 {
+
   tf_buffer_ = std::make_unique<tf2_ros::Buffer>(this->get_clock());
   tf_listener_ = std::make_unique<tf2_ros::TransformListener>(*tf_buffer_);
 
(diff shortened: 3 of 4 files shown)


SteveMacenski:
I don't necessarily think its terribly useful to open a PR called WIP without any context / information?

A-I-Ivanov:
> I don't necessarily think its terribly useful to open a PR called WIP without any context / information?

My apologies, I'm new to the OS dev style. I wasn't intending on anyone to look at it yet. If you have recommendations on best practices let me know. It should be polished up shortly.

SteveMacenski:
Usually you wouldn't open the pull request until you were ready for folks to look at it :smile: (or are the core developer of the work, so the main repo is your sandbox)

A-I-Ivanov:
>

A-I-Ivanov:
Sorry for the false start. Should be good now.

ayrton04:
Thanks for this. Will review soon.

doc/state_estimation_nodes.rst
```diff
@@ -242,6 +242,10 @@ If any of your sensors produce data with timestamps that are older than the most
 ^^^^^^^^^^^^^^^
 If ``smooth_lagged_data`` is set to *true*, this parameter specifies the number of seconds for which the filter will retain its state and measurement history. This value should be at least as large as the time delta between your lagged measurements and the current time.
 
+~max_future_queue_size
+^^^^^^^^^^^^^^^
+This parameter specifies the number of total data samples, arriving between filter runs, the filter will retain. This is to ensure excessive memory is not used as would happen if incoming data is badly time stamped or the filter ``frequency`` is badly set.
```
ayrton04, line 247: Maybe "arriving between execution cycles" instead of "arriving between filter runs"? Also, could we not control this by limiting the queue sizes on the actual sensor inputs, e.g., use the existing `<input name>_queue_size` parameters? With this option, if you have one extremely high frequency sensor and one very low frequency sensor, it seems possible that the wrong setting here will result in one sensor getting completely ignored.

include/robot_localization/ros_filter.hpp
```diff
@@ -473,6 +473,25 @@ class RosFilter : public rclcpp::Node
...
     Eigen::MatrixXd & measurementCovariance);
 
+  //! @brief Validates incoming measurement ordering and produces errors if
+  //! validation does not pass.
+  //! @param[in] topicName - The name of the topic over which this message was
+  //! received
+  //! @param[in] stamp - The timestamp associated with this message.
+  //! @return true if the measrement is newer than all known measurements.
+  //!
+  bool validateMeasurementOrdering(const std::string& topicName,
```
ayrton04, line 483: Not sure how some of these other methods retained `camelCase` variable names, but we should try to stick to ROS 2 standards and use `snake_case`, e.g., `topic_name`.

include/robot_localization/ros_filter.hpp
```diff
@@ -473,6 +473,25 @@ class RosFilter : public rclcpp::Node
...
+                                   const builtin_interfaces::msg::Time& stamp);
+
+  //! @brief Validates incoming measurements are within a reasonable timeframe.
+  //! @param[in] topicName - The name of the topic over which this message was
+  //! received
+  //! @param[in] stamp - The timestamp associated with this message.
+  //! @return true if the measrement is newer than all known measurements.
+  //!
+  bool validateMeasurementTime(const std::string& topicName,
```
ayrton04, line 492: As above.

src/ros_filter.cpp
```diff
@@ -177,30 +235,25 @@ void RosFilter<T>::toggleFilterProcessingCallback(
...
 void RosFilter<T>::accelerationCallback(
-  const sensor_msgs::msg::Imu::SharedPtr msg,
-  const CallbackData & callback_data,
-  const std::string & target_frame)
-{
-  // If we've just reset the filter, then we want to ignore any messages
-  // that arrive with an older timestamp
-  if (last_set_pose_time_ >= msg->header.stamp) {
+    const sensor_msgs::msg::Imu::SharedPtr msg,
+    const CallbackData& callback_data, const std::string& target_frame) {
```
ayrton04, line 239: Any reason to change the style here? If it passed linting before, I'd prefer that we leave it alone and minimize changes. I'm pretty sure ROS 2 standards specify that reference types be given as `TypeName & variable_name`, which is what was there before. And `target_frame` was on its own line.

src/ros_filter.cpp
```diff
@@ -302,13 +329,10 @@ void RosFilter<T>::controlStampedCallback(
...
     filter_.setControl(latest_control_, msg->header.stamp);
   } else {
-    // ROS_WARN_STREAM_THROTTLE(5.0, "Commanded velocities must be given in the
-    // robot's body frame (" << base_link_frame_id_ << "). Message frame was " <<
-    // msg->header.frame_id);
-    std::cerr <<
-      "Commanded velocities must be given in the robot's body frame (" <<
-      base_link_frame_id_ << "). Message frame was " <<
-      msg->header.frame_id << "\n";
+    std::cerr
```
ayrton04, line 332: Can we use the ROS 2 logging macros here? I think I updated these in the `rolling` branch.

src/ros_filter.cpp
```diff
@@ -319,8 +343,17 @@ void RosFilter<T>::enqueueMeasurement(
...
-  MeasurementPtr meas = MeasurementPtr(new Measurement());
 
+  rcpputils::check_true(
+      measurement_queue_.size() < max_future_queue_size_,
+      "Too many measurements on queue . This may mean the filter is "
+      "running slowly or that measurements are comming in with bad  "
+      "timestamps. max_future_queue_size_ is: " +
+          std::to_string(max_future_queue_size_));
+
+  constexpr double kMaxQueueTimeS = 0.5;
```
ayrton04, line 354: `snake_case`, please

src/ros_filter.cpp
```diff
@@ -151,7 +154,62 @@ void RosFilter<T>::reset()
...
+    return false;
+  }
+
+  return true;
+}
+
+template <typename T>
+bool RosFilter<T>::validateMeasurementOrdering(
+    const std::string& topic_name, const builtin_interfaces::msg::Time& stamp) {
+  //  Put the initial value in the lastMessagTimes_ for this variable if it's
```
ayrton04, line 183: *last_message_times_ (I realise this was copied from elsewhere)

src/ros_filter.cpp
```diff
@@ -177,30 +235,25 @@ void RosFilter<T>::toggleFilterProcessingCallback(
...
-  const std::string & topic_name = callback_data.topic_name_;
-
-  RF_DEBUG(
-    "------ RosFilter<T>::accelerationCallback (" << topic_name <<
-      ") ------\n")
-  // "Twist message:\n" << *msg);
+  RF_DEBUG("------ RosFilter<T>::accelerationCallback (" << topic_name
+                                                         << ") ------\n")
 
   if (last_message_times_.count(topic_name) == 0) {
```
ayrton04, line 250: As you did elsewhere, this block should be deleted, as this logic is now handled by `validateMeasurementOrdering`.

src/ros_filter.cpp
```diff
@@ -330,6 +363,13 @@ void RosFilter<T>::enqueueMeasurement(
   meas->latest_control_ = latest_control_;
   meas->latest_control_time_ = latest_control_time_;
   measurement_queue_.push(meas);
+
+  if ((time - measurement_queue_.top()->time_).nanoseconds() / 1e9 >
```
ayrton04, line 367: This feels totally arbitrary to me. We should add a timeout parameter. There could be valid use cases for the current time being 0.5 seconds ahead of the most recent measurement. Could we possible even use the `sensor_timeout` parameter here?

src/ros_filter.cpp
```diff
@@ -330,6 +363,13 @@ void RosFilter<T>::enqueueMeasurement(
...
+      kMaxQueueTimeS) {
+    RCLCPP_ERROR(this->get_logger(), "Warning: messages are queued at least "
```
ayrton04, line 369: We're using `RCLCPP_ERROR` and then printing out the word `Warning:`. If it's a warning, I'd remove the `Warning:` and use `RCLCPP_WARN`. If it's an error, then I'd just remove the `Warning:`.

src/navsat_transform.cpp
```diff
@@ -83,6 +83,7 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
   yaw_offset_(0.0),
   zero_altitude_(false)
 {
+
```
ayrton04, line 86: Please remove.

src/ros_filter.cpp
```diff
@@ -2115,10 +2084,7 @@ void RosFilter<T>::periodicUpdate()
 
           world_transform_broadcaster_->sendTransform(map_odom_trans_msg);
         } catch (...) {
-          // ROS_ERROR_STREAM_DELAYED_THROTTLE(5.0, "Could not obtain
-          // transform from "
-          //                                  << odom_frame_id_ << "->" <<
-          //                                  base_link_frame_id_);
+          RF_DEBUG("Filter could not get odom to baselink ids. \n")
```
ayrton04, line 2087:
```suggestion
          RF_DEBUG("Could not obtain transform from " << odom_frame_id_ <<
              " to " << base_link_frame_id_ <<  "\n")
```

src/ros_filter.cpp
```diff
@@ -2021,8 +1990,8 @@ void RosFilter<T>::periodicUpdate()
   auto filtered_position = std::make_unique<nav_msgs::msg::Odometry>();
 
   bool corrected_data = false;
-
   if (getFilteredOdometryMessage(filtered_position.get())) {
+    RF_DEBUG("Inside main loop function \n")
```
ayrton04, line 1994: This is a little ambiguous. Maybe we can think of something a little clearer?

src/ros_filter.cpp
```diff
@@ -3330,15 +3260,9 @@ bool RosFilter<T>::revertTo(const rclcpp::Time & time)
 
     if (last_history_state) {
       RF_DEBUG(
-        "Will revert to oldest state at " <<
-          filter_utilities::toSec(last_history_state->latest_control_time_) <<
-          ".\n");
-
-      // ROS_WARN_STREAM_DELAYED_THROTTLE(history_length_, "Could not revert "
```
ayrton04, line 3337: I think I put a bunch of these back and just used the `RCLCPP` equivalents in the `rolling` branch.

ayrton04 (changes requested):
Thanks for the PR! Just a few things to tidy up.

PR #758, opened 2022-05-22
Filters as ROS2 components

roncapat:
This patch make Ekf and Ukf loadable as ROS2 components.

This should allow reduction in transport latency from sensor nodes that are loadable as components too: composing the measurement sources with the filter can shorten the pub/sub overhead, by using intra-process communication.

### DIFF
--- a/src/ros_components.cpp
+++ b/src/ros_components.cpp
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2022, Patrick Roncagliolo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided
+ * with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ * contributors may be used to endorse or promote products derived
+ * from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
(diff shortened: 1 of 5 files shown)


src/ros_components.cpp
```diff
@@ -0,0 +1,66 @@
...
+   rclcpp::NodeOptions _option_helper(std::string node_name, const rclcpp::NodeOptions & options){
+      rclcpp::NodeOptions o = options;
+      std::vector<std::string> a;
+      a = o.arguments();
+      a.insert(a.begin(), node_name);
+      o.arguments(a);
+      return o;
+   }
+
+   EkfFilter::EkfFilter(const rclcpp::NodeOptions & options)
```
SteveMacenski, line 45: Why not just register the EKF / UKF files? https://github.com/cra-ros-pkg/robot_localization/blob/humble-devel/src/ukf.cpp If the alpha/beta/kappa params are stuck into the UKF file itself, there shouldn't be a need for these separate objects for componentizing

SteveMacenski: Especially if we don't need a shared pointer that you changed to `this` (assuming that works)

SteveMacenski (changes requested):

src/ros_components.cpp
roncapat: Hi, happy to share my findings... Of course I tried in a first iteration not to define the classes you saw, but just exposing existing ones. 

One problem there is that the RosFilter<> extracts the node name in the constructor (thus, I needed to inject it manually, since component instantiation is a standardized procedure). 

One other problem is that the usage of shared_from_this() lead to bad_weak_ptr exceptions due to the fact that component instantiation seems not to create the node as a shared pointer.

Third, I wanted to mimic the behaviour of the Ekf and Ukf nodes initialization.

Let me know if you need further demonstration of the first two points, I can make some branches demonstrating the various approaches I tried.

SteveMacenski: > extracts the node name in the constructor

@ayrton04 are you against just having the names default to `filter_node` or something so that this isn't a problem?

Yeah, I understand the shared pointer issue well, those changes would be needed either way

roncapat: Yeah, changing the way node name is set in the mainline code would be a simplification.

One last thing: to expose RosFilter<Ekf> as a component, the risk is to have a component named "robot_localization::RosFilter<robot_localization::Ekf>" since I tried in a lot of ways (failing) to use the typedef "RosEkf". I think that typedefs are not visible at linking/loading scope, as a symbol. I digged with "nm" tool in the shared library to confirm my theory.

SteveMacenski: Believe that would be true

roncapat: So, are there requested changes at the end?

SteveMacenski: Yes, all above :-)

roncapat: Well, from the discussion I though we ruled out some initial comments from being implemented

SteveMacenski: You're still needing a shared pointer in the constructor, even if you call it in another function. The scope of the constructor is not completed yet so there's no reference counter on the shared pointer. Thus, this still has the same issues as before. If you're finding that changing to `this` works, then that would work too for the componentization of the EKF/UKF themselves without wrappers. 

If we change the name to some logical default, that deals with that issue.

`robot_localization::RosFilter<robot_localization::Ekf>` is not super clean, I will admit, but I think if you register the type def instead of the object, that should work. Either way, that would still be fine semantically

stephen-derosa:
very excited for this!!

PR #761, opened 2022-06-28
Refactoring: Splitting periodicUpdate method into smaller methods

firemark:
Because method `RosFilter::periodicUpdate` is a big method, I cut into smaller methods.

In some methods (`validateFilterOutput`, `getFilteredAccelMessage` and `getFilteredOdometryMessage`) I changed an argument type from pointer to reference, because a pointer could be dangerous.

### DIFF
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -247,23 +247,19 @@ public:
   //!
   void updateFilterWithMeasurements(const rclcpp::Time & time);
 
-  //! @brief publish world to base link transform and position
-  //! with odometry message
-  //! @param[in] filtered_position - filtered position from
-  //! getFilteredOdometryMessage.
+  //! @brief publish world to base link transform and position with odometry message.
+  //! @param[in] filtered_position - filtered position from getFilteredOdometryMessage.
   //! @return true if data is corrected otherwise false.
   //!
-  bool publishPositionWithOdometry(nav_msgs::msg::Odometry filtered_position);
+  bool publishPositionWithOdometry(const nav_msgs::msg::Odometry & filtered_position);
 
   //! @brief Update world to base link transform using filtered position.
-  //! @param[in] filtered_position - filtered position from
-  //! getFilteredOdometryMessage.
+  //! @param[in] filtered_position - filtered position from getFilteredOdometryMessage.
   //!
   void updateWorldToBaseLinkTransform(const nav_msgs::msg::Odometry & filtered_position);
 
   //! @brief Publish world transform using frame id from filtered position.
-  //! @param[in] filtered_position - filtered position from
-  //! getFilteredOdometryMessage.
+  //! @param[in] filtered_position - filtered position from getFilteredOdometryMessage.
   //!
   void publishWorldTransform(const nav_msgs::msg::Odometry & filtered_position);
 
@@ -341,7 +337,7 @@ public:
   //! @param[out] message - The standard ROS odometry message to be validated
   //! @return true if the filter output is valid, false otherwise
   //!
-  bool validateFilterOutput(nav_msgs::msg::Odometry & message);
+  bool validateFilterOutput(const nav_msgs::msg::Odometry & message);
 
 protected:
   //! @brief Finds the latest filter state before the given timestamp and makes
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -199,14 +199,14 @@ public:
   //! @param[out] message - The standard ROS odometry message to be filled
   //! @return true if the filter is initialized, false otherwise
   //!
-  bool getFilteredOdometryMessage(nav_msgs::msg::Odometry * message);
+  bool getFilteredOdometryMessage(nav_msgs::msg::Odometry & message);
 
   //! @brief Retrieves the EKF's acceleration output for broadcasting
   //! @param[out] message - The standard ROS acceleration message to be filled
   //! @return true if the filter is initialized, false otherwise
   //!
   bool getFilteredAccelMessage(
-    geometry_msgs::msg::AccelWithCovarianceStamped * message);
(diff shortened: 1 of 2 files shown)


src/ros_filter.cpp
```diff
@@ -2006,149 +2006,26 @@ void RosFilter<T>::periodicUpdate()
...
-    // Fire off the position and the transform
-    if (!corrected_data) {
-      position_pub_->publish(std::move(filtered_position));
-    }
+  updateFilterWithMeasurements(cur_time);
 
-    if (print_diagnostics_) {
-      freq_diag_->tick();
+  // Get latest state and publish it
+  {
```
firemark: In here, I limited the scope because `filtered_position` is not required later.

I decided to resign from share pointer because it's an only local variable not saved anywhere.

`filtered_acceleration` was changed in similar way.

include/robot_localization/ros_filter.hpp
```diff
@@ -199,14 +199,14 @@ class RosFilter : public rclcpp::Node
   //! @param[out] message - The standard ROS odometry message to be filled
   //! @return true if the filter is initialized, false otherwise
   //!
-  bool getFilteredOdometryMessage(nav_msgs::msg::Odometry * message);
+  bool getFilteredOdometryMessage(nav_msgs::msg::Odometry & message);
```
ayrton04: I am struggling to find it, but at some point, there was definitely a document that suggested that in ROS 2, passing parameters as pointers was preferred. Was there a reason for this change? I'm not against it - I prefer reference parameters - but I am trying to determine if this was required for some reason, or you were just looking to be consistent.

include/robot_localization/ros_filter.hpp
```diff
@@ -245,6 +245,31 @@ class RosFilter : public rclcpp::Node
...
+  //!
+  void updateFilterWithMeasurements(const rclcpp::Time & time);
+
+  //! @brief publish world to base link transform and position
+  //! with odometry message
+  //! @param[in] filtered_position - filtered position from
+  //! getFilteredOdometryMessage.
+  //! @return true if data is corrected otherwise false.
+  //!
+  bool publishPositionWithOdometry(nav_msgs::msg::Odometry filtered_position);
```
ayrton04: If we're sticking with reference types, is there any reason this isn't const ref?

include/robot_localization/ros_filter.hpp
```diff
@@ -245,6 +245,31 @@ class RosFilter : public rclcpp::Node
   //!
   void periodicUpdate();
 
+  //! @brief Update filter with data from measurements queue.
+  //! @param[in] time - The time at which to carry out integration.
+  //!
+  void updateFilterWithMeasurements(const rclcpp::Time & time);
+
+  //! @brief publish world to base link transform and position
```
ayrton04: Would this go over 100 characters without the line break?

src/ros_filter.cpp
```diff
@@ -2006,149 +2006,26 @@ void RosFilter<T>::periodicUpdate()
...
+    nav_msgs::msg::Odometry filtered_position;
+    if (getFilteredOdometryMessage(filtered_position)) {
+      corrected_data = publishPositionWithOdometry(std::move(filtered_position));
```
ayrton04: If you make this method take a const ref, there's no need for the `std::move`. Not sure what that does in the context of a value parameter anyway.

src/ros_filter.cpp
```diff
@@ -2180,6 +2057,164 @@ void RosFilter<T>::periodicUpdate()
...
+
+  auto frame_id = filtered_position.header.frame_id;
+
+  if (frame_id == odom_frame_id_) {
+    world_transform_broadcaster_->sendTransform(world_base_link_trans_msg_);
+    return;
+  }
+
+  if (frame_id != map_frame_id_) {
+    std::cerr << "Odometry message frame_id was " <<
```
ayrton04: As long as you're here, you can replace any `cerr` statements with the actual ROS 2 logging macros, and uncomment/update the ROS 1 macros further down.

ayrton04 (changes requested):
Generally looks good, thanks. Just some smaller requests.

include/robot_localization/ros_filter.hpp
firemark: Raw pointers are more dangerous than references - pointer could be a NULL or contains wrong address, so it requires more guards in the method (e.g. https://pvs-studio.com/en/docs/warnings/v595/). So only reason is reference is safer because a compiler guards the argument.

> I am struggling to find it, but at some point, there was definitely a document that suggested that in ROS 

Unfortunately I didn't find that - but I found:

* https://design.ros2.org/articles/intraprocess_communications.html#number-of-message-copies -  based on what message should be used in intraprocess communication, subscriber with shared pointer generates less copies of message instead of  subscriber with unique pointer.
* https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/any_subscription_callback.hpp#L935 only callbacks with shared pointer generate less copies.

These links are not related with this comment but they could help with subscriptions in the future :)

Anyway I see, the best way is use [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional) instead of reference and returning bool. Could you review my next commit?

include/robot_localization/ros_filter.hpp
firemark: Half year ago I'm was experimenting with `std::move()` and it was bad.

I have changed the argument to const ref.

include/robot_localization/ros_filter.hpp
firemark: Yep. Fixed.

src/ros_filter.cpp
firemark: Fixed.

src/ros_filter.cpp
firemark: Fixed.

PR #776, opened 2022-09-19
add possibly forgotten base link frame id

incebellipipo:
Hi!

Quick question,
Could this be forgotten somehow?

Is there any intention why `child_frame_id` field left blank in the first place?

https://github.com/incebellipipo/robot_localization/blob/77a523eba6e2224f2df21abca35e22fd243afa04/src/navsat_transform.cpp#L470

It is the same case for ROS2.
https://github.com/cra-ros-pkg/robot_localization/blob/a9cb0230b8208890908d03f7a66fd9c81a21f39c/src/navsat_transform.cpp#L470-L473

Thanks!

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -467,6 +467,7 @@ namespace RobotLocalization
     // Set header information stamp because we would like to know the robot's position at that timestamp
     gps_odom.header.frame_id = world_frame_id_;
     gps_odom.header.stamp = gps_update_time_;
+    gps_odom.child_frame_id = base_link_frame_id_;
 
     // Now fill out the message. Set the orientation to the identity.
     tf2::toMsg(transformed_cartesian_gps, gps_odom.pose.pose);


ayrton04:
See the [definition](http://docs.ros.org/en/noetic/api/nav_msgs/html/msg/Odometry.html) of `nav_msgs/Odometry`:

```
# The pose in this message should be specified in the coordinate frame given by header.frame_id.
# The twist in this message should be specified in the coordinate frame given by the child_frame_id
```

In this case, we are only filling out pose data in the message, because that's all we have. In truth, the output should probably be a `geometry_msgs/PoseWithCovarianceStamped`, but it's a little late to change that.

What use case would require you to have the `child_frame_id` filled out?

incebellipipo:
**Short story**:
I see, thinking of it as `geometry_msgs/PoseWithCovarianceStamped` makes more sense. Especially after discovering the parameter `[sensor]_pose_use_child_frame`.
https://github.com/cra-ros-pkg/robot_localization/blob/0d3c7cc76461eaee77593c95b390a3a7897e30e2/src/ros_filter.cpp#L1108

That parameter is not in the documentation. I would further this PR as a documentation PR but I checked the `humble-devel` for `[sensor]_pose_use_child_frame`, it's not there anymore. So, I'm closing it 

**Long story**: 
My problem, I see that it has nothing to do with `child_frame_id` parameter, is that I've been trying to fuse `navsat_transform_node` output to second `ekf_localization_node`. Just to have the structure below.
image
I kept getting awkward results from something that should be trivial. TF lookups were failing, rotations were all over the place. I started to investigate odometry sources which where `/odometry/filtered/local`, `/odometry/filtered/global`, and `/odometry/gps`. Then I realized that the `child_frame_id` parameter was empty. As frustrated as I was, I blamed its absence. a bit confusing it was.

**Thanks!**

Issue #784, opened 2022-12-16
ROS 2 - Install include directory


oKermorgant:
Hi, 

We are building a package based on `robot_localization` to allow tight coupling with range / bearing measurements.
It basically boils down to using the provided EKF but adding new measurement types (range to anchors).

In order to compile our package we currently need to have a clone of `robot_localization` because the headers are not installed.

Would it be possible to install the `include` directory in the CMakeLists in order to avoid having to keep a local copy of the whole code?
The linking step with `rl_lib` is already fine, really what is missing to use `robot_localization` code in another project are the headers.

Thanks,

SteveMacenski:
Sounds reasonable. Can you open a PR to ament export the includes? That should do the trick. If you can do it in the next couple of hours, I can get that out in my releases today.

PR #786, opened 2022-12-19
Install headers

oKermorgant:
This PR answers https://github.com/cra-ros-pkg/robot_localization/issues/784
Install the headers to allow using `robot_localization` as a build dependency to other packages.


SteveMacenski (approved):

SteveMacenski:
@oKermorgant please base off of `ros2` so that we can have it reflected in all new distributions as well. We branch off from there

oKermorgant:
Cool, thanks for the merge and for the amazing work.

PR #800, opened 2023-03-28
Remove [sensor]_nodelay from ros2 params

borongyuan:


borongyuan:
In fact, DDS also supports nodelay settings when using TCP communication. But these should be set in the DDS configuration file, without the ROS package bothering.

doc/state_estimation_nodes.rst
```diff
@@ -245,7 +245,7 @@ If ``smooth_lagged_data`` is set to *true*, this parameter specifies the number
 ~[sensor]_nodelay
 ^^^^^^^^^^^^^^^^^
 
-Specific parameters:
+ROS1 specific parameters:
```
ayrton04: I believe this should be "ROS 1". 😛

ayrton04 (approved):
Thanks!

Issue #820, opened 2023-06-22
When using manual data and non 'odom' world reference frame name, race condition results in incorrect transform.


tgreier:
https://github.com/cra-ros-pkg/robot_localization/blob/ea976f9e65eac505f5cc7d3197ae080de4e5c10b/src/navsat_transform.cpp#L174

In our configuration, we use a manual datum and our world reference frame name is 'map' (not the default world_frame_id of 'odom'.
This results in an incorrect transform (cartesian -> world) being generated by navsat_transform.

This is a bit of a race condition because if our odometry callback is invoked before the transformCallback timer callback, then the world_frame_id is properly set, and the correct transform will be published.

ayrton04:
Would it not make sense to have logic to force the transform computation to wait until the odometry callback has fired?

tgreier:
@ayrton04 Yes, I thought about that. But that change seemed to be a fairly major alteration to how the manual datum is handled. I am happy to take a shot at changing the logic if you think that is the better solution.
I updated Pull Request #821 to wait for odom message before setting manual datum. Could you please review?

PR #821, opened 2023-06-22

tgreier:
Issue #820 

Add cfg params for default world and baselink frame names in the case where a manual datum is being used.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -103,8 +103,7 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
   use_local_cartesian_ = this->declare_parameter("use_local_cartesian", false);
   frequency = this->declare_parameter("frequency", frequency);
   delay = this->declare_parameter("delay", delay);
-  base_link_frame_id_ = this->declare_parameter("default_base_frame", "base_link");
-  world_frame_id_ = this->declare_parameter("default_world_frame", "odom");
+
   transform_timeout = this->declare_parameter("transform_timeout", transform_timeout);
 
   transform_timeout_ = tf2::durationFromSec(transform_timeout);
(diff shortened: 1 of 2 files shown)


src/navsat_transform.cpp
```diff
@@ -692,7 +706,7 @@ void NavSatTransform::odomCallback(
   world_frame_id_ = msg->header.frame_id;
   base_link_frame_id_ = msg->child_frame_id;
 
-  if (!transform_good_ && !use_manual_datum_) {
+  if (!transform_good_) {
```
tgreier, line 708: Removing the check here for use_manual_datum should be fine since setTransformOdometry() also checks this flag.
Removing the check allows setTransformOdometry to store the world and base frame names from the received odometry message.

src/navsat_transform.cpp
```diff
@@ -103,6 +103,7 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
   use_local_cartesian_ = this->declare_parameter("use_local_cartesian", false);
   frequency = this->declare_parameter("frequency", frequency);
   delay = this->declare_parameter("delay", delay);
+
```
ayrton04: Can we remove this line break?

ayrton04 (approved):

src/navsat_transform.cpp
tgreier: Done!

ayrton04:
Thanks!

tgreier:
@ayrton04 How do you manage merging changes from foxy-devel into the master branch or other later branches (such as humble)? Do I do that manually by creating pull requests for each branch?

ayrton04:
I'd normally cherry-pick them, but as we're deprecating this package soon, it'll be helpful if you'd just open PRs against each. Thanks!

PR #827, opened 2023-08-09
Add possibility to publish the inverse transform

carlos-m159:
This allows to publish the inverse transform. This feature is particularly useful when we need to work around the fact that a frame can only have a single parent.


ayrton04 (approved):
LGTM. Thanks, Carlos!

PR #830, opened 2023-09-01
Fix/set utm map frame change

MCFurry:
We found out that the `/setUTMZone` service works in fixing navsat_transform into a certain UTM zone. However, the published transform is not updated, which means that alignment errors can occur.

This PR fixes that issue. 

If desired I can create a similar PR to the ROS2 branch, but let's confirm first this is a desired fix as is.

### DIFF
--- a/include/robot_localization/navsat_transform.h
+++ b/include/robot_localization/navsat_transform.h
@@ -210,6 +210,10 @@ class NavSatTransform
     //!
     bool use_local_cartesian_;
 
+    //! @brief Whether we want to have our UTM zone fixed, and not change it if we move outside of it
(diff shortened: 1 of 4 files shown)


ayrton04:
Thanks. And yes, please PR to ROS 2.

MCFurry:
Thanks, I'll make a ROS2 PR as well!

MCFurry:
On that note, it seems that foxy-devel already switched to using geographiclib for UTM conversions: https://github.com/cra-ros-pkg/robot_localization/pull/643
But I see the default `ros2` branch did not (yet) ?

PR #833, opened 2023-09-25
Utm using geographiclib ros2 branch

Timple:
Extracting the answer from this question: https://github.com/cra-ros-pkg/robot_localization/pull/830#issuecomment-1733365235 from this comment: https://github.com/cra-ros-pkg/robot_localization/pull/821#issuecomment-1733350108

### DIFF
--- a/test/test_navsat_conversions.cpp
+++ b/test/test_navsat_conversions.cpp
@@ -38,16 +38,19 @@
 
 void NavsatConversionsTest(
   const double lat, const double lon,
-  const double UTMNorthing, const double UTMEasting, const std::string UTMZone)
+  const double UTMNorthing, const double UTMEasting,
+  const std::string UTMZone, const double gamma)
 {
   double UTMNorthing_new;
   double UTMEasting_new;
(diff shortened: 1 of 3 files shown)


test/test_navsat_conversions.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <gtest/gtest.h>
+
+#include <string>
+
+#include "robot_localization/navsat_conversions.hpp"
+
+void NavsatConversionsTest(
```
ayrton04, line 39: Did we confirm that these tests produce the same results before and after the change?

ayrton04 (approved):

test/test_navsat_conversions.cpp
Timple: Yes
But to make it explicit, I opened a separate PR that can be merged first if desired: https://github.com/cra-ros-pkg/robot_localization/pull/838

PR #835, opened 2023-10-09

tgreier:
Issue https://github.com/cra-ros-pkg/robot_localization/issues/820

Wait for odom callback before setting manual datum so that the reference frame names are properly set.

### DIFF
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -178,6 +178,11 @@ private:
     double & altitude) const;
 
   /**
+   * @brief Sets the manual datum pose to be used by the transform computation
+   */
+  void setManualDatum();
+
+  /**
    * @brief Frame ID of the robot's body frame
    *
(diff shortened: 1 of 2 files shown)


ayrton04:
As with the other PR, if you could fix the uncrustify errors, that would be great (they may not be from this PR, not sure).

tgreier:
@ayrton04 Hi! That is a large list of lint/uncrustify errors, none of which resulted from my changes. I have only changed 2 files, but the lint errors are across almost all files. I would be reluctant to make all of those changes on these PRs since these PRs are very small in scope. What do you recommend?

ayrton04:
Yeah, let's ignore it. Maybe before this package gets officially sunsetted, we'll do a separate PR to clean that all up.

ayrton04 (approved):

tgreier:
Hi @ayrton04. When will these two pull requests be merged? Thanks!

PR #836, opened 2023-10-09

tgreier:
Issue https://github.com/cra-ros-pkg/robot_localization/issues/820

Wait for odom callback before setting manual datum so that the reference frame names are properly set.


ayrton04 (approved):

ayrton04:
Would you mind fixing the lint/uncrustify errors?

PR #840, opened 2023-11-02
Melodic backport of https://github.com/cra-ros-pkg/robot_localization/pull/595

mqcmd196:
Cherry-pick https://github.com/cra-ros-pkg/robot_localization/pull/595 and adopt to `melodic-devel` branch.

### DIFF
--- a/doc/state_estimation_nodes.rst
+++ b/doc/state_estimation_nodes.rst
@@ -174,6 +174,10 @@ If *true*, the state estimation node will publish the transform from the frame s
 ^^^^^^^^^^^^^^^^^^^^^
 If *true*, the state estimation node will publish the linear acceleration state. Defaults to *false*.
 
+~permit_corrected_publication
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


SteveMacenski:
Melodic is EOL so no new binaries will be released even if this is merged. If you have to compile from source anyway to use that feature -- why not just use notice branch outright? There's no meaningful changes between those distributions to cause any API issues for you to compile `noetic-devel` against a melodic ROS version.

mqcmd196:
Thank you for your response
> Melodic is EOL so no new binaries will be released even if this is merged. If you have to compile from source anyway to use that feature -- why not just use notice branch outright? 

Yes,  but I want to let the necessary users know that this feature is available in Melodic by sending the PR. 

> There's no meaningful changes between those distributions to cause any API issues for you to compile noetic-devel against a melodic ROS version.

I cloned and built `robot_localization`'s `noetic-devel` branch on Ubuntu 18.04 ROS environment. Using the navigation, the robot showed strange behavior. I'll check deeply why it happens if I have time.

PR #852, opened 2023-12-19
fix nanosecond in header timestamp 

changx03:
The header timestamp has two components: sec and nanosec.
The code was intended to convert nanosec to sec but calling the wrong variable.
From:

```
  pose_tmp.stamp_ = tf2::timeFromSec(
    static_cast<double>(msg->header.stamp.sec) +
    static_cast<double>(msg->header.stamp.sec) / 1000000000.0);
```

To:

```
  pose_tmp.stamp_ = tf2::timeFromSec(
    static_cast<double>(msg->header.stamp.sec) +
    static_cast<double>(msg->header.stamp.nanosec) / 1000000000.0);
```

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2865,7 +2865,7 @@ bool RosFilter<T>::preparePose(
 
   pose_tmp.stamp_ = tf2::timeFromSec(
     static_cast<double>(msg->header.stamp.sec) +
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

ayrton04:
Yikes! That's a bad one. Good catch. Would you mind PR-ing this against `noetic-devel` as well, please?

ayrton04:
Never mind, this is ROS 2 only. Thanks.

PR #856, opened 2023-12-21
Feature/set utm service

Timple:
This is a forward port of both https://github.com/cra-ros-pkg/robot_localization/pull/627 and https://github.com/cra-ros-pkg/robot_localization/pull/830

Comparing these two commits to their respective PR's is probably the easiest to review :slightly_smiling_face:


src/navsat_transform.cpp
```diff
@@ -451,13 +456,20 @@ bool NavSatTransform::fromLLCallback(
...
   } else {
-    std::string utm_zone_tmp;
-    navsat_conversions::LLtoUTM(
-      latitude,
-      longitude,
-      cartesian_y,
-      cartesian_x,
-      utm_zone_tmp);
+    // std::string utm_zone_tmp;
+    // navsat_conversions::LLtoUTM(
```
SteveMacenski: Just delete it if its removed, don't comment out

SteveMacenski (approved):
I wasn't in the original discussions for this, but this seems reasonable to forward port

Timple:
Well, I'm not sure were it went wrong in the porting to ROS2, but here we are :slightly_smiling_face: 

Once upon a time, we even proposed to the `UTMtoLL` and `LLtoUTM` functions as they are pretty much empty now: https://github.com/cra-ros-pkg/robot_localization/pull/626#discussion_r598614404

Would you like a follow-up PR for this on the `ros2` branch? If we don't backport it to the releases, api compatibility is kept.

ayrton04 (approved):

ayrton04:
Thanks!

Issue #749, opened 2022-05-01
Example launch file for dual_ekf_navsat_example contains wrong remapping rule


splatter96:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries 
- Version or commit hash:
  - galactic versions in apt as of today

#### Steps to reproduce issue
Set the wait_for_datum paramter to false
Launch the provided dual_ekf_navsat_example.launch.py launch file with an appropriate simulation of sensor setup.

#### Expected behavior
All transformations, especially map-->odom and odom-->baseframe should be published with the correct covariance.

#### Actual behavior
All transformations get published, but the map-->odom covariance increases with the moved distance.
After issuing a call to the /datum service with an initial pose, the covariance behaves as expected.

#### Additional information
After looking though the code of the navsat_transform node it show, that it subsribes to the "imu" topic for measurements of the imu sensor and only start proper transformations after it receives information at least once from it.
The provided launch file includes a remapping from "imu/data" to "imu/data" this remapping should be from "imu" to "imu/data". After changing this I got the expected behaviour.

ayrton04:
Thanks for catching that. Any chance you could submit a PR?

SaiSugunSegu:
Hey ayrton04, 

Can I try and replicate the issue, and make the recommended mapping "imu/data" to "imu" and test it
?

ayrton04:
If you want to submit a PR, please do.

SaiSugunSegu:
As I see, there is no issue in the code.

Note: **when using the Robot Localization Repository - any launch file - users need to remap the default topic name to their own topic name (based on sensor output)**

SaiSugunSegu:
I don't feel its a bug, we can close this issue. 
@ayrton04 , please check and the close the issue, if you feel its a bug, please let me know, I want to work on it.

joeldushouyu:
Could it caused by the error in here?  @SaiSugunSegu 
[Error in dual_ekf_navsat_example](https://github.com/cra-ros-pkg/robot_localization/blob/ros2/launch/dual_ekf_navsat_example.launch.py#L59C1-L59C50)

Because in the source file, the imuCallback() listen on [/imu](https://github.com/cra-ros-pkg/robot_localization/blob/ros2/src/navsat_transform.cpp#L192).

I found out about this issue when I was using ros2 doctor to analyze. Because I could not get any message published on /odometry/gps after launching nav_sat .

[...]

PR #857, opened 2023-12-30
Remapp the correct IMU topic in dual_ekf_navsat_example.launch.py

joeldushouyu:
In [navsat_transform.cpp](https://github.com/cra-ros-pkg/robot_localization/blob/ros2/src/navsat_transform.cpp#L192l), it have a subscriber to listen to the imu message on"/imu". However, in the dual_ekf_navsat_example.launch.py file, it has the wrong [topic mapping](https://github.com/cra-ros-pkg/robot_localization/blob/ros2/launch/dual_ekf_navsat_example.launch.py#L59C1-L59C50)

For more detail: https://github.com/cra-ros-pkg/robot_localization/issues/749#issuecomment-1872434107

### DIFF
--- a/launch/dual_ekf_navsat_example.launch.py
+++ b/launch/dual_ekf_navsat_example.launch.py
@@ -56,7 +56,7 @@ def generate_launch_description():
             name='navsat_transform',
 	        output='screen',
             parameters=[parameters_file_path],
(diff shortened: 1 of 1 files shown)


ayrton04 (approved):

ayrton04:
Thanks!

PR #866, opened 2024-03-05
Fix throttle duration

Timple:
Some folks though milliseconds made much more sense than SI units. Leaving me to create all kinds of pull requests like this :smile:


Timple:
`test_ekf_node_interfaces` fails. Not sure how that can be caused by this?

ayrton04:
So much for REP-103.

That test fails sometimes. There's some numerical instability in it that I haven't had the cycles to run down.

ayrton04:
Thanks!

PR #876, opened 2024-04-03
TF Prefix Bug

rafal-gorecki:
The value of `tf_prefix` parameter is stored in unused variable `tf_prefix_path`. Please fix this.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -819,9 +819,8 @@ void RosFilter<T>::loadParams()
 
   // Try to resolve tf_prefix
   std::string tf_prefix = "";
-  std::string tf_prefix_path = "";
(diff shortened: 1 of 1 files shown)


ayrton04:
Ah, this is because in ROS 1, we were searching for the parameter key before loading it:

https://github.com/cra-ros-pkg/robot_localization/blob/7bebbed62c440b2643964e26d69668d3b18509c6/src/ros_filter.cpp#L848

But ROS 2 has no support for that. Technically, `tf2` [does not support ](https://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix)`tf_prefix`, but `r_l` manages this in a `tf2`-friendly way.

ayrton04 (approved):

ayrton04:
Thanks for the PR!

PR #878, opened 2024-04-17
Fix the build on rolling.

clalancette:
In particular, rolling now requires that you link against yaml-cpp::yaml-cpp, so do that here.

I will note that this will *not* work on Iron, so a new branch may need to be created for that.

This should fix the failing build as in https://build.ros2.org/view/Rbin_rhel_el964/job/Rbin_rhel_el964__robot_localization__rhel_9_x86_64__binary/44/console


ayrton04:
As luck would have it, I created an `iron-devel` branch just the other day and bloomed with it.

Thanks for this; I saw this error and was going to figure out why the yaml library linking was not behaving.

PR #884, opened 2024-05-08
Fix mixing of UTM and catesian transforms with in use_local_catesian mode

JayHerpin:
There were two places in the code where the UTM transform was being used instead of the local cartesian. These changes are basically just aligning the iron-devel branch with what is currently on noetic in terms of these two methods.

I would have also done this on the ros2 branch but I do not currently have a rolling build setup.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -461,9 +461,10 @@ bool NavSatTransform::fromLLCallback(
     bool northp_tmp;
 
     try {
-      GeographicLib::UTMUPS::Forward(latitude, longitude,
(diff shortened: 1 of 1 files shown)


SteveMacenski:
It should be done in rolling as well so that all new distributions have the work - else it will only ever be in `iron` that goes EOL in a few months. LGTM beyond linting problems

JayHerpin:
I haven't quit figured out a nice setup for emacs and ros workspaces where it plays nice with everyone's individual styles. Should be fixed now

JayHerpin:
ok to cherry pick onto jazzy after this one lands?

ayrton04 (approved):

PR #886, opened 2024-06-04
Resolve mixing of UTM and local transforms in local cartesian mode

JayHerpin:
Same as https://github.com/cra-ros-pkg/robot_localization/pull/884, but applied on the ros2 branch


SteveMacenski (approved):

ayrton04:
Thanks!

PR #888, opened 2024-06-18
Remove the diagnostic warning about multiple pose sources

svwilliams:
Remove the diagnostic warning about multiple pose sources. The general purpose of the EKF is to fuse together observations of the same variables from multiple sources.


ayrton04:
The reason for the warning is that people would often fuse multiple pose sources that had the same `frame_id`, but were not actually consistent (i.e., were not actually in the same frame). This causes the filter to jump back and forth between the two pose sources rapidly.

Regardless, the diagnostic error is wrong, and we shouldn't try to protect against bad configuration that way.

Issue #846, opened 2023-11-24
Odometry drift caused by acceleration drift removal in relative mode


rokusottervanger:
The following piece of code (introduced in #753), causes my odometry state to drift due to a nonzero resulting acceleration vector after removing the gravitational acceleration.

https://github.com/cra-ros-pkg/robot_localization/blob/804729e7bf7d1b83e2dc4b33fc8ff9cf4f2dd9a2/src/ros_filter.cpp#L2730-L2738

My `r_l` configuration is as follows
```yaml
  frequency: 10.0
  print_diagnostics: true
  base_link_frame: "base_link"
  world_frame: "odom"

  transform_time_offset: 0
  two_d_mode: false
  publish_tf: true
  smooth_lagged_data: true
  history_length: 5

  odom0: encoder_odom
  odom0_nodelay: true
  odom0_config: [false, false, false,   # X,Y,Z,
                false, false, false,  # roll,pitch,yaw,
                true,  true,  true,  # X`,Y`,Z`
                false, false, false,   # roll`,pitch`,yaw`,
                false, false, false]  # X``,Y``,Z``

  imu0: imu/data
  imu0_queue_size: 20
  imu0_remove_gravitational_acceleration: true
  imu0_nodelay: true
  imu0_relative: true
  imu0_config: [false, false, false,  # X,Y,Z,
                true,  true,  true,   # roll,pitch,yaw,
                false, false, false,  # X`,Y`,Z`,
                true,  true,  true,   # roll`,pitch`,yaw`,
                true,  true,  false]   # X``,Y``,Z``
```

I'm aware that no position data is being fused here, so odom might drift with a random walk, but I'm seeing a significant deterioration (and nonzero mean drift, so not purely random) caused by the referenced code. I'm not sure what the theory behind this piece of code is. Could someone please elaborate? @HaoguangYang maybe?

HaoguangYang:
Seems like you are using "relative" orientation to remove the gravity. The basis of line 2737 are three assumptions:
1. The [0, 0, 0, 1] orientation of base_link is "upright", with its z-axis pointing up and perpendicular to the ground plane.
2. Your IMU registers orientation [0, 0, 0, 1] when the base_link is "upright", as an origin of orientation (e.g. if an integral-based method is used to obtain roll, pitch, and yaw of the IMU, it should start with the **robot's base_link** being "upright"). However, your IMU frame may be rotated from base_link.
[...]

PR #892, opened 2024-08-03
bugfix: gravity vector now rotated to target_frame prior to removal

HaoguangYang:
Addresses Issue #846 when integrating IMU data for odometry.

In the previous version, the gravity vector was rotated to IMU frame instead of target_frame prior to gravity removal. However, acc_tmp has already been transformed into target_frame. This causes inconsistency and may have caused issue #846 .

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2713,9 +2713,8 @@ bool RosFilter<T>::prepareAcceleration(
           state(StateMemberPitch),
           state(StateMemberYaw));
 
-        // transform state orientation to IMU frame
-        trans.setBasis(stateTmp * target_frame_trans.getBasis());
-        rotNorm = trans.getBasis().inverse() * normAcc;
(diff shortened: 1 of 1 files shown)


ayrton04:
Thanks for this. It looks like it needs to be uncrustified, though:

https://build.ros2.org/job/Rpr__robot_localization__ubuntu_noble_amd64/9/testReport/

```
--- src/ros_filter.cpp
+++ src/ros_filter.cpp.uncrustify
@@ -2731 +2731 @@
-                    (trans.getBasis().inverse() * normAcc);
+            (trans.getBasis().inverse() * normAcc);
@@ -2739,2 +2739,2 @@
-                      (trans.getBasis().inverse() *
-                      (target_frame_trans.getBasis().inverse() * normAcc));
+            (trans.getBasis().inverse() *
+            (target_frame_trans.getBasis().inverse() * normAcc));
```

PR #896, opened 2024-08-23
Fix warnings when building against Rolling.

clalancette:
Rolling has moved header files (deprecating the old one), and changed the preferred form of some templated classes (like TimeSynchronizer).  Fix those warnings here.

Note that this should *only* be applied for Rolling, as it will fail to compile for earlier versions of ROS 2.

### DIFF
--- a/include/robot_localization/ros_robot_localization_listener.hpp
+++ b/include/robot_localization/ros_robot_localization_listener.hpp
@@ -37,8 +37,8 @@
 
 #include "Eigen/Dense"
 #include "geometry_msgs/msg/accel_with_covariance_stamped.hpp"
-#include "message_filters/subscriber.h"
-#include "message_filters/time_synchronizer.h"
+#include "message_filters/subscriber.hpp"
+#include "message_filters/time_synchronizer.hpp"
 #include "nav_msgs/msg/odometry.hpp"
 #include "rclcpp/qos.hpp"
 #include "rclcpp/rclcpp.hpp"
--- a/src/ros_robot_localization_listener.cpp
+++ b/src/ros_robot_localization_listener.cpp
@@ -70,9 +70,9 @@ RosRobotLocalizationListener::RosRobotLocalizationListener(
   rclcpp::SubscriptionOptions options)
 : qos1_(1),
   qos10_(10),
-  odom_sub_(node, "odom/filtered", qos1_.get_rmw_qos_profile(), options),
-  accel_sub_(node, "acceleration/filtered", qos1_.get_rmw_qos_profile(), options),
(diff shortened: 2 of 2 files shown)


SteveMacenski (approved):

SteveMacenski:
@ayrton04 is this test known to be flaky? I don't believe I've seen this failure before 

```
06:34:49 33: [test_ekf_node_interfaces-2] [ RUN      ] InterfacesTest.ImuDifferentialIO
06:34:49 33: [test_ekf_node_interfaces-2] /tmp/ws/src/robot_localization/test/test_se_node_interfaces.cpp:830: Failure
06:35:01 33: [test_ekf_node_interfaces-2] Expected: (std::abs(rollFinal)) < (0.2), actual: 1.2032755481031931 vs 0.2
06:35:01 33: [test_ekf_node_interfaces-2] 
06:35:01 33: [test_ekf_node_interfaces-2] /tmp/ws/src/robot_localization/test/test_se_node_interfaces.cpp:831: Failure
06:35:01 33: [test_ekf_node_interfaces-2] Expected: (std::abs(pitchFinal)) < (0.2), actual: 0.65010049931252023 vs 0.2
06:35:01 33: [test_ekf_node_interfaces-2] 
06:35:01 33: [test_ekf_node_interfaces-2] /tmp/ws/src/robot_localization/test/test_se_node_interfaces.cpp:832: Failure
06:35:01 33: [test_ekf_node_interfaces-2] Expected: (std::abs(yawFinal)) < (0.2), actual: 2.2377817033514558 vs 0.2
06:35:01 33: [test_ekf_node_interfaces-2] 
06:35:01 33: [ekf_node-1] [INFO] [1724420101.340558210] [test_se_node_interfaces]: Received a request to reset filter.
06:35:01 33: [test_ekf_node_interfaces-2] [  FAILED  ] InterfacesTest.ImuDifferentialIO (12004 ms)
```

ayrton04:
Whoops, please disregard this comment. I see this is rolling only.

ayrton04 (approved):

clalancette:
> Whoops, please disregard this comment. I see this is rolling only.

Yes, exactly.  I will point out that after this merge, it is no longer possible to release `robot_localization` into Humble, Iron, or Jazzy.  I might suggest making a branch prior to this commit to support those, but that is up to you.

clalancette:
> Yes, exactly. I will point out that after this merge, it is no longer possible to release `robot_localization` into Humble, Iron, or Jazzy. I might suggest making a branch prior to this commit to support those, but that is up to you.

Ah, never mind this.  There are already branches for other distributions, so we should be set here.  Thanks!

PR #897, opened 2024-08-29
Fixing IMU differential test

ayrton04:
I've been getting sick of this test failing. I'm not sure how it ever really passed, but regardless, this is a much cleaner way to do it. Now we set the filter to a specific pose (really just a specific orientation), then feed it the same orientation differentially to ensure it doesn't move, then slowly drag it back to 0 using interpolated quaternions.

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -1960,7 +1960,7 @@ void RosFilter<T>::poseCallback(
 
   RF_DEBUG(
     "------ RosFilter<T>::poseCallback (" << topic_name << ") ------\n"
-      "Pose message:\n" << msg);
+      "Pose message:\n" << geometry_msgs::msg::to_yaml(*msg));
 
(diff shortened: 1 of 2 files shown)


ayrton04:
Haha, so much for that. Going to add some debug statements to the test to see where it's going astray, since I can't replicate this locally.

ayrton04:
Narrowing the issue down. Don't yet understand why it's happening, but I can see where things are going south.

ayrton04:
OK, I think I know what's up. I need to think about how to address it. The TL;DR version is that it appears that even though I am publishing a message from publisher A and then another from publisher B, the filter is receiving publisher B's message first for some reason. Not sure if anyone is watching this PR, but does that track?

ayrton04:
OK, that method works. I don't love it, but it should get the test back on track. Will close and open this a couple more times to make sure it's consistent.

ayrton04:
Alright, going to merge this. Seems happy now.

PR #912, opened 2025-01-10
Added FromLLArray service

JayHerpin:
This is an update of https://github.com/cra-ros-pkg/robot_localization/pull/676 since that one seems to be abandoned.

This adds a service that allows conversions of lists of lat/lon pairs instead of having to request them one at a time.


include/robot_localization/navsat_transform.hpp
```diff
@@ -32,15 +32,18 @@
 #ifndef ROBOT_LOCALIZATION__NAVSAT_TRANSFORM_HPP_
 #define ROBOT_LOCALIZATION__NAVSAT_TRANSFORM_HPP_
 
-#include <memory>
+#include <stdexcept>
```
ayrton04, line 36: Nit: these were alphabetized within the groups (in this case, STL headers). Would you mind retaining that?

include/robot_localization/navsat_transform.hpp
```diff
@@ -149,13 +161,13 @@ class NavSatTransform : public rclcpp::Node
    * @brief Converts the odometry data back to GPS and broadcasts it
    * @param[out] filtered_gps The NavSatFix message to prepare
    */
-  bool prepareFilteredGps(sensor_msgs::msg::NavSatFix * filtered_gps);
+  bool prepareFilteredGps(sensor_msgs::msg::NavSatFix *filtered_gps);
```
ayrton04: The previously conformed to ROS 2 [standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html#pointer-syntax-alignment), but now doesn't. Would you mind changing it back?

include/robot_localization/navsat_transform.hpp
```diff
@@ -149,13 +161,13 @@ class NavSatTransform : public rclcpp::Node
...
 
   /**
    * @brief Prepares the GPS odometry message before sending
    * @param[out] gps_odom The odometry message to prepare
    */
-  bool prepareGpsOdometry(nav_msgs::msg::Odometry * gps_odom);
+  bool prepareGpsOdometry(nav_msgs::msg::Odometry *gps_odom);
```
ayrton04: As above

src/navsat_transform.cpp
```diff
@@ -502,6 +531,7 @@ bool NavSatTransform::setUTMZoneCallback(
   return true;
 }
 
+
```
ayrton04: Superfluous newline.

src/navsat_transform.cpp
```diff
@@ -437,9 +439,40 @@ bool NavSatTransform::fromLLCallback(
...
+                   [this] (const auto& point) { return fromLL(point); });
+  } catch(const std::runtime_error & e) {
+    return false;
+  }
+
+  response->map_points = std::move(converted_points);
+  return true;
+}
+
+geometry_msgs::msg::Point NavSatTransform::fromLL(
```
ayrton04, line 470: This throws an exception now. I just want to make sure that we're catching it everywhere it's called. I haven't looked through the code to remind myself. :)

ayrton04 (changes requested):
Thank you! Just a few minor things.

src/navsat_transform.cpp
JayHerpin: its called from:

1. NavSatTransform::fromLLCallback
2. NavSatTransform::fromLLArrayCallback

each of which does have it wrapped in a try catch

JayHerpin:
> Thank you! Just a few minor things.

No problem. Each of your comments should be handled now

ayrton04 (approved):

ayrton04:
Thanks!

PR #917, opened 2025-02-27
Fix user UTM conversion and catch GeographicErr to not crash navsat_transfrom node

MCFurry:
The service setUTM was introduced in ROS1 at first:
https://github.com/cra-ros-pkg/robot_localization/pull/627
And later also ported to ROS2:
https://github.com/cra-ros-pkg/robot_localization/pull/856

However, a small error was left here where the `gpsFixCallback()` does a forward conversion without specifying the utm-zone. So GeographicLib does the conversion based on lat/lon only. 
In case the lat/lon in the gpsfix message are in a neighbouring zone, utm-coordinates for that other zone are calculated and when `mapToLL()` converts back with the saved `utm_zone_`, also incorrect lat/lon values are produced.

You can spot the difference here, where in the original introduction `utm_zone_` was used:
https://github.com/cra-ros-pkg/robot_localization/pull/627/files#diff-e2f5adb2665b53274a429c05d5f267ae428aee0cb9e5fdb058f46cb3993a9617R636
But in the ROS2 port it dropped out:
https://github.com/cra-ros-pkg/robot_localization/pull/856/files#diff-e2f5adb2665b53274a429c05d5f267ae428aee0cb9e5fdb058f46cb3993a9617R668

Moreover, callers to `mapToLL()` are now also protected from crashing the node in case of an exception.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -428,14 +428,9 @@ bool NavSatTransform::toLLCallback(
   }
   tf2::Vector3 point(request->map_point.x, request->map_point.y,
     request->map_point.z);
-  try {
-    mapToLL(
-      point, response->ll_point.latitude, response->ll_point.longitude,
-      response->ll_point.altitude);
-  } catch (const GeographicLib::GeographicErr & e) {
-    RCLCPP_ERROR_STREAM(this->get_logger(), e.what());
-    return false;
-  }
+  mapToLL(
+    point, response->ll_point.latitude, response->ll_point.longitude,
+    response->ll_point.altitude);
 
   return true;
 }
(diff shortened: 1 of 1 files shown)


src/navsat_transform.cpp
```diff
@@ -428,9 +428,14 @@ bool NavSatTransform::toLLCallback(
...
-  mapToLL(
-    point, response->ll_point.latitude, response->ll_point.longitude,
-    response->ll_point.altitude);
+  try {
+    mapToLL(
+      point, response->ll_point.latitude, response->ll_point.longitude,
+      response->ll_point.altitude);
+  } catch (const GeographicLib::GeographicErr & e) {
+    RCLCPP_ERROR_STREAM(this->get_logger(), e.what());
+    return false;
```
Timple: I don't think this does anything. This function should be (and is probably casted to) a `void` return type.

We could perhaps set the values to NAN?

MCFurry: This is a service callback, so by returning false the caller knows it didn't succeed.

The call to `mapToLL()` can throw an exception which we now catch here.

You can easily try this with a
 `ros2 service call /toLL robot_localization/srv/ToLL "map_point:
  x: 9999999.0
  y: 9999999.0
  z: 0.0"`
  Before the fix this would crash the node with the GeoGraphicLib exception, right now it doesn't crash but prints an error and returns false.

Timple (approved):

src/navsat_transform.cpp
MCFurry: Returning NaNs now instead to make clear a function call doesn't return anything useful in that case ;-)

Timple (approved):

PR #918, opened 2025-02-27
Same UTM fix on rolling

MCFurry:
Same as: https://github.com/cra-ros-pkg/robot_localization/pull/917/ but on rolling branch

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -553,15 +553,19 @@ void NavSatTransform::mapToLL(
 
     altitude = odom_as_cartesian.getOrigin().getZ();
   } else {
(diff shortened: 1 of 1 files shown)


Timple:
I implemented this originally in this PR: https://github.com/cra-ros-pkg/robot_localization/pull/627/files#diff-e2f5adb2665b53274a429c05d5f267ae428aee0cb9e5fdb058f46cb3993a9617R442

Seems I only implemented the try-catch on the forward part: https://github.com/cra-ros-pkg/robot_localization/pull/856/files#diff-e2f5adb2665b53274a429c05d5f267ae428aee0cb9e5fdb058f46cb3993a9617R888

But this makes sense!

SteveMacenski (approved):

PR #920, opened 2025-03-11
Add a Parameters Callback to set magnetic_declination_radians at runtime

EnzoGhisoni:
Hi,
For my use case, I need to be able to set the magnetic declination at run time as the robot can be started in different geographic areas. 
The PR introduces a parameter callback to update the magnetic_declination_radians parameter value.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -916,6 +916,9 @@ rcl_interfaces::msg::SetParametersResult NavSatTransform::parametersCallback(
   for (const auto & param: parameters) {
     if (param.get_name() == "magnetic_declination_radians") {
       magnetic_declination_ = param.as_double();
+
+      // Set back transform_good_ to false to recalculate the transform
+      transform_good_ = false;
       RCLCPP_INFO(
         this->get_logger(), "The new magnetic declination is (%f) rads", magnetic_declination_);
     }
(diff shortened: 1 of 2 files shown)


src/navsat_transform.cpp
```diff
@@ -903,4 +906,21 @@ void NavSatTransform::setTransformOdometry(
...
+rcl_interfaces::msg::SetParametersResult NavSatTransform::parametersCallback(
+  const std::vector<rclcpp::Parameter> & parameters)
+{
+  rcl_interfaces::msg::SetParametersResult result;
+  result.successful = true;
+  result.reason = "success";
+  // Here update class attributes
+  for (const auto & param: parameters) {
+    if (param.get_name() == "magnetic_declination_radians") {
+      magnetic_declination_ = param.as_double();
```
ayrton04, line 918: This will update the `magnetic_declination_` variable, but the transform won't get recalulated with the new value, right? You'd want to set `transform_good_` back to `false`.

https://github.com/cra-ros-pkg/robot_localization/blob/a9c7e7cac5efee1d0cffa32c3c434d16c42ed94c/src/navsat_transform.cpp#L218C8-L218C23

EnzoGhisoni: You are right, in my use case I set a new datum right after but i can definitely update the transform_good_ variable at the same time to recalculate the transform.
I will do the change.

EnzoGhisoni: I did and test the modification. The recalculation is performed after the magnetic_declination_ parameter has been setted.

ayrton04 (approved):

PR #923, opened 2025-03-20
feat: add diff based kinematic calc

sees1:
- new class JacobianCreator
- typename T to std::shared_ptr<T>
- new enum in filter_common.hpp


sees1:
The changes are not completed and not tested, but it is possible to build a version. This was done for a robot project on a platform with Ackermann geometry. If it's relevant, I can finish and test it.

Issue #927, opened 2025-05-21
Confusion regarding setManualDatum


tahakenar:
Hi!

I am porting our software stack from ROS1 to ROS2 and confused about the implementation of ```use_manual_datum``` in [navsat_transform.cpp](https://github.com/cra-ros-pkg/robot_localization/blob/ros2/src/navsat_transform.cpp). I used to set ```wait_for_datum``` parameter to True and call set datum service when I'm okay with the position & covariance of my robot (ROS Noetic). In new implementation, navsat_transform_node sets datum before I call set datum service.

In perodic callback:
```

void NavSatTransform::transformCallback()
{
  if (!transform_good_) {
    computeTransform();

    if (transform_good_ && !use_odometry_yaw_ && !use_manual_datum_) {
      // Once we have the transform, we don't need the IMU
      imu_sub_.reset();
    }
```

After it goes into computeTransform it sets its manual datum by itself.
```
void NavSatTransform::computeTransform()
{
  // When using manual datum, wait for the receive of odometry message so
  // that the base frame and world frame names can be set before
  // the manual datum pose is set. This must be done prior to the transform computation.
  if (!transform_good_ && has_transform_odom_ && use_manual_datum_) {
    setManualDatum();
  }

  // Only do this if:
  // 1. We haven't computed the odom_frame->cartesian_frame transform before
  // 2. We've received the data we need
  if (!transform_good_ && has_transform_odom_ && has_transform_gps_ &&
    has_transform_imu_)
...
```

I could not get the implementation reason of setManualDatum in here.
```
  if (!transform_good_ && has_transform_odom_ && use_manual_datum_) {
    setManualDatum();
  }
```

ayrton04:
Would you mind asking questions on robotics.stackexchange.com? Thanks.

ayrton04:
Apologies, I see how this is an issue. It looks like it was introduced here:

https://github.com/cra-ros-pkg/robot_localization/pull/836/files

Conveniently, I _think_ this PR solves the issue:

https://github.com/cra-ros-pkg/robot_localization/issues/927

tahakenar:
I cannot replicate the issue right now but this PR seems like it solves the issue. Reviewed, and thanks.

PR #929, opened 2025-06-02
Don't process GNSS messages before SetDatum has been called

karl-schulz:
When the NavsatTransform node is starting with `wait_for_datum`, the is the possibility of a race condition. We had nodes crashing because of this. My interpretation of the race condition is this:

1) The robot publishes a fix before the `SetDatum` service is called
2) The NSTF tries to transform it to the map frame, but the map frame (still assumed at lat=0, lon=0) is far away from the received fix.
3) As a result, the NSTF publishes a odometry/gps with very high x/y values (in the range of 40km).
4) The EKF node (from this package) receives this message and sends the robot far away from the map frame, publishing a `odometry/filtered/global` an map->base_footprint TF message with high x/y values.
5) The NavsatTransform then receives `odometry/filtered/global` with it's X/Y far outside the UTM origin/range, crashing.

This is the error log:
```
localization  | [navsat_transform_node-6] [INFO] [1745838634.424235191] [navsat_transform]: Datum set.
localization  | [navsat_transform_node-6] terminate called after throwing an instance of 'GeographicLib::GeographicErr'
localization  | [navsat_transform_node-6]   what():  Latitude 48.1346d more than 20d from N pole
localization  | [ERROR] [navsat_transform_node-6]: process has died [...]
```

I think this is because of theses issues:
* The NavsatTransform should not publish `odometry/gps` before one of these conditions is met:
  * `wait_for_datum` parameter is `False`
  * The SetDatum is called with the initial datum
* The NavsatTransform should not do any kind of processing of incoming messages before the datum is set

I fixed it in this PR and tested locally in a regression test that isolates the race condition. I also added some more debugging logs to track issues better - I can also leave that part out if you want.

I'd be very happy about feedback if I got the issue right and if somebody else is having similar issues.

### DIFF
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -422,6 +422,11 @@ private:
   bool use_manual_datum_;
 
   /**
+   * @brief Whether we got a datum from the datum service or parameters
+   */
+  bool datum_is_set_;
+
+  /**
    * @brief Whether we get the transform's yaw from the odometry or IMU source
    */
   bool use_odometry_yaw_;
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -82,6 +82,7 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
   use_local_cartesian_(false),
   force_user_utm_(false),
   use_manual_datum_(false),
+  datum_is_set_(false),
   use_odometry_yaw_(false),
   cartesian_broadcaster_(*this),
   utm_meridian_convergence_(0.0),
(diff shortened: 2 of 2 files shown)


ayrton04:
Might be a stupid question, but why can't we just add this right [inside](https://github.com/cra-ros-pkg/robot_localization/blob/a837f8a2414adb653a5a64ea49a170fba447c09d/src/navsat_transform.cpp#L704) the `gpsFixCallback`?

```
if (use_manual_datum_ && !(has_transform_gps_ && has_transform_imu_))
{
  return;
}
```

If we have `wait_for_datum` (`use_manual_datum_`) enabled, the only way for `has_transform_gps_` and `has_transform_imu_`  to become true is for the datum parameter to be set or the service to be called with the datum parameter. So if we get any GPS poses before then, we can ignore them.

ayrton04:
Actually, I take it back. This is the right way to do this. It also solves https://github.com/cra-ros-pkg/robot_localization/issues/927.

ayrton04:
@karl-schulz would you mind rebasing this on `jazzy-devel` and making sure it passes the build farm checks?

karl-schulz:
Hi @ayrton04 , OK, will do!

tahakenar (approved):
Like @ayrton04 said it seems like it also solves #927 as well. Thanks for contributing.

karl-schulz:
Sorry everybody, I currently don't have capacity to port it. I'm not working on GNSS projects anymore. I hope somebody can pick up the PR here, as it's a really tricky issue and did cost us a lot of time.

karl-schulz:
Closing as I don't have capacity to merge this.

Issue #934, opened 2025-07-31
Services creates naming conflicts when running multiple EKF instances


ari-bw:
The "toggle", "set_pose" and "enable" services in the EKF localization node are created with a fixed name rather than being properly namespaced. This causes conflicts when running multiple EKF instances.

Wouldn't it make more sense to change the service names from "toggle" to "~/toggle"? This would automatically namespace the service under the node's name, following ROS2 conventions, and prevent conflict if multiple EKF's are created.

Currently, users must work around this by:
Using namespaces when launching nodes
Manually remapping service names
Launching nodes in separate ROS domains

However, these workarounds add complexity and should not be necessary for what should be standard node-specific functionality.

ayrton04:
I think `toggle` and `enable` would make sense for the local node handle, but not `set_pose`. That is an API that I think is independent of the state estimation/localization node being used.

Feel free to submit a PR!

ari-bw:
Great. I created a PR https://github.com/cra-ros-pkg/robot_localization/pull/938

ayrton04:
#940 merged. Thanks!

PR #938, opened 2025-08-14
Set "enable" and "toggle" services to be relative to the node's name

ari-bw:
This issue describes why this change is required: https://github.com/cra-ros-pkg/robot_localization/issues/934


ayrton04:
Thanks for the PR! However, this will need to go into `rolling-devel`. This is an API change to an LTS release, and if anyone's system was relying on the namespaces being where they are, that will break.

PR #948, opened 2025-09-29
Adding publishing of an REP-105 compliant ECEF transform

JayHerpin:
The navsat_transform_node can now publish an ecef earth frame. 

This functionality can be enabled by setting 'broadcast_earth_transform' to true. The frame ID used for the earth transform can be customized using the `earth_frame_id` parameter.

Note, that publishing the earth transform is only valid if use_local_cartesian is set to true.

### DIFF
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -60,6 +60,13 @@
 namespace robot_localization
 {
 
+/// \brief Computes the ECEF to ENU transform for a given geodetic origin
+/// \param lat0_deg Latitude of origin in degrees
+/// \param lon0_deg Longitude of origin in degrees
+/// \param h0_m Altitude of origin in meters
+/// \return Transform from ECEF (earth) frame to local ENU frame
+tf2::Transform computeEarthToCartesian(double lat0_deg, double lon0_deg, double h0_m);
+
 class NavSatTransform : public rclcpp::Node {
 public:
   /**
@@ -210,13 +217,6 @@ private:
    * @brief Sets the manual datum pose to be used by the transform computation
    */
   void setManualDatum();
-
-  /**
-   * @brief Computes earth to cartesian transform
-   * @note Only valid when use_local_cartesian_ is true
-   */
-  void computeEarthToCartesian(double lat0_deg, double lon0_deg, double h0_m);
-
 
   /**
    * @brief Frame ID of the robot's body frame
@@ -524,7 +524,6 @@ private:
    *       Earth transforms are not supported with UTM coordinates.
    */
   bool broadcast_earth_transform_;
-
 
   /**
    * @brief Transform from earth to cartesian (ECEF to local/UTM)
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -212,6 +212,13 @@ private:
   void setManualDatum();
 
   /**
+   * @brief Computes earth to cartesian transform
+   * @note Only valid when use_local_cartesian_ is true
+   */
+  void computeEarthToCartesian(double lat0_deg, double lon0_deg, double h0_m);
+
+
+  /**
    * @brief Frame ID of the robot's body frame
    *
    * This is needed for obtaining transforms from the robot's body frame to the
@@ -439,9 +446,9 @@ private:
   bool use_odometry_yaw_;
 
   /**
-   * @brief Used for publishing the static world_frame->cartesian transform
+   * @brief Used for publishing static transforms
    */
-  tf2_ros::StaticTransformBroadcaster cartesian_broadcaster_;
+  tf2_ros::StaticTransformBroadcaster tf_broadcaster_;
 
   /**
    * @brief UTM's meridian convergence
@@ -505,6 +512,32 @@ private:
    * can be set.
    */
   geographic_msgs::msg::GeoPose manual_datum_geopose_;
+
+  /**
+   * @brief Earth frame ID for ECEF coordinates
+   */
+  std::string earth_frame_id_;
+
+  /**
+   * @brief Whether or not we broadcast the earth transform
+   * @note Only valid when use_local_cartesian_ is true.
+   *       Earth transforms are not supported with UTM coordinates.
+   */
+  bool broadcast_earth_transform_;
+
+
+  /**
+   * @brief Transform from earth to cartesian (ECEF to local/UTM)
+   * Note: This is the mathematical transform E2C, inverted when published to TF
+   */
+  tf2::Transform earth_cartesian_transform_;
+
+  /**
+   * @brief Origin GPS coordinates for earth frame computation
+   */
+  double origin_latitude_;
+  double origin_longitude_;
+  double origin_altitude_;
 };
 
 }  // namespace robot_localization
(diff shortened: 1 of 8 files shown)


JayHerpin:
I should disclosed that AI was used in the creation of this patch... although, frankly, I don't know how much credit I want to give it; it did a pretty bad job at figuring out the transforms and I had to help it a lot.

JayHerpin:
Also, as far as testing, I actually did the work on Jazzy and tested there. If/when this review is accepted, I will also put up for review the push to the jazzy and kilted branches.

include/robot_localization/navsat_transform.hpp
```diff
@@ -211,6 +211,13 @@ class NavSatTransform : public rclcpp::Node {
    */
   void setManualDatum();
 
+  /**
+   * @brief Computes earth to cartesian transform
+   * @note Only valid when use_local_cartesian_ is true
+   */
+  void computeEarthToCartesian(double lat0_deg, double lon0_deg, double h0_m);
+
+
```
ayrton04: Nit: superfluous line break.

include/robot_localization/navsat_transform.hpp
```diff
@@ -505,6 +512,32 @@ class NavSatTransform : public rclcpp::Node {
...
+  std::string earth_frame_id_;
+
+  /**
+   * @brief Whether or not we broadcast the earth transform
+   * @note Only valid when use_local_cartesian_ is true.
+   *       Earth transforms are not supported with UTM coordinates.
+   */
+  bool broadcast_earth_transform_;
+
+
```
ayrton04: Nit: superfluous line break.

src/navsat_transform.cpp
```diff
@@ -1038,4 +1101,38 @@ rcl_interfaces::msg::SetParametersResult NavSatTransform::parametersCallback(
...
+  // Create transform to match LocalCartesian behavior
+  // LocalCartesian does: R * (input - origin)
+  // TF2 does: R * input + translation
+  // Therefore: translation = -R * origin
+  tf2::Vector3 origin_ecef(X0, Y0, Z0);
+  tf2::Vector3 translated_origin = -(R_ecef_to_enu * origin_ecef);
+  tf2::Transform E2C(R_ecef_to_enu, translated_origin);
+
+  // Store the transform (earth -> local_enu)
+  earth_cartesian_transform_ = E2C;
```
ayrton04: Instead of making this method `void`, can we make it return a transform? You could even then make the method static, which should simplify testing.

```
earth_cartesian_transform_ = computeEarthToCartesian(origin_latitude_, origin_longitude_, origin_altitude_);
```

ayrton04 (changes requested):
This looks really great, thank you! I haven't had the cycles to validate the math, so I'm going to have to trust that it's functional. Assuming this works, I'd be a big fan of deprecating anything related to UTM.

src/navsat_transform.cpp
JayHerpin: made this change, and added a few trivial unit tests to exercise it

peci1:
This is certainly an interesting addition. I still think ECEF is a bit underused in ROS and I'd like to see it used more!

However, it rather seems to me the overall approach to TF "outputs" of navsat_transform node could be changed. Currently (without this PR), there are options `broadcast_cartesian_transform`, `broadcast_cartesian_transform_as_parent_frame` and `use_local_cartesian` which all enable or disable certain TFs.

Adding another parameter that even limits the possible values of some of these seems to be increasing confusion.

I suggest restructuring this to something like this list of parameters:

- `broadcast_utm_to_world_transform`
- `broadcast_world_to_utm_transform`
- `broadcast_local_cartesian_to_world_transform`
- `broadcast_world_to_local_cartesian_transform`
- `broadcast_ecef_to_world_transform`
- `broadcast_world_to_ecef_transform`
- (possibly) `broadcast_utm_to_local_cartesian_transform`
- (possibly) `broadcast_local_cartesian_to_utm_transform`
- (possibly) also TFs `ecef`<->`utm`, `ecef`<->`local_cartesian`
- (no longer needed) ~~`use_local_cartesian`~~ (LLtoMap would be computed via ECEF)

Of course, from each pair of inverse parameters, only one could be set.

Many of these conversions are implemented in https://github.com/gazebosim/gz-math/blob/gz-math9/include/gz/math/SphericalCoordinates.hh . With the addition of [gz_math_vendor](https://index.ros.org/p/gz_math_vendor/#jazzy) to Jazzy+, this could simplify some of these computations. UTM is not covered, however.

---

As a side note, when digging into this, I always had this idea of having a *many-frames* UTM representation. So the UTM frame published by this node could actually be named `utm_33q` or similar. Then, there could be a very nice chain of transforms: `ecef->utm_XY->enu(local cartesian referenced to North)->world->odom->base_link`. At first sight, this would create a multi-parent graph in case the robot changes zone, but if the `utm_XY->world` transform were dynamic instead of static, it could be rather treated as a reparenting operation rather than multiparenting. And for cases when the robot never leaves its zone, the TF could easily be static. For this to work well, I think it'd be needed to use both latitude and longitude zones, otherwise the approximation error would be too large. To detect whether the user wants the UTM frame named `utm` or `utm_XY`, there could be a simple string-template parameter that would be evaluated. It would default to `utm`, but it would also allow e.g. `utm_{lat_zone}_{lon_zone}` or even `my_weird_{lon_zone}_utm` and similar. The other frames' names could also be configurable like this, but with no template parameters (local cartesian, ECEF).

Timple:
That looks well thought through!

Ideally we can choose to force a UTM frame instead of reparenting. For UTM it's ok to transform if you're a few kilometers over a UTM border. A lot of robotic applications run the risk of operating across a border. But most will be within some kilometer range.

In that case it's easier to stick to one UTM frame instead of flipping everytime you cross the border.

peci1:
I definitely don't suggest disabling sticky UTM zone. It's very practical and should still be configurable.

JayHerpin:
> As a side note, when digging into this, I always had this idea of having a _many-frames_ UTM representation. So the UTM frame published by this node could actually be named `utm_33q` or similar. Then, there could be a very nice chain of transforms: `ecef->utm_XY->enu(local cartesian referenced to North)->world->odom->base_link`. At first sight, this would create a multi-parent graph in case the robot changes zone, but if the `utm_XY->world` transform were dynamic instead of static, it could be rather treated as a reparenting operation rather than multiparenting. And for cases when the robot never leaves its zone, the TF could easily be static. For this to work well, I think it'd be needed to use both latitude and longitude zones, otherwise the approximation error would be too large. To detect whether the user wants the UTM frame named `utm` or `utm_XY`, there could be a simple string-template parameter that would be evaluated. It would default to `utm`, but it would also allow e.g. `utm_{lat_zone}_{lon_zone}` or even `my_weird_{lon_zone}_utm` and similar. The other frames' names could also be configurable like this, but with no template parameters (local cartesian, ECEF).

I've certainly had that same thought about the UTM frame name. IMO, that would make a lot of sense.

JayHerpin:
@ayrton04 thoughts? I like some of @peci1's comments, but am not inclined to do them as part of this review without some explicit direction. I'd be happy to do them as a follow up as well.

ayrton04:
I'll try to look at this again this week.

ayrton04:
I don't want to be too prescriptive about all the various suggestions for frames and such. All I can say is that I'd like to keep the package as REP-105 compliant as possible, and that mandates an ECEF-based `earth` frame as the "top" frame, then `map`, then `odom`. But we obviously need to continue to support all the use cases that we do now, so it won't make sense to try to remove, e.g., UTM support.

In the meantime, let's merge this, and @peci1, if you want to make a ticket for your frame idea, that would be great. We can move any discussion there.

Thanks again for this PR!

ayrton04 (approved):

PR #954, opened 2025-12-17
Fix additional negative time issues in ROS Kilted

JayHerpin:
Fixes a few more cases that https://github.com/cra-ros-pkg/robot_localization/pull/943 missed

### DIFF
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -532,7 +532,7 @@ bool RosFilter<T>::getFilteredAccelMessage(
     }
 
     // Fill header information
-    message->header.stamp = rclcpp::Time(filter_.getLastMeasurementTime());
+    message->header.stamp = filter_.getLastMeasurementTime();
     message->header.frame_id = base_link_output_frame_id_;
   }
 
@@ -2965,7 +2965,7 @@ bool RosFilter<T>::preparePose(
   tf2::Transform target_frame_trans;
   bool can_transform = ros_filter_utilities::lookupTransformSafe(
     tf_buffer_.get(), final_target_frame, pose_tmp.frame_id_,
-    rclcpp::Time(tf2::timeToSec(pose_tmp.stamp_)), tf_timeout_,
+    rclcpp::Time(tf2::timeToSec(pose_tmp.stamp_), RCL_ROS_TIME), tf_timeout_,
     target_frame_trans);
 
   // handling multiple odometry origins: convert to the origin adherent to base_link.
@@ -2975,7 +2975,7 @@ bool RosFilter<T>::preparePose(
   if (source_frame != base_link_frame_id_) {
     can_src_transform = ros_filter_utilities::lookupTransformSafe(
       tf_buffer_.get(), source_frame, base_link_frame_id_,
-      rclcpp::Time(tf2::timeToSec(pose_tmp.stamp_)), tf_timeout_,
+      rclcpp::Time(tf2::timeToSec(pose_tmp.stamp_), RCL_ROS_TIME), tf_timeout_,
       source_frame_trans);
   }
 
--- a/src/ros_filter.cpp
+++ b/src/ros_filter.cpp
@@ -2320,7 +2320,10 @@ void RosFilter<T>::periodicUpdate()
 
   // Clear out expired history data
   if (smooth_lagged_data_) {
-    clearExpiredHistory(filter_.getLastMeasurementTime() - history_length_);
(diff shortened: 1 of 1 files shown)


src/ros_filter.cpp
```diff
@@ -2965,7 +2968,7 @@ bool RosFilter<T>::preparePose(
   tf2::Transform target_frame_trans;
   bool can_transform = ros_filter_utilities::lookupTransformSafe(
     tf_buffer_.get(), final_target_frame, pose_tmp.frame_id_,
-    rclcpp::Time(tf2::timeToSec(pose_tmp.stamp_)), tf_timeout_,
+    rclcpp::Time(tf2::timeToSec(pose_tmp.stamp_), RCL_ROS_TIME), tf_timeout_,
```
SteveMacenski, line 2974: Shouldn't we use the `node->get_clock()` type of time? This assumes simulation time.

src/ros_filter.cpp
```diff
@@ -2320,7 +2320,10 @@ void RosFilter<T>::periodicUpdate()
 
   // Clear out expired history data
   if (smooth_lagged_data_) {
-    clearExpiredHistory(filter_.getLastMeasurementTime() - history_length_);
+    const rclcpp::Time last_measurement_time = filter_.getLastMeasurementTime();
+    if (last_measurement_time.nanoseconds() > history_length_.nanoseconds()) {
```
SteveMacenski, line 2327: I think you can just do a direct `rclcpp::Time` comparison https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Time.html

src/ros_filter.cpp
```diff
@@ -532,7 +532,7 @@ bool RosFilter<T>::getFilteredAccelMessage(
     }
 
     // Fill header information
-    message->header.stamp = rclcpp::Time(filter_.getLastMeasurementTime());
+    message->header.stamp = filter_.getLastMeasurementTime();
```
SteveMacenski: Any functional change here or just clarity? I don't see any changes since that cast will happen implicitly, but wanted to make sure there isn't something subtle here we should be aware of

src/ros_filter.cpp
JayHerpin: Its my understanding that when not run in simulation mode RCL_ROS_TIME reports the same as RCL_SYSTEM_TIME. Is that incorrect? At least some of the choices made here were matching the conventions from this commit https://github.com/cra-ros-pkg/robot_localization/commit/c20b934daff5269e32dfb6ad646af1423b8e01f4 which seems to be the first pass at fixing some of these issues

src/ros_filter.cpp
JayHerpin: history_length_ is actually a rclcpp::Duration, which is not directly comparable with rclcpp::Time... but given your response, clearly the code needs a clarifying comment about why this check exists at all given you wouldn't normally compare a duration and time

src/ros_filter.cpp
ayrton04: https://design.ros2.org/articles/clock_and_time.html

> The ROSTime will report the same as SystemTime when a ROS Time Source is not active. When the ROS time source is active ROSTime will return the latest value reported by the Time Source. ROSTime is considered active when the parameter use_sim_time is set on the node.

So this is correct, and is the same mechanism as ROS 1, I believe. What confuses me is why the `rclcpp::Time` [API](https://api.nav2.org/jazzy/html/classrclcpp_1_1Time.html) defaults to `SYSTEM_TIME` for any constructor except the one that takes a `Time` message. It also seems to violate the design documentation:

> It is expected that the default choice of time will be to use the ROSTime source, however the parallel implementations supporting steady_clock and system_clock will be maintained for use cases where the alternate time source is required.

ayrton04 (approved):

JayHerpin:
I'm not clear if there is a requested change for me here or not. Is there anything I can do to move this along?

ayrton04:
No, sorry, just never merged it.

I need to revisit the time management in this package. When I developed the workshop for ROSCon UK last year, I found that it didn't work well with bags/sim. I [changed the node's clock type](https://github.com/cra-ros-pkg/robot_localization/pull/943/changes), but I think I read somewhere that this should not be done unless you know what you're doing. This leads me to think I should have gone about it in a different way.

ayrton04:
Would you mind PR-ing into rolling as well? Thanks!

PR #955, opened 2025-12-26
Fix sim clock deadlock (#3)

kaidokert:
Prevent EKF and NavSat nodes from deadlocking on startup when using simulated time by avoiding clock wait blocks and validating clock readiness at runtime. The issue was introduced in #943

Bug Fixes:
  - Resolve a startup deadlock when use_sim_time is enabled by eliminating blocking waits for the ROS clock during EKF and NavSat node initialization and instead skipping processing until the clock has started.

Tests:
  - Add a C++ unit test to verify EKF and NavSat initialization do not block when use_sim_time is enabled.
  - Add integration tests plus supporting launch and helper scripts to ensure the EKF publishes filtered odometry both when the simulated clock starts before the EKF node and when it starts after.

### DIFF
--- a/src/navsat_transform.cpp
+++ b/src/navsat_transform.cpp
@@ -145,10 +145,9 @@ NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
       broadcast_cartesian_transform_as_parent_frame_);
   }
 
-  if (!this->get_clock()->started()) {
-    RCLCPP_INFO(this->get_logger(), "Waiting for clock to start...");
-    this->get_clock()->wait_until_started();
-  }
+  // Do not block here waiting for the clock to start, as it can deadlock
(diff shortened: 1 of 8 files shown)


src/navsat_transform.cpp
peci1, line 229: I think this already needs a working clock...

src/ros_filter.cpp
peci1, line 2117: This also needs a working clock.

ayrton04:
Sorry, there have been other clock-related changes in this package; do we know if this is still needed?

kaidokert:
> Sorry, there have been other clock-related changes in this package; do we know if this is still needed?

I'll review, re-test and report back. I've been running sims with this patch for past several months and it's been quite stable.

PR #959, opened 2026-02-09
Implement Lifecycle Node Support for EKF and UKF nodes

### LINKED ISSUE #958: Implement Lifecycle Node Support for EKF and UKF nodes
I am proposing the addition of ROS 2 Lifecycle (managed node) support for the ekf_node and ukf_node within the humble-devel branch. In modern production robotics stacks—particularly those leveraging Nav2—deterministic startup sequences and explicit state management are critical. Currently, robot_localization nodes begin processing and publishing transforms immediately upon startup. Transitioning these to Lifecycle Nodes will allow system architects to ensure that state estimation only begins once sensor drivers are verified as "Active," providing a more robust and predictable system initialization.

    Deterministic Startup: Ensures the filter only processes data once all sensor inputs are ready and "Active."
    Operational Control: Provides the ability to cleanly pause and resume state estimation via the Inactive state without              killing the node process.

BOOPESH-foxy:
## Changes
Adds ROS 2 managed lifecycle support for EKF/UKF state estimation nodes.

- Nodes inherit from `rclcpp_lifecycle::LifecycleNode`
- New `lifecycle_managed_node` parameter (default: **false**)
  - `false`: nodes auto-configure/activate (backward compatible)
  - `true`: nodes require external lifecycle management via `ros2 lifecycle` commands
- Implements lifecycle callbacks: on_configure, on_activate, on_deactivate, on_cleanup, on_shutdown
- Added `rclcpp_lifecycle` dependency to CMakeLists.txt and package.xml
- Updated launch files (ekf.launch.py, ukf.launch.py) to expose parameter
- Added comprehensive documentation (doc/lifecycle_support.rst)
- Updated CHANGELOG.rst

## Testing
- Builds cleanly without errors
- All lifecycle callbacks implemented and integrated
- Ready for community review and testing

Closes #958

### DIFF
--- a/doc/lifecycle_support.rst
+++ b/doc/lifecycle_support.rst
@@ -3,7 +3,7 @@ Lifecycle Node Support
 
 `robot_localization` nodes (EKF/UKF) support the ROS 2 managed lifecycle. This allows running the node in either an automatic (backward-compatible) mode or in a managed mode where an external controller performs the lifecycle transitions.
 
-Control is via the ``lifecycle_managed_node`` boolean parameter:
+Control is via the ``autostart`` launch argument:
 
 - ``false`` (default): the node will configure and activate itself automatically on startup (preserves previous behavior).
 - ``true``: the node will remain in the UNCONFIGURED state and must be transitioned to CONFIGURED/ACTIVE with `ros2 lifecycle` commands.
@@ -15,7 +15,7 @@ Launch the EKF node in managed mode:
 
 .. code-block:: bash
 
- ros2 launch robot_localization ekf.launch.py lifecycle_managed_node:=true
+ ros2 launch robot_localization ekf.launch.py autostart:=true
 
 Manage the lifecycle of a running node (replace ``/ekf_node`` with your node name if different):
 
@@ -30,6 +30,6 @@ Manage the lifecycle of a running node (replace ``/ekf_node`` with your node nam
 Notes
 -----
 
-- When running with ``lifecycle_managed_node:=true``, the node will not publish filtered output until it has been activated.
-- The launch files ``launch/ekf.launch.py`` and ``launch/ukf.launch.py`` expose the ``lifecycle_managed_node`` argument.
+- When running with ``autostart:=false`` (default), the node will not publish filtered output until it has been activated.
+- The launch files ``launch/ekf.launch.py`` and ``launch/ukf.launch.py`` expose the ``autostart`` argument.
 - Default behavior (``false``) preserves compatibility with existing setups that expect the node to start and publish immediately.
--- a/doc/lifecycle_support.rst
+++ b/doc/lifecycle_support.rst
@@ -5,8 +5,8 @@ Lifecycle Node Support
 
 Control is via the ``autostart`` launch argument:
 
-- ``false`` (default): the node will configure and activate itself automatically on startup (preserves previous behavior).
-- ``true``: the node will remain in the UNCONFIGURED state and must be transitioned to CONFIGURED/ACTIVE with `ros2 lifecycle` commands.
+- ``true`` : The node will configure and activate itself automatically on startup (preserves previous behavior).
+- ``false`` (default): The node will remain in the UNCONFIGURED state and must be transitioned to CONFIGURED/ACTIVE with ``ros2 lifecycle`` commands.
 
 Examples
 --------
@@ -15,21 +15,21 @@ Launch the EKF node in managed mode:
 
 .. code-block:: bash
 
- ros2 launch robot_localization ekf.launch.py autostart:=true
+   ros2 launch robot_localization ekf.launch.py autostart:=false
 
 Manage the lifecycle of a running node (replace ``/ekf_node`` with your node name if different):
 
 .. code-block:: bash
 
- ros2 lifecycle set /ekf_node configure
- ros2 lifecycle set /ekf_node activate
- ros2 lifecycle set /ekf_node deactivate
- ros2 lifecycle set /ekf_node cleanup
- ros2 lifecycle set /ekf_node shutdown
+   ros2 lifecycle set /ekf_node configure
+   ros2 lifecycle set /ekf_node activate
+   ros2 lifecycle set /ekf_node deactivate
+   ros2 lifecycle set /ekf_node cleanup
+   ros2 lifecycle set /ekf_node shutdown
 
 Notes
 -----
 
-- When running with ``autostart:=false`` (default), the node will not publish filtered output until it has been activated.
+- When running with ``autostart:=false``, the node will not publish filtered output until it has been activated via a lifecycle transition.
 - The launch files ``launch/ekf.launch.py`` and ``launch/ukf.launch.py`` expose the ``autostart`` argument.
-- Default behavior (``false``) preserves compatibility with existing setups that expect the node to start and publish immediately.
+- Default behavior (``true``) preserves compatibility with existing setups that expect the node to start and publish immediately.
--- a/doc/lifecycle_support.rst
+++ b/doc/lifecycle_support.rst
@@ -1,0 +1,35 @@
+Lifecycle Node Support
+======================
+
+`robot_localization` nodes (EKF/UKF) support the ROS 2 managed lifecycle. This allows running the node in either an automatic (backward-compatible) mode or in a managed mode where an external controller performs the lifecycle transitions.
+
+Control is via the ``lifecycle_managed_node`` boolean parameter:
+
+- ``false`` (default): the node will configure and activate itself automatically on startup (preserves previous behavior).
+- ``true``: the node will remain in the UNCONFIGURED state and must be transitioned to CONFIGURED/ACTIVE with `ros2 lifecycle` commands.
+
+Examples
+--------
+
+Launch the EKF node in managed mode:
+
+.. code-block:: bash
+
+ ros2 launch robot_localization ekf.launch.py lifecycle_managed_node:=true
+
+Manage the lifecycle of a running node (replace ``/ekf_node`` with your node name if different):
+
+.. code-block:: bash
+
+ ros2 lifecycle set /ekf_node configure
+ ros2 lifecycle set /ekf_node activate
+ ros2 lifecycle set /ekf_node deactivate
+ ros2 lifecycle set /ekf_node cleanup
+ ros2 lifecycle set /ekf_node shutdown
+
+Notes
+-----
+
+- When running with ``lifecycle_managed_node:=true``, the node will not publish filtered output until it has been activated.
+- The launch files ``launch/ekf.launch.py`` and ``launch/ukf.launch.py`` expose the ``lifecycle_managed_node`` argument.
+- Default behavior (``false``) preserves compatibility with existing setups that expect the node to start and publish immediately.
--- a/doc/lifecycle_support.rst
+++ b/doc/lifecycle_support.rst
@@ -5,8 +5,8 @@ Lifecycle Node Support
 
 Control is via the ``autostart`` launch argument:
 
-- ``true`` : The node will configure and activate itself automatically on startup (preserves previous behavior).
-- ``false`` (default): The node will remain in the UNCONFIGURED state and must be transitioned to CONFIGURED/ACTIVE with ``ros2 lifecycle`` commands.
+- ``true`` (default): The node will configure and activate itself automatically on startup (preserves previous behavior).
+- ``false`` : The node will remain in the UNCONFIGURED state and must be transitioned to CONFIGURED/ACTIVE with ``ros2 lifecycle`` commands.
 
 Examples
 --------
@@ -33,3 +33,4 @@ Notes
 - When running with ``autostart:=false``, the node will not publish filtered output until it has been activated via a lifecycle transition.
 - The launch files ``launch/ekf.launch.py`` and ``launch/ukf.launch.py`` expose the ``autostart`` argument.
 - Default behavior (``true``) preserves compatibility with existing setups that expect the node to start and publish immediately.
+- **Deprecation Note**: Starting in the *Lyrical* release, the default for ``autostart`` will change from ``true`` to ``false``.
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -60,7 +60,8 @@
 namespace robot_localization
 {
 
-class NavSatTransform : public rclcpp::Node {
+class NavSatTransform : public rclcpp::Node
+{
 public:
   /**
    * @brief Constructor
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -43,6 +43,7 @@
 #include "nav_msgs/msg/odometry.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp/timer.hpp"
+#include "rclcpp_lifecycle/lifecycle_node.hpp"
 #include "robot_localization/srv/from_ll.hpp"
 #include "robot_localization/srv/from_ll_array.hpp"
 #include "robot_localization/srv/set_datum.hpp"
@@ -60,7 +61,7 @@
 namespace robot_localization
 {
 
-class NavSatTransform : public rclcpp::Node
+class NavSatTransform : public rclcpp_lifecycle::LifecycleNode
 {
 public:
   /**
--- a/include/robot_localization/navsat_transform.hpp
+++ b/include/robot_localization/navsat_transform.hpp
@@ -43,7 +43,6 @@
 #include "nav_msgs/msg/odometry.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp/timer.hpp"
-#include "rclcpp_lifecycle/lifecycle_node.hpp"
 #include "robot_localization/srv/from_ll.hpp"
 #include "robot_localization/srv/from_ll_array.hpp"
 #include "robot_localization/srv/set_datum.hpp"
@@ -61,8 +60,7 @@
 namespace robot_localization
 {
 
-class NavSatTransform : public rclcpp_lifecycle::LifecycleNode
-{
+class NavSatTransform : public rclcpp::Node {
 public:
   /**
    * @brief Constructor
--- a/include/robot_localization/ros_filter.hpp
+++ b/include/robot_localization/ros_filter.hpp
@@ -424,39 +424,6 @@ protected:
   void aggregateDiagnostics(
     diagnostic_updater::DiagnosticStatusWrapper & wrapper);
 
-  //! @brief Declare a parameter if it has not been declared already.
-  //!
-  //! This replaces the old local lambda in loadParams() to keep parameter
-  //! declaration logic centralized and reusable.
-  //! @param[in] name - Parameter name
-  //! @param[in] type - Parameter type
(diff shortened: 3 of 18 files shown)


include/robot_localization/ros_filter.hpp
```diff
@@ -496,6 +534,13 @@ class RosFilter : public rclcpp::Node
     std::vector<bool> & updateVector, Eigen::VectorXd & measurement,
     Eigen::MatrixXd & measurementCovariance);
 
+  //! @brief Whether the node requires external lifecycle management
+  //!
+  //! When false (default), the node auto-transitions through configure and activate
+  //! on startup, maintaining backward compatibility. When true, external lifecycle
+  //! management via ros2 lifecycle commands is required.
+  bool lifecycle_managed_node_;
```
SteveMacenski: I think it should always be a managed node and have no auto-transition like this. There are ways to do this via the launch file instead.

launch/ekf.launch.py
```diff
@@ -24,12 +25,22 @@
...
 def generate_launch_description():
+    # Declare lifecycle management parameter
+    lifecycle_managed_param = DeclareLaunchArgument(
+        'lifecycle_managed_node',
+        default_value='false',
+        description='Enable lifecycle management for the EKF node')
+    
     return LaunchDescription([
+        lifecycle_managed_param,
         launch_ros.actions.Node(
```
SteveMacenski: If you use the `LifecycleNode` then there's an `autostart`  param in rolling + newer that'll auto transition if desired

launch/ukf.launch.py
```diff
@@ -24,12 +25,22 @@
 from launch.actions import DeclareLaunchArgument
 
 def generate_launch_description():
+    # Declare lifecycle management parameter
+    lifecycle_managed_param = DeclareLaunchArgument(
+        'lifecycle_managed_node',
+        default_value='false',
+        description='Enable lifecycle management for the UKF node')
+    
     return LaunchDescription([
```
SteveMacenski, line 29: Ditto, and then remove this launch configuration for the `lifecycle_managed_node`

src/ekf_node.cpp
```diff
@@ -30,21 +30,82 @@
...
 
+namespace
+{
+std::atomic_bool g_sigint_requested{false};
+
+void sigintHandler(int signum)
+{
+  static_cast<void>(signum);
+  g_sigint_requested.store(true);
+}
```
SteveMacenski: This should also be removed from all nodes

SteveMacenski: In general many of the changes here are not necessary if not `lifecycle_managed_node`

src/ros_filter.cpp
```diff
@@ -827,6 +1025,27 @@ void RosFilter<T>::differentiateMeasurements(const rclcpp::Time & current_time)
 template<typename T>
 void RosFilter<T>::loadParams()
 {
+  auto declare_parameter_if_not_declared =
```
SteveMacenski: Remove this lambda

src/ros_filter.cpp
```diff
@@ -2077,6 +2299,11 @@ void RosFilter<T>::poseCallback(
 template<typename T>
 void RosFilter<T>::initialize()
```
SteveMacenski: If we're doing lifecyckle, this method should be removed and its implementation should be happening on configure / activate as appropriate

SteveMacenski (changes requested):
Good first stab - in general lets not change unrelated things like declaring parameters lambdas

include/robot_localization/ros_filter.hpp
```diff
@@ -496,6 +530,7 @@ class RosFilter : public rclcpp::Node
     std::vector<bool> & updateVector, Eigen::VectorXd & measurement,
     Eigen::MatrixXd & measurementCovariance);
 
+
```
SteveMacenski: Remove the extra line please :-)

include/robot_localization/navsat_transform.hpp
```diff
@@ -60,7 +60,8 @@
 namespace robot_localization
 {
 
-class NavSatTransform : public rclcpp::Node {
+class NavSatTransform : public rclcpp::Node
```
SteveMacenski: Should this also be a lifecycle node?

src/ekf_node.cpp
```diff
@@ -33,18 +33,53 @@
...
+
+  RCLCPP_INFO(
+    filter->get_logger(),
+    "Lifecycle node initialized; use lifecycle transitions to configure/activate.");
+
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(filter->get_node_base_interface());
+  while (rclcpp::ok()) {
+    executor.spin_some();
+  }
```
SteveMacenski: You should do `spin()` I think

src/ekf_node.cpp
```diff
@@ -33,18 +33,53 @@
...
+
+  // Ensure lifecycle node is properly shut down to avoid warnings on exit.
```
SteveMacenski: I think we  can remove this, its up to the application to do this properly.

src/ros_filter.cpp
```diff
@@ -151,44 +187,214 @@ RosFilter<T>::~RosFilter()
...
+
+  topic_subs_.clear();
+  RCLCPP_WARN(get_logger(), "[%s]: Finalized. System resources released.", get_name());
+  return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+}
+
 template<typename T>
 void RosFilter<T>::reset()
 {
-  // Get rid of any initial poses (pretend we've never had a measurement)
```
SteveMacenski, line 158: Can you readd all the comments / extra lines?

src/ros_filter.cpp
```diff
@@ -1192,23 +1388,19 @@ void RosFilter<T>::loadParams()
       &RosFilter<T>::toggleFilterProcessingCallback, this,
       std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
 
-  // Init the last measurement time so we don't get a huge initial delta
```
SteveMacenski, line 1201: Ditto everywhere :-)

src/navsat_transform.cpp
```diff
@@ -65,7 +66,7 @@ using namespace std::chrono_literals;
 namespace robot_localization
 {
 NavSatTransform::NavSatTransform(const rclcpp::NodeOptions & options)
-: Node("navsat_transform_node", options),
+: rclcpp_lifecycle::LifecycleNode("navsat_transform_node", options),
```
SteveMacenski: Does this need to implement the configure/activate/deactivate/cleanup/shutdown states?

SteveMacenski (changes requested):
Otherwise, I think its ready for Tom to take a look

Do look over the PR diff @BOOPESH-foxy - there's still some new spaces added / previous removed that should be reverted.

src/navsat_transform.cpp
BOOPESH-foxy: While I've updated navsat_transform to inherit from LifecycleNode for basic architectural alignment, I haven't implemented the full internal state logic there yet. Since navsat_transform is often used as a helper rather than a core filter, I believe it's best to keep this PR focused on getting the EKF/UKF implementation perfect. I am happy to follow up with a deeper navsat_transform refactor in a separate PR if we decide it's needed.

BOOPESH-foxy:
I’ve pushed an update to further minimize the diff by aligning whitespace and include ordering with the current rolling-devel source. As this is my first contribution here, I’m learning a lot from your feedback and appreciate the guidance through these iterations.
 
All local tests (43/43) are passing, including the linter and node interface checks. Ready for another look when you have a moment!

src/navsat_transform.cpp
SteveMacenski: This is pretty core as the main tool for GPS handling in ROS 2 :-)

SteveMacenski (commented):
Otherwise LGTM -- I think this just needs @ayrton04 's stamp of approval (minus needing nav sat transform)

src/navsat_transform.cpp
BOOPESH-foxy: I completely agree. Given that navsat_transform is a core tool for GPS handling, it deserves a more rigorous lifecycle implementation.

To keep the commit history and validation for the EKF/UKF work clean and 'ready-to-merge,' I’m going to handle the navsat_transform refactor in a separate branch and PR. This allows for a deeper dive into moving its subscribers and TF listeners into the on_activate stage without delaying the approval of the core filters. I'll link the new issue/PR here once it's live!

SteveMacenski: OK - then we should revert the change for its base class in this PR as well

BOOPESH-foxy: Hi Steve, I checked the diff again. I actually reverted the navsat_transform source or header files in this PR — now all changes are isolated to the RosFilter base (for EKF/UKF) and the specific EKF/UKF nodes.

I have finalized the isolation. In commit 2132564, I’ve ensured that navsat_transform remains a standard node with zero changes to its source or header. I've just opened the new issue for the deep refactor of navsat_transform and will begin that work in a fresh branch today!

doc/lifecycle_support.rst
```diff
@@ -0,0 +1,35 @@
+Lifecycle Node Support
+======================
+
+`robot_localization` nodes (EKF/UKF) support the ROS 2 managed lifecycle. This allows running the node in either an automatic (backward-compatible) mode or in a managed mode where an external controller performs the lifecycle transitions.
+
+Control is via the ``autostart`` launch argument:
+
+- ``true`` : The node will configure and activate itself automatically on startup (preserves previous behavior).
+- ``false`` (default): The node will remain in the UNCONFIGURED state and must be transitioned to CONFIGURED/ACTIVE with ``ros2 lifecycle`` commands.
```
ayrton04: I'm not sure I agree with it being the default straight away, as literally everyone else using the package will be expecting the current behaviour. I might suggest we:

1. Until they cut Lyrical, does it make sense for this to default to `true`, but if the parameter is not explicitly provided, we warn the user about it?
2. Once Lyrical is cut, we change the default to `false`

This seems like a gentler transition path.

src/ros_filter.cpp
```diff
@@ -131,6 +132,42 @@ RosFilter<T>::RosFilter(const rclcpp::NodeOptions & options)
...
+  this->declare_parameter("publish_acceleration", false);
+  this->declare_parameter("permit_corrected_publication", false);
+  this->declare_parameter("predict_to_current_time", false);
+  this->declare_parameter("two_d_mode", false);
+  this->declare_parameter("smooth_lagged_data", false);
+  this->declare_parameter("reset_on_time_jump", false);
+  this->declare_parameter("use_control", false);
+  this->declare_parameter("stamped_control", true);
+  this->declare_parameter("dynamic_process_noise_covariance", false);
+  this->declare_parameter("disabled_at_startup", false);
```
ayrton04, line 154: Would we consider this parameter to be redundant now? As it stands, this parameter controls the `enabled_` flag, which states whether the filter runs. It can be set here at startup, and then controlled via a service call. Given the tick-tock deprecation model I am suggesting, we would leave this alone until Lyrical is cut, but then I wonder if we need to just let the lifecycle architecture manage this.

src/ros_filter.cpp
```diff
@@ -151,6 +188,191 @@ RosFilter<T>::~RosFilter()
...
+    timer_->cancel();
+    timer_.reset();
+  }
+
+  position_pub_.reset();
+  accel_pub_.reset();
+  set_pose_sub_.reset();
+  control_sub_.reset();
+  stamped_control_sub_.reset();
+  topic_subs_.clear();
```
ayrton04, line 347: I'm assuming the lifecycle management happens on the same thread as the rest of the execution. If not, then we'd want to call `reset()` (currently line 331) _after_ we shut down all the subs. But then, if there's any chance of that happening asynchronously, then a lot of this is not thread-safe, so I'm imagining we're fine.

src/ros_filter.cpp
```diff
@@ -151,6 +188,191 @@ RosFilter<T>::~RosFilter()
...
+    get_logger(), "[%s]: Transitioning to 'shutdown' state. Process exiting.",
+    get_name());
+
+  // Stop timer if running
+  if (timer_) {
+    timer_->cancel();
+    timer_.reset();
+  }
+
+  topic_subs_.clear();
```
ayrton04, line 379: Why do we do this here and in `on_cleanup`, but not the rest?

src/ros_filter.cpp
```diff
@@ -882,14 +1103,14 @@ void RosFilter<T>::loadParams()
...
-    std::string("base_link"));
-  base_link_output_frame_id_ = this->declare_parameter(
-    "base_link_frame_output",
-    base_link_frame_id_);
+  this->get_parameter("map_frame", map_frame_id_);
+  this->get_parameter("odom_frame", odom_frame_id_);
+  this->get_parameter("base_link_frame", base_link_frame_id_);
+
+  // Handle base_link_frame_output with fallback to base_link_frame
+  if (!this->get_parameter("base_link_frame_output", base_link_output_frame_id_)) {
```
ayrton04: Would it make more sense to use [get_parameter_or](https://docs.ros.org/en/rolling/p/rclcpp/generated/classrclcpp_1_1Node.html#_CPPv4I0ENK6rclcpp4Node16get_parameter_orEbRKNSt6stringER10ParameterTRK10ParameterT)?

```
this->get_parameter_or("base_link_frame_output", base_link_output_frame_id_, base_link_frame_id_);
```

src/ros_filter.cpp
```diff
@@ -916,7 +1137,9 @@ void RosFilter<T>::loadParams()
    *
    * The default is the latter behavior (broadcast of odom->base_link).
    */
-  world_frame_id_ = this->declare_parameter("world_frame", odom_frame_id_);
+  if (!this->get_parameter("world_frame", world_frame_id_)) {
```
ayrton04: Another case for `get_parameter_or`.

ayrton04:
Thanks for this PR. I'm working my way through it. I'll try to wrap up tomorrow.

src/ros_filter.cpp
```diff
@@ -1237,25 +1467,39 @@ void RosFilter<T>::loadParams()
...
 
       // Consider odometry transformation from the child_frame_id instead of the base_link_frame_id
-      bool pose_use_child_frame = this->declare_parameter(
-        odom_topic_name + std::string("_pose_use_child_frame"), false);
+      std::string child_frame_param = odom_topic_name + std::string("_pose_use_child_frame");
+      if (!this->has_parameter(child_frame_param)) {
+        this->declare_parameter(child_frame_param, false);
+      }
+
+      // Consider odometry transformation from the child_frame_id instead of the base_link_frame_id
```
ayrton04: Nit: repeated the comment on line 1469.

ayrton04 (changes requested):
Overall, this is really great, thank you! I just think that, given the proximity to cutting Lyrical, we might want to be more gentle with its introduction. I am not married to this opinion, though. I'm interested to hear others' thoughts.

BOOPESH-foxy:
Thank you for the detailed review Tom ! This 'Tick-Tock' transition path makes a lot of sense to avoid breaking existing deployments.

    I will update autostart to default to true and add the warning for the transition to Lyrical.

    I'll refactor the parameter lookups using get_parameter_or as suggested—that definitely makes the code much cleaner.

    I'll consolidate the subscription resets to ensure consistency across the lifecycle states.

    Regarding disabled_at_startup, I agree it’s redundant with the lifecycle architecture; I’ll leave it for now to preserve the API and we can pull it in the Lyrical release.

I'll push these updates shortly!

doc/lifecycle_support.rst
SteveMacenski: I agree - anyway these launch files in RL are demonstrations of the software so I think automatic bringup is the most sensible (Nav2 autostarts as well, though by a different mechanism).

src/ros_filter.cpp
SteveMacenski: I'd leave both, I think they mean semantically different things and making people migrate via lifecycle would be annoying without much benefit.

BOOPESH-foxy:
I've updated the parameter logic, doc files(.rst), launch defaults as discussed and updated it the branch in commit a9f0bb732fc22472529ca7fdfa3cf058c06ab430. 

Regarding the on_deactivate and on_cleanup logic, I'd like your guidance on the teardown sequence :

During my initial testing of moving subscriptions to on_activate, I encountered significant numerical instability (NaNs) due to the timing jump between node creation and activation. To maintain the mathematical stability of the EKF/UKF:

    Subscription Management: Should we explicitly .reset() every individual subscriber (set_pose_sub_, control_sub_, etc.) in on_deactivate to ensure a total halt of data flow, or is clearing topic_subs_ sufficient in your view?

    Teardown Order: To address potential thread-safety concerns with asynchronous callbacks, do you prefer a strict 'unplug then wipe' order (clearing all subscribers before calling the internal filter reset())?

I wanted to confirm your preferred architectural pattern for this teardown before finalizing those blocks to ensure they align with the project's long-term maintenance goals.

src/ros_filter.cpp
ayrton04: Works for me!

ayrton04:
> I encountered significant numerical instability (NaNs) due to the timing jump between node creation and activation.

Are you certain this was the cause of the NaN values?

> Subscription Management: Should we explicitly .reset() every individual subscriber (set_pose_sub_, control_sub_, etc.) in on_deactivate to ensure a total halt of data flow, or is clearing topic_subs_ sufficient in your view?

I would say that `on_deactivate` should leave the node in the same state that it was before `on_activate` was called for the first time.

> Teardown Order: To address potential thread-safety concerns with asynchronous callbacks, do you prefer a strict 'unplug then wipe' order (clearing all subscribers before calling the internal filter reset())?

I'm not sure this is actually a concern, but I'd recommend just making sure the teardown order happens in reverse from the construction order.

BOOPESH-foxy:
Hi @ayrton04,

Thank you for the detailed feedback! I've addressed your concerns in commit https://github.com/cra-ros-pkg/robot_localization/pull/959/commits/f07bf664c5c6890e94ce9f77e242f798f973e13f

You're absolutely right about restoring the pre-activation state. I've updated on_deactivate() to now destroy publishers (not just deactivate them) so the node returns to the exact state it was in before the first on_activate() call. Subscribers remain active since they're created in on_configure(), which aligns with the standard lifecycle pattern.

Reverse Construction Order: Implemented strict reverse teardown order in on_deactivate():
    Stop timer
    Deactivate and destroy publishers
    Disable filter and call reset()
    This mirrors the construction order from on_activate() in reverse.

Thread Safety: While you mentioned you weren't sure if this was a concern, I added enabled_ flag checks to all sensor callbacks (odometryCallback, imuCallback, poseCallback, twistCallback, accelerationCallback) as a defensive measure. This prevents any potential race conditions during state transitions.

Filter State Reset: Added reset() call in on_deactivate() to clear the measurement queue, filter history, and timing variables. This prevents stale data and timing jumps when reactivating the node.

I've tested both EKF and UKF nodes through multiple activate/deactivate cycles and confirmed clean transitions without NaN errors or data leakage.

Let me know if you'd like any adjustments!

ayrton04 (approved):

ayrton04:
Thanks again!

PR #969, opened 2026-04-28
Support for rosdoc2 build

mwopfner:
The changes enable that `rosdoc2 build -p .` works in the root of the repository. Thus the documentation should be published during the rosdistro build. 
The old documentation build (`make html`) still works.
 
The only necessary change was to move the  `conf.py` file into a subfolder so that `rosdoc2` does not use it, because it is not needed and produces errors.

The new generated documentation looks complete but I did not looked through every page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@ build
 install
 log
 doc/html
+doc/.build
+docs_build
+docs_output
 *.cproject
 *.project
 *.settings
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -c prev_config
 SPHINXBUILD   = sphinx-build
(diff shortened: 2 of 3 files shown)


ayrton04:
Very helpful, thank you!

ayrton04 (approved):

mwopfner:
If you find any problems, or missing things I can try to fix them. Currently also the C++ API documentation is generated. If the documentation should be closer to the original one this can be achieved by adding a new `conf.py` file into the `/doc` folder that is than used by `rosdoc2`.

```
# Configuration file for the Sphinx documentation builder.
# settings will be overridden by rosdoc2, so we add here only custom settings

copyright = u'2016, Tom Moore'
html_logo = 'images/rl_small.png'
```
This documentation is than really close to the old one only with the new style of the ros2 documentation.

ayrton04:
@mwopfner just a few things:

https://github.com/cra-ros-pkg/robot_localization/pull/970

PR #970, opened 2026-04-28
Fixing rosdoc

ayrton04:


mwopfner:
I have tested it with `rosdoc2 build -p .` but it fails. To fix it the line `sphinx_sourcedir: 'doc'` in `rosdoc2.yaml` must be removed or a `conf.py` file must be created in `/doc`. It complains that it can not find a `conf.py` in the doc folder.

I tested it with this one:
```
# Configuration file for the Sphinx documentation builder.
# settings will be overridden by rosdoc2, so we add here only custom settings

copyright = u'2016, Tom Moore'
html_logo = 'images/rl_small.png'
```
and then the generation works fine

SteveMacenski:
I'll be honest that I don't know much about rosdoc, but if it works; it works! :smile: 

Nothing here looks suspect to me from a cursory look

mwopfner:
I tested with the new changes and now both doc generations work and also the docs looks good as far as I can see it.

mwopfner:
According to the ros2 documentation (https://docs.ros.org/en/jazzy/How-To-Guides/Documenting-a-ROS-2-Package.html#ci-docs-ros-org) there is an additional entry required in https://github.com/ros/rosdistro/blob/master/rolling/distribution.yaml#L7459 so that the documentation is actually published with rosdistro.

I think the entry should look like this:
```
    doc:
      type: git
      url: https://github.com/cra-ros-pkg/robot_localization.git
      version: rolling-devel
```

ayrton04:
@mwopfner yes, it's the same stanza we used [for the ROS 1 distros](https://github.com/ros/rosdistro/blob/ccf50b0545d95d67f7d32a09d0eedfb80cb64e1f/noetic/distribution.yaml#L9427-L9430).

